PR #2, opened 2016-12-14
Replace List<Subscription> with CompositeSubscription

Plastix:
CompositeSubscription is RxJava's representation of a group of
Subscriptions.


luoser:
Nice change. Thanks!

PR #3, opened 2016-12-14
Remove APT plugin

Plastix:
The latest Android Gradle plugin (>2.2.0) has built-in support for
annotation processors. Thus, we don't need APT anymore. More info:
http://www.littlerobots.nl/blog/Whats-next-for-android-apt/


luoser:
great find, thanks!

PR #5, opened 2016-12-15
Update git sha only in release build

riyaz:
- setting the git sha in default config will break the incremental build and increase the build time for debug builds since BuildConfig will always be changed if there is a new commit, this might not be intended for debug build
- we can set git sha only in release build

Source :

- https://www.reddit.com/r/androiddev/comments/3az8qd/add_a_git_sha_to_your_crash_reporting_tools/cshk90h/?st=iwpqdxmw&sh=55670f2c

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -33,7 +33,7 @@ android {
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
 
         buildConfigField "String", "GIT_SHA", "\"DUMMY\""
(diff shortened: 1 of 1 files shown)


Plastix:
This gradle plugin https://github.com/zsoltk/paperwork works well for generating git sha without breaking incremental compilation.

luoser:
thanks for this suggestion! we actually like using the sha and date in our internal builds as well to provide more information for debugging. we just merged in #92 to store the sha and date so closing this

PR #13, opened 2016-12-15
Fix to use uppercase letter notation to long value.

hotchemi:
### Overview

- This change just fixes to use uppercase letter notation to long value.


hotchemi:
@luoser Any update?

luoser:
yep, this is a good fix--it's good practice to use uppercase `L`, and we do use it in other places of our codebase. could you possibly change the lowercase `l`s in `HelpActivity` as well while we're at it?  thanks!

hotchemi:
Sure thing! Addressed.

PR #25, opened 2016-12-18
Remove unused class member (viewModel) in ProjectCardViewHolder

miga9:
Removed an unused variable of DiscoveryViewModel in ProjectCardViewHolder.


luoser:
heh woops, nice catch

PR #27, opened 2016-12-19
Apply slideInFromLeft transition to LoginToutActivity

miga9:
This PR defines Activity exit transition for LoginToutActivity. 

The exit transition was not defined; therefore, the default exit transition was being triggered which does not match the enter transition.


miga9:
@luoser Hi, do you have any updates on this one?

luoser:
Our LoginTout is actually triggered from multiple places: the side menu, the "Onboarding" Login callout in the Discovery feed, the empty Activity feed, an empty Comments feed, and when starring the project as a logged out user.

In general, we treat the LoginTout as a modal--the only place where the Tout appears via `slideInFromRight` is the side menu, which makes me think that we should remove that enter transition rather than refactoring the rest

luoser:
closing this for now-- an overall transition pass is on our radar

PR #29, opened 2016-12-21
Upgrade to 25

luoser:
## what
* Upgraded Android from 23 to 25 and outdated dependency versions. 
* Updated `Makefile`, `.gitignore`, and `script/bootstrap_configs` to include new secrets
* Updated `circle.yml` to get on 25 build tools etc
* Bonus bug fixes due to dependency injection


Makefile
```diff
@@ -13,15 +13,29 @@ submodules:
 secrets:
 	-rm -rf vendor/native-secrets
 	-git clone https://github.com/kickstarter/native-secrets vendor/native-secrets
-	if [ ! -d vendor/native-secrets ]; \
```
luoser: i essentially reverted `secrets:` to what it was before adding this `if` statement; the if statement only copied over examples but not the actual secret files for when `vendor/native-secrets` exists

script/bootstrap_configs
```diff
@@ -7,4 +7,14 @@ do
...
+
+VARIANTS="externalMin21Debug externalPre21Debug externalMin21Release externalPre21Release \
+  internalMin21Debug internalPre21Debug internalMin21Release internalPre21Release"
+
 cd ..
+
+# Copy google services over. Fallback to example if they don't exist.
+for v in $VARIANTS
+do
+  cp vendor/native-secrets/android/"$v"/google-services.json app/src/"$v"/google-services.json || cp config/google-services.example.json app/src/"$v"/google-services.json
```
luoser: ideally i'd like for this to be within the `Makefile` but it is nice to have a straightforward way of stepping through each build variant's director

circle.yml
```diff
@@ -14,7 +14,12 @@ dependencies:
   override:
     - $ANDROID_HOME/tools/android list target
     - make bootstrap-circle
-    - echo y | android update sdk --no-ui --all --filter android-23,build-tools-23.0.1,extra-android-m2repository,extra-google-m2repository,extra-android-support
+    - echo y | android update sdk --no-ui --all --filter android-25
```
luoser: > Note: Only install one package at a time, as echo y will only work for one license. If you don’t do it this way, Android will give you an error message but won’t fail and your build continues. This is undesirable since it makes debugging very difficult.

from https://circleci.com/docs/android/#dependencies

app/src/main/java/com/kickstarter/KSApplication.java
```diff
@@ -86,7 +85,7 @@ public boolean isInUnitTests() {
   }
 
   private void checkForCrashes() {
-    final String appId = this.build.isExternal()
+    final String appId = Build.isExternal()
```
luoser: this was crashing on app start since it's invoked in `onCreate()` for release builds and the injected `build` was `null` before this check

Makefile
```diff
@@ -13,15 +13,29 @@ submodules:
...
-		mkdir -p app/src/debug/res/values \
-			&& cp app/src/main/java/com/kickstarter/libs/utils/Secrets.java.example app/src/main/java/com/kickstarter/libs/utils/Secrets.java \
-			&& cp config/koala_endpoint.xml.example app/src/debug/res/values/koala_endpoint.xml \
-			&& cp config/koala_endpoint.xml.example app/src/main/res/values/koala_endpoint.xml \
-			&& cp -rf config/WebViewJavascriptInterface.java app/src/main/java/com/kickstarter/libs/WebViewJavascriptInterface.java \
-			&& cp -rf config/KSWebViewClient.java app/src/main/java/com/kickstarter/services/KSWebViewClient.java \
-			|| true; \
-	fi
+
+	# TODO: temp bootstrap-circle workaround. Consolidate the cp services commands.
```
luoser: todo related to my comment below on `build_configs` script

app/src/main/java/com/kickstarter/KSApplication.java
mbrandonw: sorry just saw this... i would think we want this properly injected tho right? why did injection fail?

luoser: it failed because [`checkForCrashes()`](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/KSApplication.java#L42) happens before the [build is injected](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/KSApplication.java#L61)

PR #38, opened 2016-12-27
Upgrade Dagger to 2.8 and use static provides

Plastix:
According to [Android documentation](https://developer.android.com/training/articles/perf-tips.html#PreferStatic), static methods are 15-20% faster. This PR makes all `@Provides` methods static when applicable. 

More info on Dagger's static provide methods: https://youtu.be/iwjXqRlEevg?t=5m7s


luoser:
heya, sorry that we've punted on this for _so_ long but if you could fix the spacing so that it matches the Kickstarter style (should be accessible via Preferences > Code Style) and merge latest would be great to get this in!

luoser:
closing here so we can do the merge in #189

PR #39, opened 2016-12-28
Remove @Retention from custom Dagger qualifiers

Plastix:
`@Qualifier` itself is annotated with `@Retention(RUNTIME)`.


luoser:
oldie but goodie, thanks!

PR #40, opened 2016-12-29
Round reward conversion label up

luoser:
## what
Rounded our USD conversion text up so we avoid showing `$0` amounts in rewards.

 → 

We want to be sure to only round `UP` for currency in the reward conversion, since generally in other use cases we want the currency to be rounded down.

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/RewardViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/RewardViewModel.java
@@ -210,7 +210,7 @@ public final class RewardViewModel extends ActivityViewModel<RewardViewHolder> i
       .subscribe(usdConversionTextViewIsHidden);
 
     projectAndReward
-      .map(pr -> ksCurrency.format(pr.second.minimum(), pr.first, true, true))
+      .map(pr -> ksCurrency.format((float) Math.ceil(pr.second.minimum()), pr.first, true, true))
       .compose(takeWhen(
         shouldDisplayUsdConversion
           .filter(BooleanUtils::isTrue)
--- a/app/src/main/java/com/kickstarter/viewmodels/RewardViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/RewardViewModel.java
@@ -24,6 +24,7 @@ import com.kickstarter.viewmodels.outputs.RewardViewModelOutputs;
 
 import org.joda.time.DateTime;
 
+import java.math.RoundingMode;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


app/src/test/java/com/kickstarter/viewmodels/RewardViewModelTest.java
```diff
@@ -497,6 +497,28 @@ public void testUsdConversionNotShownForUSProject() {
...
 
   @Test
+  public void testUsdConversionTextRoundsUp() {
+    // Set user's country to US.
+    final Config config = ConfigFactory.configForUSUser();
+    final Environment environment = environment();
+    environment.currentConfig().config(config);
+    final RewardViewModel vm = new RewardViewModel(environment);
+
+    // Set project's country to CA and reward minimum to $0.40.
```
christopherwright: This says $0.40 but the minimum is set to 0.3f

app/src/test/java/com/kickstarter/viewmodels/RewardViewModelTest.java
```diff
@@ -497,6 +497,28 @@ public void testUsdConversionNotShownForUSProject() {
...
+    final Project project = ProjectFactory.project().toBuilder().country("CA").build();
```
christopherwright: Can do `ProjectFactory.caProject()`

app/src/main/java/com/kickstarter/viewmodels/RewardViewModel.java
```diff
@@ -210,7 +210,7 @@ public RewardViewModel(final @NonNull Environment environment) {
       .subscribe(usdConversionTextViewIsHidden);
 
     projectAndReward
-      .map(pr -> ksCurrency.format(pr.second.minimum(), pr.first, true, true))
+      .map(pr -> ksCurrency.format((float) Math.ceil(pr.second.minimum()), pr.first, true, true))
```
christopherwright: It might be nice if `format` could take a rounding mode option, instead of manually using `ceil` and casting back to a float.

app/src/test/java/com/kickstarter/viewmodels/RewardViewModelTest.java
luoser: woooooops :neckbeard:

app/src/main/java/com/kickstarter/viewmodels/RewardViewModel.java
luoser: fine

app/src/test/java/com/kickstarter/viewmodels/RewardViewModelTest.java
```diff
@@ -497,6 +497,28 @@ public void testUsdConversionNotShownForUSProject() {
...
+    environment.currentConfig().config(config);
+    final RewardViewModel vm = new RewardViewModel(environment);
+
+    // Set project's country to CA and reward minimum to $0.30.
+    final Project project = ProjectFactory.caProject();
+    final Reward reward = RewardFactory.reward().toBuilder().minimum(0.3f).build();
+
+    final TestSubscriber<String> usdConversionTextViewText = TestSubscriber.create();
+    vm.outputs.usdConversionTextViewText().subscribe(usdConversionTextViewText);
+    final TestSubscriber<Boolean> usdConversionSectionIsHidden = TestSubscriber.create();
```
christopherwright: We can remove this and the next line since we don't assert on the `usdConversionSectionIsHidden` subscription, otherwise happy for this to be merged

PR #42, opened 2017-01-05
Suppress secrets error

stephencelis:
This should suppress the error emitted when attempting to clone the private, non-OSS repo, hopefully reducing the number of false positives reported.


mbrandonw:
thannnnnnnnkssssss!

Issue #30, opened 2016-12-22
Unable to run ./script/bootstrap - Function Not Found


timkaboya:
So I encountered #16 but on manually doing the file commands, that seemed to fix it.  
Now this script `./script/bootstrap` is giving this error. 

```
./script/bootstrap
./script/bootstrap: 3: ./script/bootstrap: function: not found

./script/bootstrap: 6: ./script/bootstrap: Syntax error: "}" unexpected
Makefile:2: recipe for target 'bootstrap' failed
make: *** [bootstrap] Error 2
```
It seems as tho its failing on function "say" in the script. Whats that function for? 

@Kickstarter This code you threw up online is not running...

LetItRock:
@timkaboya did you tried to run app without this bootstrap?

hodannyy:
@timkaboya I ran into the same issue and have a fix. Check out my pull request #43.

rome753:
on my Ubuntu 16, edit the ./script/bootstrap file with a text editor,
change "function say {..." to "say() {...",
and also "function loudspeaker {..." to "loudspeaker() {...",
then apt install ruby and curl,
then ok.

luoser:
closing now as @hodanny 's fix has been merged.

Issue #28, opened 2016-12-20
How to "make bootstrap" in windows


ngochuu90:
Hello

I have checkout android-oss by Github Desktop. But I cannot "make bootstrap" in Windows
![cmd](https://cloud.githubusercontent.com/assets/9895294/21338904/dff6d80e-c6ac-11e6-92cb-436bc0e6eea5.PNG)

michaldrabik:
@ngochuu90 #7 . But you'll probably encounter #16 on the way.

saisasanksunkavalli:
cd into the project repo, Run Window Bash (Git Bash) and run ./script/bootstrap. Then import the project in Android Studio, it will build automatically

PatidarNikunj:
Hi,

I'm not able to RUN bootstrap, getting errors:

```
$ ./Makefile
./Makefile: line 1: bootstrap:: command not found
./script/bootstrap: line 5: ruby: command not found
find: warning: you have specified the -depth option after a non-option argument -name, but options are not positional (-depth affects tests specified before it as well as those specified after it).  Please specify options before other arguments.

find: paths must precede expression: 1
Usage: find [-H] [-L] [-P] [-Olevel] [-D help|tree|search|stat|rates|opt|exec] [path...] [expression]
./script/bootstrap: line 5: ruby: command not found
./script/bootstrap: line 28: bundle: command not found
./script/bootstrap: line 5: ruby: command not found
cp: cannot stat 'vendor/native-secrets/android/externalMin21Debug/google-services.json': No such file or directory
cp: cannot stat 'vendor/native-secrets/android/externalPre21Debug/google-services.json': No such file or directory
cp: cannot stat 'vendor/native-secrets/android/externalMin21Release/google-services.json': No such file or directory
cp: cannot stat 'vendor/native-secrets/android/externalPre21Release/google-services.json': No such file or directory
cp: cannot stat 'vendor/native-secrets/android/internalMin21Debug/google-services.json': No such file or directory
cp: cannot stat 'vendor/native-secrets/android/internalPre21Debug/google-services.json': No such file or directory
cp: cannot stat 'vendor/native-secrets/android/internalMin21Release/google-services.json': No such file or directory
cp: cannot stat 'vendor/native-secrets/android/internalPre21Release/google-services.json': No such file or directory
mkdir: cannot create directory ‘/codestyles’: Permission denied
./script/bootstrap: line 10: ruby: command not found
[...]

Issue #7, opened 2016-12-15
Windows can't run make


ren545457803:
don't have command 'run'

luoser:
ah yes. unfortunately we don't use Windows for development on our team so we can definitely use some suggestions on how we can improve on that front!

AuspeXeu:
I am currently using Bash on Ubuntu on Windows for these purposes

https://msdn.microsoft.com/en-gb/commandline/wsl/about

ren545457803:
@AuspeXeu Thanks

ashishbhandari:
good to install cygwin on windows and run that command.. :)

Qamar4P:
This work for me. Run
**D:\CodeLab\Android-oss>**`"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\nmake" bootstrap`

Qamar4P:
Hey guys thanks to @LetItRock
1. Did copy and rename files manually refered in Makefile eg. "mkdir(Create directory) -p app/src/debug/res/values and  cp (Copy) app/src/main/java/com/kickstarter/libs/utils/Secrets.java.example app/src/main/java/com/kickstarter/libs/utils/Secrets.java
2. Run Window Bash (in my case Git Bash) as administrator and run ./script/bootstrap

Then build project works in Window.

mbrandonw:
Thanks for the tips for getting things to work in windows! Gonna close this for now, but will use this as a reference for when the issue comes up again!

PR #43, opened 2017-01-06
Use POSIX syntax in bootstrap script

hodanny:
This change addresses #7 #28 #30. I believe many Windows users are using [Bash on Ubuntu for Windows](https://msdn.microsoft.com/en-gb/commandline/wsl/about) or a POSIX-compliant shell to run the bootstrap script. The `function` keyword is not defined on POSIX. This change improves portability and should not break the bash shell on OSX. I am able to build and run the Kickstarter app successfully on Windows.

http://unix.stackexchange.com/questions/73750/difference-between-function-foo-and-foo

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,11 +1,11 @@
 #!/usr/bin/env sh
 
-function say {
+say () {
   str=$1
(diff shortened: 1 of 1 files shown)


luoser:
thank you for this change, and for providing the reference

PR #44, opened 2017-01-06
Pass categories from DiscoveryActivity to Fragment

luoser:
## what
This is some much needed infrastructural cleanup in our `DiscoveryViewModel` and `DiscoveryFragmentViewModel`. It paves a clearer route for tackling our blank sorts Discovery bug, trust me.

## changes
The main change is passing `rootCategories` as an `output` from `DiscoveryActivity` to the `DiscoveryFragment` via an `input`, rather than calling `fetchCategories` again in the fragment.

We need the categories in the fragment for our `DiscoveryUtils.fillRootCategoryForFeaturedProjects` function:
```java
  /**
   * Given a list of projects and root categories this will determine if the first project is featured
   * and is in need of its root category. If that is the case we will find its root and fill in that
   * data and return a new list of projects.
   */
```

Before: 4 fetch categories requests on app startup
image

After: 1 fetch categories request on app startup
image

Why didn't we do this earlier? Good question. An oversight in a large (sorts) project, my responsibility.

* Another major win is disentangling our `params` logic. I will explain below.
* A lot of noise is from modernizing the two Discovery VMs, i.e. moving inputs and outputs to the bottom of the file.

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
+++ b/app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
@@ -5,6 +5,8 @@ import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentPagerAdapter;
 
+import com.kickstarter.libs.utils.DiscoveryUtils;
+import com.kickstarter.models.Category;
 import com.kickstarter.services.DiscoveryParams;
 import com.kickstarter.ui.ArgumentsKey;
 import com.kickstarter.ui.fragments.DiscoveryFragment;
@@ -50,14 +52,14 @@ public final class DiscoveryPagerAdapter extends FragmentPagerAdapter {
   /**
    * Take current params from activity and pass to the appropriate fragment.
    */
-  public void takeParams(final @NonNull DiscoveryParams params, final int currentPosition) {
+  public void takeParams(final @NonNull List<Category> categories, final @NonNull DiscoveryParams params) {
     Observable.from(fragmentManager.getFragments())
       .ofType(DiscoveryFragment.class)
       .filter(frag -> {
         final int fragmentPosition = frag.getArguments().getInt(ArgumentsKey.DISCOVERY_SORT_POSITION);
-        return currentPosition == fragmentPosition;
+        return DiscoveryUtils.positionFromSort(params.sort()) == fragmentPosition;
       })
-      .subscribe(frag -> frag.updateParams(params));
+      .subscribe(frag -> frag.updateParams(params));  // todo: add categories next
   }
 
   /**
--- a/app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
+++ b/app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
@@ -50,16 +50,29 @@ public final class DiscoveryPagerAdapter extends FragmentPagerAdapter {
   }
 
   /**
+   * Passes along root categories to its fragment position to help fetch appropriate projects.
+   */
+  public void takeCategoriesForPosition(final @NonNull List<Category> categories, final int position) {
+    Observable.from(fragmentManager.getFragments())
+      .ofType(DiscoveryFragment.class)
+      .filter(frag -> {
+        final int fragmentPosition = frag.getArguments().getInt(ArgumentsKey.DISCOVERY_SORT_POSITION);
+        return fragmentPosition == position;
+      })
+      .subscribe(frag -> frag.takeCategories(categories));
+  }
+
+  /**
    * Take current params from activity and pass to the appropriate fragment.
    */
-  public void takeParams(final @NonNull List<Category> categories, final @NonNull DiscoveryParams params) {
+  public void takeParams(final @NonNull DiscoveryParams params) {
     Observable.from(fragmentManager.getFragments())
       .ofType(DiscoveryFragment.class)
       .filter(frag -> {
         final int fragmentPosition = frag.getArguments().getInt(ArgumentsKey.DISCOVERY_SORT_POSITION);
         return DiscoveryUtils.positionFromSort(params.sort()) == fragmentPosition;
       })
-      .subscribe(frag -> frag.updateParams(params));  // todo: add categories next
+      .subscribe(frag -> frag.updateParams(params));
   }
 
   /**
--- a/app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
@@ -239,29 +239,29 @@ public final class DiscoveryFragmentViewModel extends FragmentViewModel<Discover
   @Override public void paramsFromActivity(final @NonNull DiscoveryParams params) {
     paramsFromActivity.onNext(params);
   }
-  @Override public void projectCardViewHolderClick(final @NonNull ProjectCardViewHolder viewHolder, final Project project) {
+  @Override public void projectCardViewHolderClick(final @NonNull ProjectCardViewHolder viewHolder, final @NonNull Project project) {
     clickProject.onNext(project);
   }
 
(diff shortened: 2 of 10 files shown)


app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
```diff
@@ -48,14 +50,27 @@ public CharSequence getPageTitle(final int position) {
   }
 
   /**
+   * Passes along root categories to its fragment position to help fetch appropriate projects.
+   */
+  public void takeCategoriesForPosition(final @NonNull List<Category> categories, final int position) {
```
luoser: A review of how `fragmentManager` works: not all 5 sorts fragments are guaranteed to exist at any time. We pass along the position here from the Activity to ensure the categories are passed to a fragment that actually exists.

app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.java
```diff
@@ -223,7 +77,7 @@ public DiscoveryViewModel(final @NonNull Environment environment) {
       .mergeWith(topFilterRowClick)
       .map(NavigationDrawerData.Section.Row::params);
 
-    // Merge various param data sources to one true selected params value.
```
luoser: > one true selected params value

lol at myself because. We didn't need `params`, `paramsOnSortChange`, and `paramsOnDrawerSelection`. In fact, having those three separate `params` observables caused some values to over-emit.

We just need `params` to set DiscoveryActivity UI, and the `params` merged with sort position for the fragments.

app/src/main/java/com/kickstarter/viewmodels/outputs/DiscoveryViewModelOutputs.java
```diff
@@ -30,11 +31,17 @@
    * Emits when the params of a particular page should be updated. The page will be responsible for
    * taking those params and creating paginating projects from it.
    */
-  Observable<Pair<DiscoveryParams, Integer>> updateParamsForPage();
```
luoser: we no longer need to pass the `Integer` position here because with cleaner params, the `DiscoveryParams` contains the actual `sort` that can be used to derive the position!

app/src/test/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModelTest.java
```diff
@@ -39,11 +39,12 @@ public void testProjectsEmitWithNewCategoryParams() {
     final TestSubscriber<Boolean> hasProjects = new TestSubscriber<>();
     vm.outputs.projects().map(ListUtils::nonEmpty).subscribe(hasProjects);
 
-    // Load initial params from activity.
+    // Load initial params and root categories from activity.
     vm.inputs.paramsFromActivity(DiscoveryParams.builder().sort(DiscoveryParams.Sort.HOME).build());
+    vm.inputs.rootCategories(CategoryFactory.rootCategories());
 
     // Should emit current fragment's projects.
-    hasProjects.assertValues(false, true);
```
luoser: the projects no longer emit an initial empty array because there is no api delay with 2 requests: one for projects and one for categories

app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.java
```diff
@@ -334,7 +196,105 @@ public DiscoveryViewModel(final @NonNull Environment environment) {
...
+  private final PublishSubject<InternalBuildEnvelope> newerBuildIsAvailable = PublishSubject.create();
+  private final PublishSubject<Boolean> openDrawer = PublishSubject.create();
+  private final PublishSubject<Integer> pageChanged = PublishSubject.create();
+  private final PublishSubject<Integer> pagerCreatedPage = PublishSubject.create();
+  private final PublishSubject<NavigationDrawerData.Section.Row> parentFilterRowClick = PublishSubject.create();
+  private final PublishSubject<Void> profileClick = PublishSubject.create();
+  private final PublishSubject<Void> settingsClick = PublishSubject.create();
+  private final PublishSubject<NavigationDrawerData.Section.Row> topFilterRowClick = PublishSubject.create();
+
+  private final PublishSubject<List<Integer>> clearPages = PublishSubject.create();
```
luoser: some of the clarity from cleaning up these VMs is being able to ask ourselves questions like, "Why are some outputs `PublishSubjects`?" 🤔

app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
theginbin: cool!

app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
```diff
@@ -301,4 +181,87 @@ private void saveLastSeenActivityId(final @Nullable Activity activity) {
...
+    this.rootCategories.onNext(rootCategories);
+  }
+  @Override public void clearPage() {
+    clearPage.onNext(null);
+  }
+  @Override public void discoveryOnboardingViewHolderLoginToutClick(final @NonNull DiscoveryOnboardingViewHolder viewHolder) {
+    discoveryOnboardingLoginToutClick.onNext(true);
+  }
+  @Override public void nextPage() {
+    nextPage.onNext(null);
```
christopherwright: I'm down for compressing some of these three line methods onto one line, if they fit.

app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
```diff
@@ -301,4 +181,87 @@ private void saveLastSeenActivityId(final @Nullable Activity activity) {
...
+  }
+  @Override public void paramsFromActivity(final @NonNull DiscoveryParams params) {
+    paramsFromActivity.onNext(params);
+  }
+  @Override public void projectCardViewHolderClick(final @NonNull ProjectCardViewHolder viewHolder, final Project project) {
```
christopherwright: Should we add an annotation around nullability for project?

app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
```diff
@@ -301,4 +181,87 @@ private void saveLastSeenActivityId(final @Nullable Activity activity) {
...
+    clickProject.onNext(project);
+  }
+
+  @Override public Observable<Activity> activity() {
```
christopherwright: I think all the outputs could also have non null annotations on the return type.

app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.java
```diff
@@ -334,7 +196,105 @@ public DiscoveryViewModel(final @NonNull Environment environment) {
...
+  private final PublishSubject<DiscoveryParams> updateParamsForPage = PublishSubject.create();
+  private final BehaviorSubject<DiscoveryParams> updateToolbarWithParams = BehaviorSubject.create();
+
+  public final DiscoveryViewModelInputs inputs = this;
+  public final DiscoveryViewModelOutputs outputs = this;
+
+  @Override public void childFilterViewHolderRowClick(final @NonNull ChildFilterViewHolder viewHolder, final @NonNull NavigationDrawerData.Section.Row row) {
+    childFilterRowClick.onNext(row);
+  }
+  @Override public void discoveryPagerAdapterCreatedPage(final DiscoveryPagerAdapter adapter, final int position) {
```
christopherwright: Is the adapter non null?

app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
luoser: yes

app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.java
luoser: yes. i've been sloppy with my annotations here i shall do a pass

PR #48, opened 2017-01-17
Refactor Discovery to properly recover from stopped processes

luoser:
## what
When an app goes into the background for a while, the Android system reclaims memory by stopping idle processes. Our app was not recovering correctly when relaunching from a stopped process state because we were using incorrect methods from our `DiscoveryPagerAdapter` as `input`s which did not give our `DiscoveryFragment` a proper resume kick.

## before:
process-no-bueno

## how
1. Disentangle `DiscoveryViewModel` logic via #44.
2. Reliably reproduce the bug using Android's DDMS. This was the main challenge. [This SO answer](http://stackoverflow.com/questions/5287237/simulate-killing-of-activity-in-emulator/37978109#37978109) was very helpful.  
3. Debug the `skip(1)` and `take(1)`s that [we did not feel great about](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.java#L94-L97). Simplify overall `pagerSelectedPage` logic. Feel sound with our `RxJava` knowledge and operators/Subjects/Observables.
4. Investigate `FragmentPagerAdapter` API again. This is how we found a better pager method, [setPrimaryItem()](https://github.com/kickstarter/android-oss/commit/dc36a0b0ac4d9d0362db49f762057ffef9e4988d). Why this is better:
    - `getItem()` only pings once each time a `Fragment` is created. This would not ping on return from stopped state. See [FragmentPagerAdapter docs](https://developer.android.com/reference/android/support/v4/app/FragmentPagerAdapter.html) for more info.
    - `RxViewPager`'s `pageSelection`'s [immediate ping on subscribe](https://github.com/JakeWharton/RxBinding/blob/master/rxbinding-support-v4/src/main/java/com/jakewharton/rxbinding2/support/v4/view/RxViewPager.java#L25-L36) made it difficult to sync with when the pager was actually ready.
    - `setPrimaryItem()` tells us the current page, and when the Fragment is ready. .:., better!
5. Investigate [our concern with observing on the main thread within the VM](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.java#L90-L93). Turns out we do want a `0` delay for the `pagerAdapter` inputs, since there is a frame of delay between when the Fragment is created and ready to receive the inputs. We also [recently introduced the `.compose(observeForUI)` vs `.observeOn(AndroidSchedulers.mainThread())` pattern](https://github.com/kickstarter/android-private/pull/230), and now know more about when to use which.

## after:
process-bueno


willisplummer:
👏

app/src/main/java/com/kickstarter/libs/rx/transformers/Transformers.java
```diff
@@ -176,6 +176,9 @@ private Transformers() {}
   /**
    * If called on the main thread, schedule the work immediately. Otherwise delay execution of the work by adding it
    * to a message queue, where it will be executed on the main thread.
+   *
+   * This is useful for `RecyclerView`s where new data can appear immediately rather than waiting for a frame
```
christopherwright: I might slightly rephrase this –

> This is particularly useful for `RecyclerView`s; if subscriptions in these views are delayed for a frame, then the views temporarily shows recycled content and frame rate stutters. To address that, we can use `observeForUI` to execute the work immediately rather than wait for a frame.

PR 230 in the old repo goes into more detail.

PR #51, opened 2017-01-26
🎉 Add Either class, in Kotlin

luoser:
## heck yes!
Say hello to our first Kotlin class. 🤸‍♂️ 

## what
I ran into this 🤔 💭 moment when setting up our code for native comment support on project updates: since we have comments for *either* projects or updates, and Swift in fact handles this with [its own `Either` type](https://github.com/kickstarter/Kickstarter-Prelude/blob/master/Prelude/Either.swift), 💡, add one to android. Maybe even add it in Kotlin.

Things to note:
* [`companion object`](https://kotlinlang.org/docs/reference/object-declarations.html#companion-objects) allows for us to access `left` and `right` as static methods, while keeping our private constructor
    * this ensures that `Either` objects cannot be created with both values, i.e. 
    ```kotlin
    new Either<Int, String>(1, "don't do this!")
    ```
    * **[EDIT]** Using [sealed class](https://kotlinlang.org/docs/reference/classes.html#sealed-classes) instead, thanks for the suggestion @stephencelis  
* `!= null` checks are repeated rather than the defined `isLeft` and `isRight` functions to satisfy compiler optional warnings ¯\_(ツ)_/¯
* I stopped after adding `mapLeft` due to time, but coming soon will be `flatten()`, `flatMap()`, etc. when we have more time to figure out how to define these functions elegantly without enums

s/o to @mbrandonw for helping me explore this, and @christopherwright for the continued moral support

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/Either.kt
+++ b/app/src/main/java/com/kickstarter/libs/Either.kt
@@ -1,0 +1,33 @@
+package com.kickstarter.libs
+
+class Either<out A, out B> private constructor (val left: A?, val right: B?) {
+  companion object {
+    fun <A, B> left(left: A): Either<A, B> {
+      return Either(left, null)
+    }
+
+    fun <A, B> right(right: B): Either<A, B> {
+      return Either(null, right)
+    }
+  }
+
+  fun <C> either(ifLeft: (A) -> C, ifRight: (B) -> C): C {
+    if (this.left != null) {
+      return ifLeft(this.left)
+    }
+
+    if (this.right != null) {
+      return ifRight(this.right)
+    }
+
+    throw Exception("Exception: neither left nor right values found")
+  }
+
+  fun isLeft(): Boolean {
+    return this.left != null
+  }
+
+  fun isRight(): Boolean {
+    return this.right != null
+  }
+}
--- a/app/src/main/java/com/kickstarter/libs/Either.kt
+++ b/app/src/main/java/com/kickstarter/libs/Either.kt
@@ -30,4 +30,22 @@ class Either<out A, out B> private constructor (val left: A?, val right: B?) {
   fun isRight(): Boolean {
     return this.right != null
   }
+
+  /**
+   * Maps the right side of an `Either` value.
+   *
+   * @param transform:    A transformation
+   * @return              A new `Either` value.
+   */
+  fun <C> map(transform: (B) -> C): Either<A, C> {
+    if (this.left != null) {
+      return Companion.left(this.left)
+    }
+
+    if (this.right != null) {
+      return Companion.right(transform(this.right))
+    }
+
+    throw Exception()
+  }
 }
--- a/app/src/main/java/com/kickstarter/libs/Either.kt
+++ b/app/src/main/java/com/kickstarter/libs/Either.kt
@@ -15,12 +15,10 @@ class Either<out A, out B> private constructor (val left: A?, val right: B?) {
     if (this.left != null) {
       return ifLeft(this.left)
     }
-
     if (this.right != null) {
       return ifRight(this.right)
     }
-
-    throw Exception("Exception: neither left nor right values found")
+    throw Exception("Exception: Invalid left or right values.")
   }
 
   fun isLeft(): Boolean {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/libs/Either.kt
```diff
@@ -0,0 +1,63 @@
+package com.kickstarter.libs
+
+class Either<out A, out B> private constructor (val left: A?, val right: B?) {
```
stephencelis: What do you think of using sealed classes for this?

``` kotlin
sealed class Either<out L, out R> {
  class Left<out L, out R>(val left: L) : Either<L, R> {
    ...
  }
  class Right<out L, out R>(val right: R) : Either<L, R> {
    ...
  }
}

Either.Left("Error!")
Either.Right(1)

when(either) {
  is Left -> either.left
  is Right -> either.right
}
```

Seems to be a common pattern for tagged unions in Kotlin, and prevents even the private ability of instantiating an invalid type.

luoser: image

app/src/main/java/com/kickstarter/libs/Either.kt
```diff
@@ -0,0 +1,61 @@
+package com.kickstarter.libs
+
+sealed class Either<out A, out B> {
```
mbrandonw: should everything be marked as `public`?

app/src/test/java/com/kickstarter/libs/EitherTest.kt
```diff
@@ -0,0 +1,53 @@
...
+    val square: (Int) -> Int = { it * it }
+    assertEquals(
+      Either.Left<Int, String>(9).left,
+      Either.Left<Int, String>(3).mapLeft(square).left()
+    )
+  }
+
+  @Test fun testEither_Right() {
+    val intOrString = Either.Right<Int, String>("hello")
+    assertEquals("hello", intOrString.right)
```
mbrandonw: are the parens optional? should it not be `intOrString.right()`?

app/src/main/java/com/kickstarter/libs/Either.kt
luoser: for posterity `public` is the default modifier in Kotlin so we get redundant warnings

pakoito:
I have my reservations about the way Either is implemented. The use of isLeft() and isRight() allows for a scape hatch from static analysis which should be enforced. The encoding to enforce it is similar to the Java version seen here https://github.com/pakoito/RxSealedUnions

EDIT: okay, I'm late to the discussion and it'd be great if you gave that project's README a look and gave me an opinion.

stephencelis:
@pakoito Not sure I follow. Could you clarify? This is just a starting point and the API is based on [our Swift implementation](https://github.com/kickstarter/Kickstarter-Prelude/blob/d77f43dea770b2bb2c6554fe5364582a6369bf30/Prelude/Either.swift). The only nullables here come from the `left()` and `right()` fold operations, which we don't typically use outside of tests. In general we use `either(ifLeft, ifRight)` for folding.

pakoito:
Fair enough them, but I'd remove all of them.

stephencelis:
For what it's worth, our implementations are all based on [Haskell's Data.Either](https://hackage.haskell.org/package/base-4.9.0.0/docs/Data-Either.html), which also implement these functions. The types should be statically sound, but if we're missing anything, let us know!

pakoito:
map, mapRight, isLeft and isRight can all be expressed in terms of either(). I didn't know about Haskell's implementation, so probably I am being too anal about it.

app/src/main/java/com/kickstarter/libs/Either.kt
```diff
@@ -0,0 +1,61 @@
...
+  class Left<out L, out R>(internal val left: L) : Either<L, R>()
+  class Right<out L, out R>(internal val right: R) : Either<L, R>()
+
+  fun <C> either(ifLeft: (A) -> C, ifRight: (B) -> C): C = when(this) {
```
Miha-x64: Shouldn't this be `inline`? Also, this function may be called `apply`.

app/src/main/java/com/kickstarter/libs/Either.kt
```diff
@@ -0,0 +1,61 @@
+package com.kickstarter.libs
+
+sealed class Either<out A, out B> {
+  class Left<out L, out R>(internal val left: L) : Either<L, R>()
```
d3xter: I'd use **Nothing** as the unsued type in Left and Right,
e.g. _class Left<out L>(internal val left: L) : Either<L, Nothing>()_

This allows to create an instance of Left with _val left = Left(15)_ without specifying the right type.

PR #52, opened 2017-01-27
New take on VM structure

mbrandonw:
Right now we have 3 files for every view model: the inputs interface, outputs interface and the main view model. That's a bit of a bummer, and makes it difficult to see everything a VM is about in one file. We did this cause Java doesnt like multiple sibling classes/interfaces to sit in the same file :/

So, to fix, let's just add a namespace that will house the inputs/outputs/vm all at once. That's what I've done here. Whatdyall think?


app/src/main/java/com/kickstarter/viewmodels/ActivityFeed.java
```diff
@@ -0,0 +1,252 @@
...
+import rx.Observable;
+import rx.subjects.BehaviorSubject;
+import rx.subjects.PublishSubject;
+
+public interface ActivityFeed {
+
+  interface Inputs extends ActivityFeedAdapter.Delegate {
+    /**
+     * Invoke when pagination should happen.
+     */
```
luoser: maybe we should single-line these comments like so:
```java
/** Invoke when pagination should happen. */
```
that way we can keep this comment style if need be for longer comments, but keep these interfaces from getting too long

mbrandonw: totes, much better

PR #54, opened 2017-02-03
Filter backings from recommended projects

kyeah:
hey hi what up

## what is this

This adds `backed(-1)` to the recommendations discovery params.

## why would you do that

We filter user's backed projects from recommendations on web, and we also filter them in the native Thanks activity. Filtering them here will homogenize the backer experience across our platforms and stop people from asking jeff what's up with that rec engine.

The Thanks activity also filters backings from staff picks, so that's a possibility here.


luoser:
bitmoji

PR #55, opened 2017-02-03
Add IntentMapper banner helper and tests

luoser:
Support for our 🐨 app banner open event.

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/intentmappers/IntentMapper.java
+++ b/app/src/main/java/com/kickstarter/ui/intentmappers/IntentMapper.java
@@ -24,4 +24,14 @@ public abstract class IntentMapper {
 
     return Uri.parse(string);
   }
+
+  public static boolean intentFromAppBanner(final @NonNull Intent intent) {
+    final Uri uri = uri(intent);
+    if (uri == null) {
+      return false;
+    } else {
+      final String queryParam = uri.getQueryParameter("app_banner");
+      return queryParam != null && queryParam.equals("1");
+    }
+  }
 }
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/ui/intentmappers/IntentMapper.java
```diff
@@ -24,4 +24,14 @@
...
     return Uri.parse(string);
   }
+
+  public static boolean intentFromAppBanner(final @NonNull Intent intent) {
+    final Uri uri = uri(intent);
+    if (uri == null) {
+      return false;
+    } else {
+      final String queryParam = uri.getQueryParameter("app_banner");
+      return queryParam != null && "1".equals(queryParam);
```
christopherwright: Is the `queryParam != null` part here redundant?

app/src/main/java/com/kickstarter/ui/intentmappers/IntentMapper.java
```diff
@@ -24,4 +24,14 @@
 
     return Uri.parse(string);
   }
+
+  public static boolean intentFromAppBanner(final @NonNull Intent intent) {
```
christopherwright: The method name makes this read like it's returning an intent from a banner – might consider something like `appBannerIsSet`

app/src/main/java/com/kickstarter/ui/intentmappers/IntentMapper.java
luoser: yes

PR #56, opened 2017-02-06
Modernize Project VM

luoser:
## what
More groundwork for a more concise update comments PR coming 🔜 .

* Rename `inputs` and `outputs` to have more explicit names
* Implementing new View Model structure (consolidates `Inputs` and `Outputs` to one interface file containing the `ViewModel` class) (re: #52)

**Note**: there is no logic change in this PR, only renaming and restructuring.


luoser:
re: squash we are going to go with a `ProjectViewModel.ViewModel` refactor instead 😬 
up for name change suggestions in the future!

christopherwright:
Yeah I feel like this is a 50/50 call. It is nice not having to jump between files, and it makes it clear that these things are all one unit – but accessing this via `ProjectViewModel.ViewModel` is pretty awkward. Also means that our vm files grow pretty lengthy, so it becomes harder to jump around within the file. Am happy to go with whatever y'all are most comfortable with though.

mbrandonw:
> but accessing this via ProjectViewModel.ViewModel is pretty awkward.

i think the only time we do that is with the `@ RequiresActivityViewModel`, so it's not something you construct very often. that was our main justification.

luoser:
it's accessed in the `@Requires...`, and also every time we instantiate a vm in a test ( ͡↑ ͜ʖ ͡↑)

re: lengthy files, that is indeed the tradeoff. files like `RewardViewModel` will have a very long header, but I have to say it has been pretty nice working with inputs and outputs in one place, and seeing the inputs and outputs immediately when opening a ViewModel file feels like a nice abstract of what is going to happen

PR #57, opened 2017-02-06
Add AnimationUtils

luoser:
## what
Lil helper class to clean up some logic we've repeated in a few activities (and will repeat in a few activities to come). Open to renaming the `AnimationUtils` I just figured it was the most honest.


app/src/main/java/com/kickstarter/libs/utils/AnimationUtils.kt
```diff
@@ -0,0 +1,23 @@
+package com.kickstarter.libs.utils
+
+import android.view.animation.AlphaAnimation
+import android.view.animation.Animation
+
+object AnimationUtils {
+
+  @JvmOverloads
+  fun disappearAnimation(duration: Long = 300L): Animation {
```
mbrandonw: so siiiiick!

for posterity: this `@JvmOverloads` makes kotlin export two functions, one with the default and one without.

PR #58, opened 2017-02-09
Enable comments on updates

luoser:
## what
The base ability to load and compose comments natively on project updates. 

update-comments-1

Turns out there is a _bunch_ of web view navigation work we need in place for an optimal experience which should merge in before release, alas here is the first step.

Going to make a helper for testing here: https://github.com/kickstarter/kickstarter/pull/5737, but in the meanwhile feel free to test on `ios3launch` HQ.

## what's next
* Instrumentation
* Additional index url logic
* External url logic
* Web view optimization: navigating back,


app/src/main/java/com/kickstarter/libs/Koala.java
```diff
@@ -356,6 +356,11 @@ public void trackVideoStart(final @NonNull Project project) {
     client.track("Project Video Start", KoalaUtils.projectProperties(project));
   }
 
+  // PROJECT UPDATES
+  public void trackViewedUpdates() {
+    client.track("Viewed Updates");
```
luoser: yes

PR #59, opened 2017-02-15
Instrument updates, add KoalaContext

luoser:
## what
Adding and implementing new events for updates, which includes a new `KoalaContext` `enum class` in Kotlin.

### here's why i chose `enum class`
_tl;dr using `sealed class` would be a smidge faster but `enum class` is much cleaner. both sealed and enum are faster than static java classes._

We typically use `enum`s sparingly in Java since they are costly and don't provide us with much more functionality than if we were to use `static class`es to hold global values. We use enums quite a bit in Swift because they are first class and have type inference, i.e. [a perfect use case for koala contexts](https://github.com/kickstarter/ios-oss/blob/master/Library/Koala/Koala.swift#L126-L136).

Kotlin seemed like a natural choice for recreating our contexts, and there are in fact `enum class`es that provide the exact functionality that we would like. Docs:

https://kotlinlang.org/docs/reference/enum-classes.html

I was skeptical about performance so I did some research and manual benchmarking to see how `enum class` performance would hold up to [`static class`es in Java](https://github.com/kickstarter/android-oss/blob/1d87b123407b85348f24fd1d32139a969795fcbc/app/src/main/java/com/kickstarter/libs/KoalaContext.java) and `sealed class`es as suggested by [a few posts on the internet](https://kotlin.link/articles/Swift-vs-Kotlin-for-real-iOS-Android-apps.html).

An example of a `sealed class` KoalaContext implementation is:

```kotlin
sealed class KoalaContext {
  sealed class Comments {
    object PROJECT : Comments()
    object UPDATE : Comments()

    fun trackingString(): String = when(this) {
      is PROJECT -> "project"
      is UPDATE -> "update"
    }
  }
}

// accessed in Java as:
KoalaContext.Comments.PROJECT.INSTANCE;
```

(Note: using the convert-to-Kotlin tool on the static java class returned a bunch of nested objects that looked something like this: 
```kotlin
object KoalaContext{
  object Comments {
    val PROJECT = "project"
    val UPDATE = "updates"
  }
}

// accessed in Java as as:
KoalaContext.Comments.INSTANCE.getPROJECT();
```
but the type would ultimately be a String.)

In the end, when "benchmarking" (measuring the time it takes to print each sealed, static, and enum `Comments.Project` context value 100_000 times), resulted in:
```
sealed class: 		28373ms 🥇
static class: 		30956ms 🥉
enum class: 		29766ms 🥈 
```

Yep, the `sealed class` approach is technically the winner, but in terms of readability and ease of use I think it's worth continuing with `enum class`. Plus it's faster than using static java classes anyways! 

The downsides are of course the slight performance ding and the inability to mix associated value types (i.e. if we also want an enum class to hold String and Int values), but that's something we can keep in mind for future implementations (all of our Koala contexts are strings).


app/src/main/java/com/kickstarter/models/Update.java
```diff
@@ -18,6 +18,7 @@
   public abstract @Nullable Integer commentsCount();
   public abstract @Nullable Boolean hasLiked();
   public abstract long id();
+  public abstract boolean isPublic();
```
christopherwright: Is this setting the right value from the API, or is it always just initialized to `false`? Our serializer names this field `public` rather than `is_public`, and I couldn't see if we're handling that here.

app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java
```diff
@@ -200,6 +185,58 @@ public CommentsViewModel(final @NonNull Environment environment) {
       .take(1)
       .compose(bindToLifecycle())
       .subscribe(__ -> refresh.onNext(null));
+
+    final Observable<Update> update = projectOrUpdate.map(Either::right);
+
+    // Todo: add a pageCount to RecyclerViewPaginator to track loading newer comments.
+    Observable.combineLatest(project, update, Pair::create)
+      .compose(takeWhen(nextPage.skip(1)))
```
christopherwright: Are we skipping because `nextPage` emits for the initial load?

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
```diff
@@ -59,7 +59,9 @@ public void testClickingInterfaceElements() {
     goToLogin.assertNoValues();
     goToProject.assertNoValues();
     goToProjectUpdate.assertNoValues();
+    koalaTest.assertValues("Activity View");
```
christopherwright: It might be nice to pull out these strings into one place rather than duplicating them across the codebase

luoser: agreed!

app/src/main/java/com/kickstarter/models/Update.java
luoser: oh dang good catch. this will always be `false` since indeed, `public` rather than `is_public` is what's serialized. 

i missed this when looking at [our swift Update model]( https://github.com/kickstarter/ios-ksapi/blob/master/KsApi/models/Update.swift#L11) and neglecting the [actual decoding part](https://github.com/kickstarter/ios-ksapi/blob/master/KsApi/models/Update.swift#L46)

app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java
luoser: 🤔 this is what I assumed (admittedly I didn't change the tracking logic itself when refactoring here). funny enough `nextPage` does not emit immediately on initial load, but it does emit when there are no comments. will look into `RecyclerViewPaginator` to check out this logic

luoser: https://github.com/kickstarter/android-oss/pull/60

app/src/main/java/com/kickstarter/models/Update.java
luoser: ran into an unexpected wall--looks like there's no straightforward way of deserializing this reserved name (`"public"`) while using `@AutoParcel` so i may just leave this field out until we actually need it so I can move on 😬

discussion: https://github.com/frankiesardo/auto-parcel/issues/6

app/src/main/res/values/strings_koala.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+  <string name="Activity_Load_More">Activity Load More</string>
```
christopherwright: Let's have a lil chat on these fellas on Tuesday. I think a super simple approach would be to just have a class with a bunch of static strings. We won't have a need to localize these events, and using static strings means we avoid the need to pass around a context.

luoser: ah for sure--definitely could've gone the `IntentKey` route here

app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java
```diff
@@ -200,6 +186,65 @@ public CommentsViewModel(final @NonNull Environment environment) {
       .take(1)
       .compose(bindToLifecycle())
       .subscribe(__ -> refresh.onNext(null));
+
+    final Observable<Update> update = projectOrUpdate.map(Either::right);
+
+    // NB: A temporary workaround for nextPage ping on empty state. Better fix is to filter out
+    // non-scrollable views in `RecyclerViewPaginator`.
+    final Observable<Void> nonEmptyNextPage = comments
```
luoser: this is the temp solution for empty page `nextPage` pings until #60 merges, as that needs more testing but is not a blocker to merging this.

sockeqwe:
Hey @luoser, 
is the benchmark sealed class vs. enum vs. static field open source too? Didnt expected such a result, would like to take a closer look. Don't get me wrong, I do believe that your results are correct although I'm surprised. I would like to take a look at kotlins byte code tweaks for sealed classes vs. static fields in a clean environment (which I'm obviously​ too lazy to write my own tests / benchmark for :smile:)

luoser:
Hi @sockeqwe, the results were indeed surprising. 

The "benchmark" test I ran does come with a caveat since I lazily ran them in our app--not an ideal clean environment to really analyze the differences between the Kotlin byte code differences. 

[Please excuse the noise in this branch](https://github.com/kickstarter/android-oss/compare/context-benchmarking), but it contains the code I used to measure the times, namely 

1. [this is the sealed class context](https://github.com/kickstarter/android-oss/blob/b1b0f2ba9067ce5e8afdc953792ec47033accf8d/app/src/main/java/com/kickstarter/libs/ConvertedKoalaContext.kt) 
2. [this is the enum class context](https://github.com/kickstarter/android-oss/blob/b1b0f2ba9067ce5e8afdc953792ec47033accf8d/app/src/main/java/com/kickstarter/libs/KoalaContext.kt) 
3. [this is the static class context](https://github.com/kickstarter/android-oss/blob/b1b0f2ba9067ce5e8afdc953792ec47033accf8d/app/src/main/java/com/kickstarter/libs/StaticKoalaContext.java)
4. and [this is the very lazy benchmarking code](https://github.com/kickstarter/android-oss/blob/b1b0f2ba9067ce5e8afdc953792ec47033accf8d/app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.java#L155-L184) that i actually ended up running 100_000 times rather than 1_000_000

Definitely not the most scientific approach, and I would also love to do a more in-depth analysis at some point :neckbeard:

Dk35840:
Unsubscribed me

sockeqwe:
Thank you so much @luoser !

PR #60, opened 2017-02-17
Improve nextPage pagination signal

luoser:
## what
Realized while working on #59 that our `nextPage` input was incorrectly triggering on empty views e.g. empty ActivityFeed, Comments, Profile.

[This line in `RecyclerViewPaginator`](https://github.com/kickstarter/android-oss/compare/the-real-next-page?expand=1#diff-249dfc88ace9282edb448638e270ace4R37) attempts to filter out empty RecyclerViews by filtering for layout manager items != 0, however this is insufficient/pointless since our recycler views actually hold other views along with recycled views.

In this specific case, we realized there was a suspicious `nextPage.skip(1)` happening in `CommentsViewModel`, which skipped tracking the paginate event. On top of that we found that `nextPage` pinged for the empty state since the "empty" recycler view still contained 2 views.

## the fix
We can filter for a boolean on the recycler view that tells us if the view is scrollable:

```java
 .filter(__ -> recyclerView.canScrollVertically(DIRECTION_DOWN))
```

This filters for an actual scrollable `recyclerView`, which is what we want since we only care about scroll events in the paginator.


app/src/main/java/com/kickstarter/libs/ApiPaginator.java
```diff
@@ -34,6 +34,7 @@
   private final @NonNull Func1<List<Data>, List<Data>> pageTransformation;
   private final boolean clearWhenStartingOver;
   private final @NonNull Func2<List<Data>, List<Data>, List<Data>> concater;
+  private final boolean distinctUntilChanged;
```
luoser: @mbrandonw lmk if i am adding this (Android paginator version of `skipRepeats`) correctly here

mbrandonw: looks good to me!

PR #61, opened 2017-02-21
Improve webviews for updates

luoser:
## what
A few web view improvements as the last piece of implementing Update comments:

* Koala event for opening external links
* New request handler for additional update index urls in the ProjectUpdatesActivity/VM
* New request handler for additional update urls in the UpdateActivity/VM
* Refactored `back()` logic for web views

And tests and bonus view model cleanups, etc. I'll point out interesting things below via. comments.

### DIFF
--- a/app/src/main/java/com/kickstarter/services/KSUri.java
+++ b/app/src/main/java/com/kickstarter/services/KSUri.java
@@ -21,10 +21,6 @@ public final class KSUri {
     return isKickstarterUri(uri, webEndpoint) && Secrets.RegExpPattern.API.matcher(uri.getHost()).matches();
   }
 
-  public static boolean isCookiesUri(final @NonNull Uri uri, final @NonNull String webEndpoint) {
-    return isKickstarterUri(uri, webEndpoint) && uri.getPath().equals("/cookies");
-  }
-
   public static boolean isDiscoverCategoriesPath(final @NonNull String path) {
     return DISCOVER_CATEGORIES_PATTERN.matcher(path).matches();
   }
@@ -36,10 +32,6 @@ public final class KSUri {
 
   public static boolean isDiscoverPlacesPath(final @NonNull String path) {
     return DISCOVER_PLACES_PATTERN.matcher(path).matches();
-  }
-
-  public static boolean isHelloUri(final @NonNull Uri uri, final @NonNull String webEndpoint) {
-    return isKickstarterUri(uri, webEndpoint) && uri.getPath().equals("/hello");
   }
 
   public static boolean isHivequeenUri(final @NonNull Uri uri, final @NonNull String webEndpoint) {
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/kickstarter/ui/activities/ProjectUpdatesActivity.java
```diff
@@ -85,6 +97,11 @@ private boolean handleProjectUpdateUriRequest(final @NonNull Request request, fi
     return true;
   }
 
+  private boolean handleProjectUpdatesUriRequest(final @NonNull Request request, final @NonNull WebView webView) {
+    this.viewModel.inputs.goToUpdatesRequest(request);
+    return false;
```
luoser: this handler, as well as [`handleProjectUpdateUriRequest` in UpdateViewModel](https://github.com/kickstarter/android-oss/pull/61/files#diff-e5291b7fe239077b3e0a95cab8dd7691R114), return `false` rather than true because we don't actually want the web view to handle the request--we just want to use this method as a callback to let our view model know that the relevant url request has been intercepted

app/src/main/java/com/kickstarter/ui/activities/WebViewActivity.java
```diff
@@ -47,12 +46,19 @@ protected void onCreate(final @Nullable Bundle savedInstanceState) {
       .subscribe(webView::loadUrl);
   }
 
-  @OnClick(R.id.back_button)
-  public void backButtonClick() {
-    back();
+  @Override
+  public void back() {
+    if (webView.canGoBack()) {
```
luoser: from the docs: https://developer.android.com/guide/webapps/webview.html#HandlingNavigation

seems pretty chill to either navigate back through the webview stack, or invoke our `BaseActivity`'s `back()` method which handles general back navigation

app/src/main/java/com/kickstarter/ui/activities/WebViewActivity.java
```diff
@@ -47,12 +46,19 @@ protected void onCreate(final @Nullable Bundle savedInstanceState) {
       .subscribe(webView::loadUrl);
   }
 
-  @OnClick(R.id.back_button)
```
luoser: this `@OnClick(R.id.back_button)` was redundant since this method exists in our base `KSToolbar`

app/src/main/java/com/kickstarter/viewmodels/UpdateViewModel.java
```diff
@@ -60,49 +67,81 @@ public ViewModel(final @NonNull Environment environment) {
...
-    private final PublishSubject<Request> goToCommentsRequestSubject = PublishSubject.create();
-    private final PublishSubject<Request> goToProjectRequestSubject = PublishSubject.create();
-    private final PublishSubject<Void> shareButtonClickedSubject = PublishSubject.create();
+    /**
+     * Parses a request for project and update params.
+     *
+     * @param request   Comments or update request.
+     * @return          Pair of project param string and update param string.
+     */
+    private @NonNull Pair<String, String> projectUpdateParams(final @NonNull Request request) {
```
luoser: this method is repeated from `ProjectUpdatesViewModel`, with the same todo of making a better, universal `Navigation` class in the future to handle these requests :/

app/src/test/java/com/kickstarter/viewmodels/UpdateViewModelTest.java
```diff
@@ -21,17 +22,46 @@
 import rx.observers.TestSubscriber;
 
 public final class UpdateViewModelTest extends KSRobolectricTestCase {
+  private final Intent defaultIntent = new Intent()
```
luoser: something new I'm trying out here--maybe it's nice to have a default `intent` in our view model tests. this eliminates the need to repeatedly create new intents (e.g. `vm.intent(new Intent()....)`) when starting a test that doesn't need to keep a reference to the intent objects

it would also give us an immediate reminder / some guidance as to what our view model's intent is and should be initialized with.

app/src/main/java/com/kickstarter/services/KSUri.java
```diff
@@ -86,6 +86,10 @@ public static boolean isProjectUpdateUri(final @NonNull Uri uri, final @NonNull
     return isKickstarterUri(uri, webEndpoint) && PROJECT_UPDATE_PATTERN.matcher(uri.getPath()).matches();
   }
 
+  public static boolean isProjectUpdatesUri(final @NonNull Uri uri, final @NonNull String webEndpoint) {
```
luoser: yep good call

PR #63, opened 2017-02-28
Refactor Project VM

luoser:
## wut
More project view model modernization / cleanup:
* Removed our weird ` private final BehaviorSubject<Project> project = BehaviorSubject.create();` that was neither an input nor an output
* Made all outputs `Observable`s
* Added `currentProject` logic (I think back then we just didn't use `Observable.merge(...)` yet) to distinguish from `initialProject` logic as the project changes on login / starring
* Updated 🐨 events
* Renamed outputs to be more explicit. A lot of noise is from this, sry not sry

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.java
@@ -6,7 +6,6 @@ import android.support.annotation.Nullable;
 import android.util.Pair;
 
 import com.kickstarter.libs.ActivityViewModel;
-import com.kickstarter.libs.Config;
 import com.kickstarter.libs.CurrentConfigType;
 import com.kickstarter.libs.CurrentUserType;
 import com.kickstarter.libs.Environment;
@@ -25,7 +24,6 @@ import com.kickstarter.ui.viewholders.ProjectViewHolder;
 import java.net.CookieManager;
 
 import rx.Observable;
-import rx.subjects.BehaviorSubject;
 import rx.subjects.PublishSubject;
 
 import static com.kickstarter.libs.rx.transformers.Transformers.combineLatestPair;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.java
```diff
@@ -25,10 +25,10 @@
 import java.net.CookieManager;
 
 import rx.Observable;
-import rx.subjects.BehaviorSubject;
```
luoser: bitmoji

app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.java
```diff
@@ -170,24 +172,35 @@ public ViewModel(final @NonNull Environment environment) {
...
+
+      this.playVideo = currentProject.compose(takeWhen(this.playVideoButtonClicked));
+      this.showShareSheet = currentProject.compose(takeWhen(this.shareButtonClicked));
+      this.startCampaignWebViewActivity = currentProject.compose(takeWhen(this.blurbTextViewClicked));
+      this.startCheckoutActivity = currentProject.compose(takeWhen(this.backProjectButtonClicked));
+      this.startCreatorBioWebViewActivity = currentProject.compose(takeWhen(this.creatorNameTextViewClicked));
+      this.startCommentsActivity = currentProject.compose(takeWhen(this.commentsTextViewClicked));
+      this.startManagePledgeActivity = currentProject.compose(takeWhen(this.managePledgeButtonClicked));
+      this.startProjectUpdatesActivity = currentProject.compose(takeWhen(this.updatesTextViewClicked));
+      this.startViewPledgeActivity = currentProject.compose(takeWhen(this.viewPledgeButtonClicked));
```
luoser: if you're wondering where all these "new" outputs are coming from they are not new we used to do all this logic within the getter functions with our `this.project` BehaviorSubject 😱

app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.java
```diff
@@ -121,73 +121,81 @@ public ViewModel(final @NonNull Environment environment) {
...
-        .subscribe(__ -> this.showLoginTout.onNext(null));
-
-      this.shareButtonClickedSubject
+      final Observable<Project> currentProject = Observable.merge(
+        initialProject,
+        projectOnUserChangeStar,
+        starredProjectOnLoginSuccess
+      );
+
+      this.showStarredPrompt = projectOnUserChangeStar.mergeWith(starredProjectOnLoginSuccess)
```
christopherwright: I noticed in the release app that this prompt isn't showing – is that an issue for you too? Does this change fix anything there?

app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.java
christopherwright: 🥇

app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.java
luoser: it should still be showing, maybe it was a project ending within 48 hours?

we don't show the toast for that since it says `"We'll remind you 48 hours before this project ends!"` which also makes me realize that we should add logic in the future to show a different toast for that case

PR #67, opened 2017-03-16
Tidy the build by extracting lib versions

jonathan-caryl:
Make it easier to update SDK versions by extracting them into variables.

Define the SDK versions in the root project, so they're available in any future library modules that may be added.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -131,40 +131,41 @@ repositories {
 
 dependencies {
     compile fileTree(include: ['*.jar'], dir: 'libs')
-    compile 'com.android.support:appcompat-v7:25.1.0'
-    compile 'com.android.support:cardview-v7:25.1.0'
-    compile 'com.android.support:design:25.1.0'
+    compile "com.android.support:appcompat-v7:$supportLibVersion"
+    compile "com.android.support:cardview-v7:$supportLibVersion"
+    compile "com.android.support:design:$supportLibVersion"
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -131,40 +131,41 @@ repositories {
...
     compile 'com.jakewharton.timber:timber:3.0.1'
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     compile 'com.stripe:stripe-android:1.0.3'
-    debugCompile 'com.squareup.leakcanary:leakcanary-android:1.3.1'
-    releaseCompile 'com.squareup.leakcanary:leakcanary-android-no-op:1.3.1'
-    compile 'com.squareup.okhttp3:logging-interceptor:3.1.2'
-    compile 'com.squareup.okhttp3:okhttp-urlconnection:3.1.2'
+    debugCompile "com.squareup.leakcanary:leakcanary-android:$leakCanaryVersion"
+    releaseCompile "com.squareup.leakcanary:leakcanary-android-no-op:$leakCanaryVersion"
+    compile "com.squareup.okhttp3:okhttp:$okhttp3Version"
```
jonathan-caryl: Both `logging-interceptor` and `okhttp-urlconnection` depend on `okhttp`, it seems odd not to list it explicitly.

luoser:
thanks for the suggestion! however our team prefers not having to do an additional lookup for each dependency and is fine with seeing the build versions listed next to them

PR #68, opened 2017-03-16
fix path for tools:context

jonathan-caryl:
A few of the layout files have an incorrect path for `tools:context`. Fix them, to keep Android Studio happy.


luoser:
thanks!

jonathan-caryl:
thanks @luoser 

I opened PR #69 a while back too, it'd be great if that could be merged, or closed if it's not wanted.

PR #69, opened 2017-03-16
Switch to TextInputEditText, add password hint on login

jonathan-caryl:
Using `TextInputEditText` inside `TextInputLayout` in place of a regular `EditText` allows better behaviour if used in extract mode, and is good practice.

Also, password hint is now supported by `TextInputLayout` so enable that.

### DIFF
--- a/app/src/debug/res/values/ids.xml
+++ b/app/src/debug/res/values/ids.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <item name="visible" type="id"/>
+    <item name="masked" type="id"/>
+</resources>
--- a/app/src/main/res/layout/login_form_view.xml
+++ b/app/src/main/res/layout/login_form_view.xml
@@ -1,26 +1,28 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout
-  xmlns:android="http://schemas.android.com/apk/res/android"
-  xmlns:app="http://schemas.android.com/apk/res-auto"
-  android:orientation="vertical"
-  android:layout_width="match_parent"
-  android:layout_height="fill_parent"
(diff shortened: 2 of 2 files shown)


app/src/debug/res/values/ids.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <item name="visible" type="id"/>
```
jonathan-caryl: Without these IDs defined the layout preview doesn't display correctly in Android Studio. They aren't necessary at runtime, it seems to make more sense to restrict them to debug builds only.

app/src/main/res/layout/login_form_view.xml
```diff
@@ -1,43 +1,46 @@
...
+    android:gravity="center_horizontal"
+    tools:showIn="@layout/login_layout">
 
   <android.support.design.widget.TextInputLayout
     app:hintTextAppearance="@style/BodySecondary"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:layout_marginTop="@dimen/grid_2">
+    android:layout_marginTop="@dimen/grid_2"
+    android:hint="@string/login_placeholder_email" >
```
jonathan-caryl: You can specify the `hint` in the child `EditText` but if you look at the source for `TextInputLayout` you can see that

        if (mHintEnabled && TextUtils.isEmpty(mHint)) {
            setHint(mEditText.getHint());
            // Clear the EditText's hint as we will display it ourselves
            mEditText.setHint(null);
        }

so it makes more sense to set it directly on the `TextInputLayout`

luoser:
Sorry for not responding earlier, but the existing layout on master does have a hint set for login. Noted for future development however that we will continue to use `TextInputLayout`. Thanks!

PR #72, opened 2017-03-23
Fix missing override annotations

egugue:
I fixed missing `override` annotations which should be added at `exitTransition` method.


luoser:
thanks!

PR #73, opened 2017-03-23
[Messages] Message threads

luoser:
## what
Part 1 is implementing the `MessageThreads` layer, including:
* Entry button from Profile
* `MessageThreads` Activity, VM, VMTest
* `MessageThreadsAdapter`
* `MessageThread` ViewHolder, VM, VMTest
* all relevant xml files

## design
Designs are not finalized yet so don't worry about UI here pls, just the models, data flow, etc. There will be subsequent PRs for tapping into Message threads, composing messages, and design finalization.

## a bug
I found a gnarly little bug when implementing the `SwipeRefresher` layout that @mbrandonw will work on fixing. Note that if you run this and pull to refresh, the refresher will spin but not go away. This won't be fixed until this `ApiPaginator` bug is fixed.


app/build.gradle
```diff
@@ -63,6 +63,7 @@ android {
             }
         }
         debug {
+            minifyEnabled false
```
luoser: we found this change to help make the debugger a little more reliable...breakpoints weren't hitting consistently, but something else we can do is upgrade AS 😁

app/src/main/java/com/kickstarter/ui/viewholders/MessageThreadViewHolder.java
```diff
@@ -0,0 +1,86 @@
...
+      .subscribe(this.participantNameTextView::setText);
+
+    this.viewModel.outputs.unreadIndicatorImageViewHidden()
+      .compose(bindToLifecycle())
+      .compose(observeForUI())
+      .subscribe(ViewUtils.setGone(this.unreadIndicatorImageView));
+  }
+
+  @Override
+  public void bindData(final @Nullable Object data) throws Exception {
```
mbrandonw: feel we should rename this to align with how we think about things now.

no need to do now... just saying for future

app/src/main/java/com/kickstarter/viewmodels/MessageThreadsViewModel.java
```diff
@@ -0,0 +1,119 @@
...
+      this.unreadCountTextViewText = this.currentUser.loggedInUser()
+        .map(User::unreadMessagesCount)
+        .filter(count -> ObjectUtils.isNotNull(count) && IntegerUtils.isNonZero(count))
+        .map(NumberUtils::format);
+
+      this.unreadCountTextViewHidden = this.currentUser.loggedInUser()
+        .map(user -> {
+          final Integer count = user.unreadMessagesCount();
+          return ObjectUtils.isNull(count) || IntegerUtils.isZero(count);
+        });
```
mbrandonw: a small refactor to `this.unreadCountTextViewText` and `this.unreadCountTextViewHidden` might first compute a lil helper observable that could be used more expressively:

```java
final Observable<Integer> unreadCount = this.currentuser.loggedInUser
  .map(User::unreadMessagesCount)
  .compose(Transformers.coalesce(0))

this.unreadCountTextViewText = unreadCount.map(NumberUtils::format)
this.unreadCountTextViewHidden = unreadCount.map(c -> c == 0)
```

whatdya think?

app/src/main/java/com/kickstarter/services/ApiClient.java
```diff
@@ -32,11 +32,13 @@
 import com.kickstarter.services.apiresponses.CategoriesEnvelope;
 import com.kickstarter.services.apiresponses.CommentsEnvelope;
 import com.kickstarter.services.apiresponses.DiscoverEnvelope;
+import com.kickstarter.services.apiresponses.MessageThreadsEnvelope;
 import com.kickstarter.services.apiresponses.StarEnvelope;
 
 import java.util.Arrays;
 import java.util.List;
 
+import retrofit2.Response;
```
luoser: yeah, since [this line](https://github.com/kickstarter/android-oss/pull/73/files/9a79e5c0bc5ee52805653e10ffb7b611818fd4ea#diff-33e9703158d672ad42df5e7fa677e9a1R208) is the first time we're using an intermediate variable for some additional logic on the `Response`

app/src/main/java/com/kickstarter/viewmodels/MessageThreadsViewModel.java
luoser: oh yeah i actually like that a lot better, since the `coalesce` will take care of the `null` checks as well. thanks!

PR #75, opened 2017-03-29
[Messages] Message thread messages

luoser:
## what
The messages of a message thread layer of messages. lol

 

The only functionality here is seeing the messages of the thread. Designs are still skeletal so pls refrain from dissing my xml, all other buttons and the reply field are not yet implemented. Nothing fancy happening here either, just your typical Activities/ViewHolders/ViewModels/Tests!


app/src/main/res/values/dimens.xml
```diff
@@ -24,6 +24,15 @@
   <dimen name="grid_35">280dp</dimen>
   <dimen name="grid_40">320dp</dimen>
 
+  <!-- Kickstarter six-point grid keylines -->
```
luoser: ehh, something that would be nice to figure out maybe before merging this is if hot swapping out our legacy 8pt grid system with 6pt values would mess everything up....maybe i'll experiment with that while this is being reviewed 🤔

luoser: ok i just did a quick dry run of switching to a 6pt grid and it's not something that should block messages, since it'll take more time to phase in the new grid i'll chat with @maggieKSR about it 

for the mean time i'm okay with appending new grid values with `new_` that way when we search and replace it'll be easy to find them

mbrandonw: the other option is to use 8pt grid for inbox until we switch everything to 6pt. do you think that's any better than `_new_` grids? 

i defer to yer judgement!

app/src/main/java/com/kickstarter/viewmodels/MessageHolderViewModel.java
```diff
@@ -0,0 +1,119 @@
...
+
+  interface Outputs {
+    /** Emits a boolean that determines whether the creator's avatar image should be hidden. */
+    Observable<Boolean> creatorAvatarImageHidden();
+
+    /** Emits the url for the creator's avatar image. */
+    Observable<String> creatorAvatarImageUrl();
+
+    /** Emits a boolean to determine whether the message body view should align the parent view's end and right. */
+    Observable<Boolean> messageBodyTextViewAlignParentEnd();
```
mbrandonw: that's a fun output!

app/src/main/java/com/kickstarter/viewmodels/MessageHolderViewModel.java
```diff
@@ -0,0 +1,119 @@
...
+
+    public ViewModel(final @NonNull Environment environment) {
+      super(environment);
+
+      this.client = environment.apiClient();
+      this.currentUser = environment.currentUser();
+
+      final Observable<User> freshUser = this.client.fetchCurrentUser()
+        .retry(2)
+        .compose(neverError());
```
mbrandonw: heyooo, don't think we should be refreshing the user from a view holder! this would execute multiple times based on how many messages are visible at once.

whatcha tryna do with this jazz?

app/src/main/java/com/kickstarter/viewmodels/MessageHolderViewModel.java
```diff
@@ -0,0 +1,119 @@
...
+public interface MessageHolderViewModel {
+
+  interface Inputs {
+    /** Call to configure the view model with a message. */
+    void configureWith(Message message);
+  }
+
+  interface Outputs {
+    /** Emits a boolean that determines whether the creator's avatar image should be hidden. */
+    Observable<Boolean> creatorAvatarImageHidden();
```
mbrandonw: i think this output should be something like `participantAvatarImageHidden`, since the other person in the convo won't necessarily be the creator (like in the future when we have creator tools).

same for the other creator output...

app/src/main/java/com/kickstarter/ui/activities/MessagesActivity.java
```diff
@@ -0,0 +1,116 @@
...
+
+@RequiresActivityViewModel(MessagesViewModel.ViewModel.class)
+public final class MessagesActivity extends BaseActivity<MessagesViewModel.ViewModel> {
+  private KSCurrency ksCurrency;
+  private KSString ksString;
+  private MessagesAdapter adapter;
+
+  protected @Bind(R.id.backing_amount_text_view) TextView backingAmountTextViewText;
+  protected @Bind(R.id.backing_info_view) View backingInfoView;
+  protected @Bind(R.id.messages_creator_name_text_view) TextView creatorNameTextView;
```
mbrandonw: think we need to rename all of the creator specific stuff to just "participant" in order to leave room for future feature development

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -0,0 +1,155 @@
...
+        .filter(mu -> mu.first.backing() == null)
+        .switchMap(mu ->
+          mu.first.project().isBacking()
+            ? this.client.fetchProjectBacking(mu.first.project(), mu.second).materialize()
+            : this.client.fetchProjectBacking(mu.first.project(), mu.first.participant()).materialize()
+        )
+        .share();
+
+      final Observable<Backing> backing = Observable.merge(
+        messageThread.map(MessageThread::backing).filter(ObjectUtils::isNotNull),
```
mbrandonw: looks like we could drop this `filter` if yer doing the other `filter` down below

app/src/main/java/com/kickstarter/viewmodels/MessageHolderViewModel.java
luoser: oof good catch, i forgot that we were working with a VH here and that each view gets recycled. i was just doing what we usually do in a (activity) view model to fetch the freshest user in case the `environment.currentUser()` is stale ([i.e. ProfileViewModel](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/ProfileViewModel.java#L95-L98))

seems like we don't need this logic in this view holder however since we only need the current user's `id` value to see if they are the sender

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -0,0 +1,155 @@
...
+        .share();
+
+      final Observable<MessageThreadEnvelope> messageThreadEnvelope = envelopeNotification.compose(values());
+
+      final Observable<Notification<Backing>> backingNotification = Observable.combineLatest(
+        messageThread,
+        this.currentUser.observable(),
+        Pair::create
+      )
+        .filter(mu -> mu.first.backing() == null)
```
luoser: yep

PR #77, opened 2017-04-06
Not all signup errors are created equal.

mbrandonw:
Some signup errors aren't really errors (tho their http status code is 4xx), such as when we ask the user to verify their email. This will come up more often with guest checkout stuff.

So, let's just not show the "Sign up error" title in the dialog, and just show the message that the server provided us.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,12 +1,10 @@
 buildscript {
-    ext.kotlin_version = '1.0.7'
     repositories {
         jcenter()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.2.3'
         classpath 'com.google.gms:google-services:3.0.0'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -1,10 +1,12 @@
 buildscript {
+    ext.kotlin_version = '1.0.7'
```
mbrandonw: android studio told me to do this. do i do something bad?

luoser: 🤔 shouldn't have to do this in the `android/build.gradle`...maybe clean and rebuild?
also we are behind a version of android studio right now which funks up syncing sometimes if you want to upgrade feel free to!

mbrandonw: ill revert this for now so that i can merge

PR #79, opened 2017-04-11
Bump gradle version

coryroy:
## what
Originally the plan was simply to bump version of gradle to 2.3.1 and gradle wrapper to 3.3

This had some fall on effects as we were relying on:
`fragmentManager.getFragments()`

but this with the new gradle version this not shows correctly as hidden so we had to come up with another solution.

## how
Keeping a static HashMap of the fragments by adapter position:
```java
private static Map<Integer, DiscoveryFragment> fragmentMap;
```

seems to work very well.  WeakReferences were tried but it seemed to cause many problems with fragments disappearing on config changes.

Some important things with the implementation:

```java
/**
  * Don't pull directly from the map as it may be null
  */
  private @NonNull DiscoveryFragment safeGetFragment(final int position) {
    DiscoveryFragment fragment = fragmentMap.get(position);
    if (fragment == null) {
      fragment = getItem(position);
    }

    return fragment;
  }
```

and also in the constructor for the adapter:

```java
if (fragmentMap == null) {
  fragmentMap = new HashMap<>();
}
```

If you don't do the null check the map will get wiped out on rotate etc which you don't want as the fragments themselves are still around.


mbrandonw:
7974012-116640902_17-s4-v1

luoser:
oooh interesting, some lint failures with these gradle changes. we're gonna look into these tomorrow a.m. before merging

coryroy:
I pushed a new commit to the branch.  Now inheriting from AppCompatTextView in accordance with lint.

coryroy:
second lint issue is going to be a larger problem:
https://code.google.com/p/android/issues/detail?id=234673

Looks like this method is marked as hidden in the support api!

app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
```diff
@@ -40,8 +41,12 @@ public void setPrimaryItem(final @NonNull ViewGroup container, final int positio
   }
 
   @Override
-  public @NonNull Fragment getItem(final int position) {
-    return DiscoveryFragment.newInstance(position);
+  public
+  @NonNull
+  Fragment getItem(final int position) {
```
mbrandonw: we keep all of the modifiers in the same line of the function.

did something in our check style say to do this?

app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
```diff
@@ -58,38 +63,20 @@ public CharSequence getPageTitle(final int position) {
...
    */
   public void takeCategoriesForPosition(final @NonNull List<Category> categories, final int position) {
-    Observable.from(fragmentManager.getFragments())
-      .ofType(DiscoveryFragment.class)
-      .filter(frag -> {
-        final int fragmentPosition = frag.getArguments().getInt(ArgumentsKey.DISCOVERY_SORT_POSITION);
-        return fragmentPosition == position;
-      })
-      .subscribe(frag -> frag.takeCategories(categories));
+    fragmentMap.get(position).takeCategories(categories);
```
mbrandonw: sooooooooo much better!

app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
```diff
@@ -40,8 +41,12 @@ public void setPrimaryItem(final @NonNull ViewGroup container, final int positio
   }
 
   @Override
-  public @NonNull Fragment getItem(final int position) {
-    return DiscoveryFragment.newInstance(position);
+  public
+  @NonNull
+  Fragment getItem(final int position) {
+    DiscoveryFragment fragment = DiscoveryFragment.newInstance(position);
```
mbrandonw: looks like check style wants this to be `final`. we like to `final` all the things

app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
luoser: looks like something android studio does automatically when auto-completing override methods, cory and i are gonna pair this morning and i'll explain our code style!

coryroy:
This build is still failing during :app:lintExternalPre21Release, but the lint file isn't even created. I think it's crashing during lint!

bitmoji

mbrandonw:
> This build is still failing during :app:lintExternalPre21Release, but the lint file isn't even created. I think it's crashing during lint!

@coryroy according to the [lint output](https://285-76278501-gh.circle-artifacts.com/0/home/ubuntu/android-oss/app/build/reports/lint-results-externalPre21Release.html) it's pointing at this line:

```java
pages.stream().forEach(page -> fragmentMap.get(page).clearPage());
```

I didn't catch this initially, but this isn't valid java 6 code because it's using java 8 features. we need to change this to the old school way.

app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
```diff
@@ -1,25 +1,24 @@
...
 import java.util.List;
-
-import rx.Observable;
+import java.util.Map;
 
 public final class DiscoveryPagerAdapter extends FragmentPagerAdapter {
   private final Delegate delegate;
   private final FragmentManager fragmentManager;
   private List<String> pageTitles;
+  private Map<Integer, DiscoveryFragment> fragmentMap;
```
christopherwright: Main question I have here is whether there are any issues with holding onto a reference to these fragments – can it potentially cause any issues with GC? I couldn't see a clear explanation on how this works from a scan of the docs, so would just hope that we have confidence that this would not cause issues with destroying fragments.

coryroy: @christopherwright Good point.  Do you think putting weak references in the map would be a better idea.

.gitignore
```diff
@@ -27,8 +27,14 @@ proguard/
 # Log Files
 *.log
 
-# OSX files
+# OS generated files #
+######################
```
luoser: think we should cut out this comment line, to match the rest of this file's style

.gitignore
```diff
@@ -27,8 +27,14 @@ proguard/
...
 .DS_Store
+.DS_Store?
```
luoser: is there a reason why there's a `?` appended?

app/src/main/java/com/kickstarter/ui/adapters/DiscoveryPagerAdapter.java
```diff
@@ -40,8 +38,11 @@ public void setPrimaryItem(final @NonNull ViewGroup container, final int positio
   }
 
   @Override
-  public @NonNull Fragment getItem(final int position) {
-    return DiscoveryFragment.newInstance(position);
+  @NonNull
```
luoser: in general, we put our nullability annotations right in front of the type which it is annotating, i.e. 

```java
public @NonNull DiscoveryFragment
``` 

i'll update our styleguide to include this, since it's something we discussed a while ago as a team but neglected to include in the styleguide!

app/src/main/java/com/kickstarter/ui/fragments/DiscoveryFragment.java
```diff
@@ -44,7 +43,8 @@
 
   public DiscoveryFragment() {}
 
-  public static @NonNull Fragment newInstance(final int position) {
+  @NonNull
```
luoser: same here with `@NonNull`

coryroy: I think the auto-formatter messed this up as I didn't specifically change this. Is it possible to change the settings file so this doesn't happen on auto-format?

luoser:
alrighty, we're ready to rock n roll. for anyone on an older version of Android Studio, be sure to upgrade to 2.3.1!

PR #84, opened 2017-04-25
Remove getFragments() usage

luoser:
## what
It all began in #79 when @coryroy discovered that the upgraded linter no longer allowed us to use `getFragments()` from `FragmentManager`. This fix removes our usage of `getFragments()` from our `DiscoveryPagerAdapter`, a method we relied on to get references to our fragments safely for our sort pages.

The plan is to merge this in first, then we can go ahead with merging in #79 to get us onto the latest greatest Android Studio and gradle. 

## why
Turns out that `FragmentManager` always had an annotation to restrict its `getFragments()` usage, but since the method was still public our linter previously allowed it:
```java
@RestrictTo(LIBRARY_GROUP)
public abstract List<Fragment> getFragments();
```

## how
We know the `FragmentPagerAdapter` is finicky with keeping its references to instantiated fragments on rotation, so the biggest challenge was, is, and will always be to keep reliable references to non-null fragment views. Let's use our existing infrastructure to help us:

1. Create the `DiscoveryFragment`s in the Activity and pass them to the adapter. This way we know for sure that the lifecycle of the fragments is tied with the lifecycle of our parent activity.
2. Use the pager adapter's [`instantiateItem`](https://developer.android.com/reference/android/support/v4/app/FragmentPagerAdapter.html#instantiateItem) method to update our fragments when they get re-instantiated.
3. Add an `isInstantiated` helper in the `DiscoveryFragment` to filter for fragments that are actually ready to take data.

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/DiscoveryActivity.java
+++ b/app/src/main/java/com/kickstarter/ui/activities/DiscoveryActivity.java
@@ -30,6 +30,7 @@ import com.kickstarter.ui.toolbars.DiscoveryToolbar;
 import com.kickstarter.ui.views.SortTabLayout;
 import com.kickstarter.viewmodels.DiscoveryViewModel;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


christopherwright:
Looks good – have we tried stopping the process (like in https://github.com/kickstarter/android-oss/pull/48) just to test some of the gnarlier edges here?

luoser:
@christopherwright why yes i just did and it survived! 👍 😌

PR #89, opened 2017-04-26
Add feature flags to config, implement for messages

luoser:
## what
1. Added `features` to our config model so we can start tapping into our feature flags to use them! Config is already a part of our `environment` so this was a breeze.

2. Implemented flag for messages in our `ProfileViewModel`.

3. Modernized `ProfileViewModelTest` and added additional tests.

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/CurrentUser.java
+++ b/app/src/main/java/com/kickstarter/libs/CurrentUser.java
@@ -73,7 +73,7 @@ public class CurrentUser extends CurrentUserType {
   }
 
   @Override
-  public Observable<User> observable() {
+  public @NonNull Observable<User> observable() {
     return user;
   }
 }
--- a/app/src/main/java/com/kickstarter/ui/activities/ProfileActivity.java
+++ b/app/src/main/java/com/kickstarter/ui/activities/ProfileActivity.java
@@ -23,6 +23,7 @@ import com.kickstarter.libs.utils.ViewUtils;
 import com.kickstarter.models.Project;
 import com.kickstarter.ui.IntentKey;
 import com.kickstarter.ui.adapters.ProfileAdapter;
+import com.kickstarter.ui.views.IconButton;
 import com.kickstarter.viewmodels.ProfileViewModel;
 import com.squareup.picasso.Picasso;
 
@@ -45,6 +46,7 @@ public final class ProfileActivity extends BaseActivity<ProfileViewModel.ViewMod
   protected @Bind(R.id.created_text_view) TextView createdTextView;
   protected @Bind(R.id.backed_text_view) TextView backedTextView;
   protected @Bind(R.id.divider_view) View dividerView;
+  protected @Bind(R.id.messages_button) IconButton messagesButton;
   protected @Bind(R.id.recycler_view) RecyclerView recyclerView;
 
   @Override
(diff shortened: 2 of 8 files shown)


app/src/main/java/com/kickstarter/ui/activities/ProfileActivity.java
```diff
@@ -33,19 +34,22 @@
...
   private ProfileAdapter adapter;
   private RecyclerViewPaginator paginator;
 
   protected @Bind(R.id.avatar_image_view) ImageView avatarImageView;
-  protected @Bind(R.id.user_name_text_view) TextView userNameTextView;
-  protected @Bind(R.id.created_count_text_view) TextView createdCountTextView;
   protected @Bind(R.id.backed_count_text_view) TextView backedCountTextView;
-  protected @Bind(R.id.created_text_view) TextView createdTextView;
   protected @Bind(R.id.backed_text_view) TextView backedTextView;
+  protected @Bind(R.id.created_count_text_view) TextView createdCountTextView;
```
luoser: lotta noise here from alphabetizing

app/src/main/java/com/kickstarter/libs/CurrentUser.java
```diff
@@ -73,7 +73,7 @@ public void refresh(final @NonNull User freshUser) {
   }
 
   @Override
-  public Observable<User> observable() {
+  public @NonNull Observable<User> observable() {
```
luoser: our [interface's annotation of this](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/libs/CurrentUserType.java#L38) was also `@NonNull` so we should be consistent

app/src/main/java/com/kickstarter/viewmodels/ProfileViewModel.java
```diff
@@ -141,6 +148,9 @@ public ViewModel(final @NonNull Environment environment) {
       )
         .map(p -> p.first || p.second);
 
+      this.messagesButtonHidden = this.currentConfig.observable()
+        .map(config -> !config.features().getOrDefault("android_messages", false));
```
mbrandonw: so I think we want the default to be `true`, i.e. if we deleted the feature flag from the server we would interpret that as the feature being _on_ for everyone, not _off_.

does that make sense?

app/src/test/java/com/kickstarter/viewmodels/ProfileViewModelTest.java
```diff
@@ -210,38 +178,49 @@ public void testProfileViewModel_EmitsUserNameAndAvatar() {
...
+
+    this.avatarImageViewUrl.assertValues(user.avatar().medium());
+    this.userNameTextViewText.assertValues(user.name());
+  }
+
+  @Test
+  public void testProfileViewModel_MessagesButtonHidden() {
+    final Config config = ConfigFactory.config()
+      .toBuilder()
+      .features(Collections.singletonMap("android_messages", false))
```
mbrandonw: what about the case when `"android_messages": true` _and_ `"android_messages"` is absent? 🤔

app/src/main/java/com/kickstarter/viewmodels/ProfileViewModel.java
luoser: bitmoji

app/src/test/java/com/kickstarter/viewmodels/ProfileViewModelTest.java
luoser: 🤔 

i'll add more tests

PR #90, opened 2017-04-27
Enable commenting for creators

luoser:
## what
Creators can now comment on their own projects 😬. I refactored and cleaned up some VM and test stuff, didn't modernize either yet figured that can come at a later pass so we can focus on logic changes here.


app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java
```diff
@@ -47,6 +49,13 @@ public CommentsViewModel(final @NonNull Environment environment) {
     this.client = environment.apiClient();
     this.currentUser = environment.currentUser();
 
+    final Observable<User> currentUser = Observable.merge(
```
luoser: ehhh not super happy with this current user, think i may take a pass again tmrw a.m. to walk away from using `currentUser.observable()` like this

app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java
```diff
@@ -109,21 +118,41 @@ public CommentsViewModel(final @NonNull Environment environment) {
...
 
+    final Observable<Boolean> currentUserIsCreator = Observable.combineLatest(
+      currentUser,
+      project.map(Project::creator),
+      Pair::create
+    )
+      .map(userAndCreator -> userAndCreator.first != null && userAndCreator.first.id() == userAndCreator.second.id());
+
+    final Observable<Boolean> userCanComment = Observable.combineLatest(
+      currentUserIsCreator,
```
mbrandonw: since `currentUserIsCreator` is only used in this observable could we reduce the number of `combineLatest`s used and do all of that work in this one observable?

I think we are safe cause you gate this observables by a `takeWhen` later, but just good practice to not overuse `combineLatest`.

app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java
```diff
@@ -47,6 +49,13 @@ public CommentsViewModel(final @NonNull Environment environment) {
     this.client = environment.apiClient();
     this.currentUser = environment.currentUser();
 
+    final Observable<User> currentUser = Observable.merge(
+      this.currentUser.observable(),
+      this.currentUser.observable()
+        .compose(takeWhen(this.loginSuccess))
+        .flatMap(__ -> this.client.fetchCurrentUser().compose(neverError())).share()
```
luoser: ah right right, since fetching the user is only contingent on the `loginSuccess`, not `currentUser.observable()` !

app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java
luoser: yep can do, i think i had that moment of "hm thats a lot of `combineLatest`" but the code for some outputs read nicely

but they'll still read nicely

PR #92, opened 2017-05-01
Improve build time by caching some build config info

coryroy:
## what
Speed up the build

## how
Implement build config caching via Paperwork. It stores git hash and commit time in a json file rather than regenerating each build.

Notes: my build time went from 45 seconds to 2 seconds!


luoser:
cool that you saw such a big change! let's chat about this later today, in general we try to limit our dependencies and avoid passing `context` around our codebase, so let's explore what our other options may be

coryroy:
I also tried just enabling the build cache by adding `android.enableBuildCache=true` to the `gradle.properties` file. That brings subsequent builds to 11 secs down from 47, but not nearly as impressive as the Paperwork solution.

coryroy:
Paperwork removed? Want to have a look?

.gitignore
```diff
@@ -57,6 +57,8 @@ app/kickstarter.keystore
 app/signing.gradle
 # Google API key for GCM, overwritten during builds depending on the variant
 app/google-services.json
+# Paperwork helps with caching build info
+paperwork.json
```
luoser: is this still needed now that paperwork is removed?

coryroy: No. It isn't. I will remove it.

coryroy: image
Nice catch.

app/build.gradle
```diff
@@ -187,6 +184,35 @@ dependencies {
...
+    return value
+}
+
+def commitTime() {
+    def file = new File(project.buildDir, "commit-timestamp.txt")
+    if (file.exists()) {
+        return file.text
+    }
+
+    // TODO: upgrade gradle and use `getHeadGitCommit.date` instead
```
luoser: what version of gradle do we need to be at to use this?

gradle.properties
```diff
@@ -15,4 +15,6 @@
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
\ No newline at end of file
+# tl;dr: Projects are coupled if they change each others config and ours are not
+org.gradle.parallel=true
```
luoser: do we have more context/research as to why we should be doing this?

coryroy: Just to speed up the build. As the link in the gradle.properties says, it is perfectly safe as long as your subprojects configuration don't rely on any other subprojects configuration.  If we notice any weirdness, it is simple to turn this off.

app/build.gradle
coryroy: My mistake. Was a reference to a private method in source solution, removed.

gradle.properties
coryroy: Source: https://zeroturnaround.com/rebellabs/making-gradle-builds-faster/

> Besides the obvious speedup you’ll get from executing the build over using multiple >threads at the same time, you get the following additional benefits:

> 
> Configuration of projects in parallel.
> Re-use of configuration for unchanged projects.
> Project-level up-to-date checks.
> Using pre-built artifacts in the place of building dependent projects.
> The last two points are particularly important. You rarely work all over the place, most probably your changes to the code are nicely contained in a few projects at a time. That means that Gradle will be able to figure out and avoid building unnecessary projects over and over again. And the work that’s not done is the fastest work ever.
> 

Just waiting on your go ahead...

luoser: reading through the docs linked in the comments of the `gradle.properties`:

> It may not be immediately apparent, but using key Gradle features like the allprojects and subprojects keywords automatically cause your projects to be coupled. This is because these keywords are used in a build.gradle file, which defines a project. Often this is a “root project” that does nothing more than define common configuration, but as far as Gradle is concerned this root project is still a fully-fledged project, and by using allprojects that project is effectively coupled to all other projects. Coupling of the root project to subprojects does not impact 'configuration on demand', but using the allprojects and subprojects in any subproject's build.gradle file will have an impact.

since we do use `allprojects` and `subprojects` in our android's `build.gradle` we cannot enable this flag

https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:decoupled_projects

app/build.gradle
```diff
@@ -187,6 +184,34 @@ dependencies {
...
 }
 
+// SHA and timestamp caching courtesy of https://github.com/gdg-x/frisbee/blob/develop/app/build.gradle#L193-L218
+def commitSha() {
+    def file = new File(project.buildDir, "commit-sha.txt")
+    if (file.exists()) {
+        return file.text
+    }
+
+    def value = 'git rev-parse --short HEAD'.execute([], project.rootDir).text.trim()
```
luoser: seems like we can use `getHeadGitCommit().abbreviatedId` rather than this manual parsing now right?

app/build.gradle
```diff
@@ -187,6 +184,34 @@ dependencies {
...
+    return value
+}
+
+def commitTime() {
+    def file = new File(project.buildDir, "commit-timestamp.txt")
+    if (file.exists()) {
+        return file.text
+    }
+
+    def date = new Date()
```
luoser: seems like we can use `getHeadGitCommit().date` here right?

gradle.properties
luoser: we can comment this line back out now with that knowledge!

coryroy: Ok. That's good to know! I'll just remove it from the gradle.properties file all together so nobody is tempted to turn it on. :)

luoser: i think we should just revert it to being commented out, as it is by default, 1. to reduce PR change noise and 2. in case ever in the future we do need to revisit this, we'll know what was here

we'll always have this PR and our discussions here as reference

app/build.gradle
```diff
@@ -1,20 +1,17 @@
...
 
-def gitSha = 'git rev-parse --short HEAD'.execute([], project.rootDir).text.trim()
-def date = new Date();
-def buildDate = date.format("yyyy-MM-dd'T'HH:mm:ss'Z'", TimeZone.getTimeZone("UTC"))
 def externalPrivateVersion = new File(project.rootDir.path + "/app/external_version_code.txt").text.trim().toInteger()
 def externalPublicVersion = new File(project.rootDir.path + "/app/external_version_name.txt").text.trim()
 def internalPrivateVersion = new File(project.rootDir.path + "/app/internal_version_code.txt").text.trim().toInteger()
 def internalPublicVersion = new File(project.rootDir.path + "/app/internal_version_name.txt").text.trim()
-def isCircle = "true".equals(System.getenv("CIRCLECI"))
+def isCircle = "true" == System.getenv("CIRCLECI")
```
tasomaniac: 👍 much simpler

app/build.gradle
tasomaniac: Executing git like this will most probably break in windows machines. Please check. 
That's why we used [grgit][https://github.com/ajoberstar/gradle-git/] in frisbee. I believe it is also more readable like this `getHeadGitCommit().abbreviatedId`

app/build.gradle
```diff
@@ -187,6 +184,34 @@ dependencies {
     androidTestCompile 'com.android.support:support-annotations:25.1.0'
 }
 
+// SHA and timestamp caching courtesy of https://github.com/gdg-x/frisbee/blob/develop/app/build.gradle#L193-L218
+def commitSha() {
```
tasomaniac: You may consider putting these method in another named gradle file and apply them just like `apply from: 'quality.gradle'` above. Just a suggestion.

gradle.properties
coryroy: Ok. I'll just keep it, but commented out. 👍

app/build.gradle
coryroy: @luoser `getHeadGitCommit().abbreviatedId` would require the [grgit](https://github.com/ajoberstar/gradle-git/) library that @tasomaniac has recommended. Is that something you want to try?

app/build.gradle
coryroy: @tasomaniac @luoser I like this idea. It does look a bit messy with having utility functions inside the main build file.

app/build.gradle
luoser: ah right, i misread a comment you had in a previous iteration ([here](https://github.com/kickstarter/android-oss/pull/92/commits/2f5281137a42a62c9914f2f9ab2cb7eddcb4cf6b)) and thought it was because of a gradle versioning thing that we didn't have access to the method

i think it's fine to keep this, as well as line 207, as is for now. if you have a windows machine to test and improve this on, that'd be great, but if not sadly we'll have to punt the support for windows for right now to move on with at least this build improvement

coryroy:
@luoser Is this good to go? I think I've addressed everything. If not please let me know. 
bitmoji

gradle.properties
```diff
@@ -15,4 +15,5 @@
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
\ No newline at end of file
+# org.gradle.parallel=true
+#android.enableBuildCache=true
```
luoser: we can remove line 19 here since it's something you added that we didn't end up using

app/build.gradle
luoser: let's leave these here for now -- it's nice to just be able to look into one build.gradle file to see everything that's going on

PR #94, opened 2017-05-03
Add survey api

coryroy:
## What
In order to add the ability to take surveys in the android client we must first build out the api layer with the app

## How
Add api, model and factory to prepare for surveys feature


coryroy:
Ooops. Yes I will add that!

PR #95, opened 2017-05-03
Don't put the POTD at the top of search pages

coryroy:
## What
When performing searches it seemed weird that the top project had nothing to do with the search term

## How
Change the criteria for showing the POTD to not include search

### DIFF
--- a/app/src/main/java/com/kickstarter/services/DiscoveryParams.java
+++ b/app/src/main/java/com/kickstarter/services/DiscoveryParams.java
@@ -415,7 +415,11 @@ public abstract class DiscoveryParams implements Parcelable {
    * POTD comes back.
    */
   public boolean shouldIncludePotd() {
-    return isAllProjects() && page() != null && page() == 1 && (sort() == null && (term() == null || term().isEmpty()) || sort() == Sort.HOME);
+    return isAllProjects()
+      && page() != null
+      && page() == 1
+      && (sort() == null || sort() == Sort.HOME)
+      && (term() == null || term().isEmpty());
   }
 
   /**
--- a/app/src/main/java/com/kickstarter/services/DiscoveryParams.java
+++ b/app/src/main/java/com/kickstarter/services/DiscoveryParams.java
@@ -415,7 +415,7 @@ public abstract class DiscoveryParams implements Parcelable {
    * POTD comes back.
    */
   public boolean shouldIncludePotd() {
-    return isAllProjects() && page() != null && page() == 1 && (sort() == null || sort() == Sort.HOME);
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kickstarter/services/DiscoveryParams.java
```diff
@@ -415,7 +415,7 @@
    * POTD comes back.
    */
   public boolean shouldIncludePotd() {
-    return isAllProjects() && page() != null && page() == 1 && (sort() == null || sort() == Sort.HOME);
+    return isAllProjects() && page() != null && page() == 1 && (sort() == null && (term() == null || term().isEmpty()) || sort() == Sort.HOME);
```
mbrandonw: I think the parentheses got a little messed up here. if you expand it a bit you will see it's written as:

```swift
return isAllProjects() 
  && page() != null 
  && page() == 1 
  && (
    sort() == null 
    && (term() == null || term().isEmpty()) 
    || sort() == Sort.HOME
);
```

since `||` has a lower precedence than `&&` we will get `true` for this expression if `sort() == Sort.HOME` and will completely discard the `sort() == null` and `term()` portion of the condition.

what you really want is more like:

```java
return isAllProjects()
  && page() != null 
  && page() == 1 
  && (sort() == null || sort() == Sort.HOME)
  && (term() == null || term().isEmpty());
```

coryroy: ok. changed it.

app/src/test/java/com/kickstarter/libs/utils/DiscoveryParamsUtilsTest.java
```diff
@@ -5,49 +5,59 @@
...
+import org.junit.Assert;
+import org.junit.Test;
 
-public class DiscoveryParamsUtilsTest extends TestCase {
+public class DiscoveryParamsUtilsTest {
 
+  @Test
   public void testRefTag() {
-    assertEquals(
+    Assert.assertEquals(
```
mbrandonw: I'm down to keep the static imports `import static org.junit.Assert.*;` so that we can just do `assertEquals`. whatdya think?

coryroy: That's how I did it the first time, but milkrun wouldn't allow it, so I changed them

luoser: also agree that we should keep these static as it cuts down on the noise and it's the way we access the method throughout our codebase

PR #96, opened 2017-05-03
[Messages] Reply to message thread

luoser:
## what
Inching forward on messages. This is the ability to reply to message threads, design should be ignored for now to focus on functionality.

What's missing is replying to the remaining MessageSubjects, i.e. backings and projects, but that will come as we approach those screens.


app/src/main/java/com/kickstarter/services/ApiService.java
```diff
@@ -128,6 +130,17 @@
   @POST("/v1/users/reset")
   Observable<Response<User>> resetPassword(@Body ResetPasswordBody body);
 
+  @POST("/v1/message_threads/{message_thread_id}/messages")
+  Observable<Response<Message>> sendMessageToThread(@Path("message_thread_id") long messageThreadId, @Body MessageBody body);
+
+  @POST("/v1/projects/{project_id}/backers/{backer_id}/messages")
+  Observable<Response<Message>> sendMessageToBacker(
```
luoser: sneak preview of the other `sendMessage` endpoints we'll be using

app/src/main/java/com/kickstarter/services/ApiClient.java
```diff
@@ -335,6 +337,14 @@ public ApiClient(final @NonNull ApiService service, final @NonNull Gson gson) {
   }
 
   @Override
+  public @NonNull Observable<Message> sendMessageToThread(final @NonNull MessageThread messageThread, final @NonNull String body) {
+    return service
+      .sendMessageToThread(messageThread.id(), MessageBody.builder().body(body).build())
```
mbrandonw: body builder build body 💪

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -39,14 +46,20 @@
...
     Observable<List<Message>> messages();
 
+    /** Emits the participant name to be displayed. */
+    Observable<String> participantNameTextViewText();
+
     /** Emits the project name to be displayed. */
     Observable<String> projectNameTextViewText();
+
+    /** Emits an empty string when we want to clear the message edit text. */
+    Observable<String> setEmptyMessageEditText();
```
mbrandonw: seems we could just call this `setMessageEditText` and not even communicate that we are only emitting empty strings. the view doesn't need to know that, it just needs to follow orders!

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -39,14 +46,20 @@
...
+
+    /** Emits when we should show a message error toast. */
+    Observable<Boolean> showMessageErrorToast();
```
mbrandonw: what's the boolean for?

seems like instead this should be a `String` that emits the error message you plucked from the error envelope via `first(env.errorMessages())` and default to something if that doesn't exist.

app/src/main/java/com/kickstarter/ui/viewholders/MessageViewHolder.java
```diff
@@ -44,10 +46,10 @@ public MessageViewHolder(final @NonNull View view) {
       .compose(observeForUI())
       .subscribe(this::setParticipantAvatarImageView);
 
-    this.viewModel.outputs.messageBodyTextViewAlignParentEnd()
+    this.viewModel.outputs.messageBodyCardViewAlignParentEnd()
```
mbrandonw: 👍

app/src/main/java/com/kickstarter/services/ApiService.java
mbrandonw: 👀

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -64,7 +77,23 @@ public ViewModel(final @NonNull Environment environment) {
...
+        .share();
+
+      final Observable<Message> messageSent = messageNotification.compose(values());
+
+      this.setEmptyMessageEditText = messageSent.map(__ -> "");
+
+      final Observable<Notification<MessageThreadEnvelope>> envelopeNotification = Observable.merge(
+        messageThread,
+        messageThread.compose(takeWhen(messageSent))
+      )
```
mbrandonw: :mama-mia:

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
luoser: makes sense! this Boolean is doing nothing and it should be a string

this `Boolean` value comes from an initial iteration where I had hoped we could get around using a `PublishSubject` as an output for toasts (since we don't want these errors to show again and again on rotation) but doing so required more logic around the output vs. just using a publish subject for now to get that behavior for free ¯\_(ツ)_/¯

PR #85, opened 2017-04-25
Improve Search Result page

coryroy:
## What
- Add Baseball style project cards to search
- Add clearer reference tags
- Refactor into new style with ViewModels inside ViewHolders

PR #99, opened 2017-05-08
Search UI pass

luoser:
## what
Follow up to #85 containing UI detail and ViewModel cleanup for external release.

Left is Android before, right is current iOS:

 

After:

## todo
- [x] get @maggieKSR 's 👀 on this but in the meanwhile y'all can evaluate the VM logic!

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/ProjectSearchResultHolderViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/ProjectSearchResultHolderViewModel.java
@@ -30,6 +30,9 @@ public interface ProjectSearchResultHolderViewModel {
     /** Emits the formatted days to go text to be displayed. */
     Observable<String> deadlineCountdownValueTextViewText();
 
+    /** Emits the project clicked by the user. */
+    Observable<Project> notifyDelegateOfResultClick();
+
     /** Emits the percent funded text to be displayed. */
     Observable<String> percentFundedTextViewText();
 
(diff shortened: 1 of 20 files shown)


app/src/main/java/com/kickstarter/viewmodels/ProjectSearchResultHolderViewModel.java
```diff
@@ -18,113 +18,102 @@
 
 public interface ProjectSearchResultHolderViewModel {
 
-  final class Data {
```
luoser: using a `Pair` helped us clean up a lot of boilerplate code. in general, using a `Data` class for view models is nice, but typically for when there are 3+ pieces of data is when it really gives us a win

app/src/main/res/layout/featured_search_result_view.xml
```diff
@@ -15,87 +16,101 @@
...
+            tools:text="25"/>
+
+          <TextView
+            android:id="@+id/search_result_deadline_unit_text_view"
+            style="@style/BodySecondary"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            tools:text="days to go"/>
+
+        </LinearLayout>
```
coryroy: This should really be one text view with html.  Since you are rewriting the UI anyway, it makes sense to do it properly.  If someone has a small screen and a large font, you will get some very weird UI with this chained textview method.  Also it doesn't localize properly and word order is fixed.

app/src/main/java/com/kickstarter/viewmodels/ProjectSearchResultHolderViewModel.java
coryroy: This is fine. I still prefer data structures to pairs, but I love me less lines of code, too!

PR #100, opened 2017-05-08
Project card refactor

swoopej:
yossss first android pr.  going to add explanatory comments in the morning to explain all the new stuff.

This is just modernizing and refactoring the `ProjectCardViewHolder` and `ProjectCardHolderViewModel` and adding a bunch of tests!  No new behavior just trying to clean it up with @luoser.

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
@@ -34,9 +34,9 @@ public interface ProjectCardHolderViewModel {
   }
 
   interface Outputs {
-    Observable<String> backersCountText();
+    Observable<String> backersCountTextViewText();
     Observable<Boolean> backingViewGroupIsGone();
-    Observable<String> categoryNameText();
+    Observable<String> categoryNameTextViewText();
     Observable<String> deadlineCountdownText();
     Observable<Boolean> featuredViewGroupIsGone();
     Observable<List<User>> friendsForNamepile();
@@ -45,10 +45,10 @@ public interface ProjectCardHolderViewModel {
     Observable<Boolean> friendBackingViewIsHidden();
     Observable<Boolean> fundingUnsuccessfulTextViewIsGone();
     Observable<Boolean> metadataViewGroupIsGone();
-    Observable<Project> projectOutput();
+    Observable<Project> projectForDeadlineCountdownDetail();
     Observable<Integer> percentageFunded();
     Observable<Boolean> percentageFundedProgressBarIsGone();
-    Observable<String> percentageFundedText();
+    Observable<String> percentageFundedTextViewText();
     Observable<String> photoUrl();
     Observable<String> blurbText();
     Observable<String> nameText();
@@ -70,9 +70,9 @@ public interface ProjectCardHolderViewModel {
     public ViewModel(final @NonNull Environment environment) {
       super(environment);
 
-      this.projectOutput = this.project;
+      this.projectForDeadlineCountdownDetail = this.project;
 
-      this.backersCountText = this.project
+      this.backersCountTextViewText = this.project
         .map(Project::backersCount)
         .map(NumberUtils::format);
 
@@ -82,7 +82,7 @@ public interface ProjectCardHolderViewModel {
       this.blurbText = this.project
         .map(Project::blurb);
 
-      this.categoryNameText = this.project
+      this.categoryNameTextViewText = this.project
         .map(Project::category)
         .map(Category::name)
         .compose(coalesce(""));
@@ -109,9 +109,9 @@ public interface ProjectCardHolderViewModel {
 
       this.fundingUnsuccessfulTextViewIsGone = this.project
         .map(p ->
-          !p.state().equals(Project.STATE_CANCELED) &&
-            !p.state().equals(Project.STATE_FAILED) &&
-            !p.state().equals(Project.STATE_SUSPENDED));
+          !p.state().equals(Project.STATE_CANCELED)
+            && !p.state().equals(Project.STATE_FAILED)
+            && !p.state().equals(Project.STATE_SUSPENDED));
 
       this.imageIsInvisible = this.project
         .map(Project::photo)
@@ -133,7 +133,7 @@ public interface ProjectCardHolderViewModel {
       this.percentageFundedProgressBarIsGone = this.project
         .map(ProjectUtils::isCompleted);
 
-      this.percentageFundedText = this.project
+      this.percentageFundedTextViewText = this.project
         .map(Project::percentageFunded)
         .map(NumberUtils::flooredPercentage);
 
@@ -186,10 +186,10 @@ public interface ProjectCardHolderViewModel {
     private final PublishSubject<Project> project = PublishSubject.create();
     private final PublishSubject<Void> projectClicked = PublishSubject.create();
 
-    private final Observable<String> backersCountText;
+    private final Observable<String> backersCountTextViewText;
     private final Observable<Boolean> backingViewGroupIsGone;
     private final Observable<String> blurbText;
-    private final Observable<String> categoryNameText;
+    private final Observable<String> categoryNameTextViewText;
     private final Observable<String> deadlineCountdownText;
     private final Observable<Boolean> featuredViewGroupIsGone;
     private final Observable<String> friendAvatarUrl;
@@ -202,10 +202,10 @@ public interface ProjectCardHolderViewModel {
     private final Observable<Project> notifyDelegateOfProjectClick;
     private final Observable<Integer> percentageFunded;
     private final Observable<Boolean> percentageFundedProgressBarIsGone;
-    private final Observable<String> percentageFundedText;
+    private final Observable<String> percentageFundedTextViewText;
     private final Observable<String> photoUrl;
     private final Observable<Boolean> potdViewGroupIsGone;
-    private final Observable<Project> projectOutput;
+    private final Observable<Project> projectForDeadlineCountdownDetail;
     private final Observable<Boolean> projectStateViewGroupIsGone;
     private final Observable<DateTime> projectCanceledAt;
     private final Observable<DateTime> projectFailedAt;
@@ -229,8 +229,8 @@ public interface ProjectCardHolderViewModel {
       this.projectClicked.onNext(null);
     }
 
-    @Override public @NonNull Observable<String> backersCountText() {
-      return this.backersCountText;
+    @Override public @NonNull Observable<String> backersCountTextViewText() {
+      return this.backersCountTextViewText;
     }
     @Override public @NonNull Observable<Boolean> backingViewGroupIsGone() {
       return this.backingViewGroupIsGone;
@@ -238,8 +238,8 @@ public interface ProjectCardHolderViewModel {
     @Override public @NonNull Observable<String> blurbText() {
       return this.blurbText;
     }
-    @Override public @NonNull Observable<String> categoryNameText() {
-      return this.categoryNameText;
+    @Override public @NonNull Observable<String> categoryNameTextViewText() {
+      return this.categoryNameTextViewText;
     }
     @Override public @NonNull Observable<String> deadlineCountdownText() {
       return this.deadlineCountdownText;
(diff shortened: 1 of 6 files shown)


luoser:
fwiw i assume full responsibility for ci failures etc. was too excited to have this PR opened we will go through this and all our shiny new outputs tmrw am

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
```diff
@@ -0,0 +1,279 @@
...
+    Observable<String> categoryNameText();
+    Observable<String> deadlineCountdownText();
+    Observable<Boolean> featuredViewGroupIsGone();
+    Observable<List<User>> friendsForNamepile();
+    Observable<String> friendAvatarUrl();
+    Observable<Boolean> imageIsInvisible();
+    Observable<Boolean> friendBackingViewIsHidden();
+    Observable<Boolean> fundingUnsuccessfulTextViewIsGone();
+    Observable<Boolean> metadataViewGroupIsGone();
+    Observable<Project> projectOutput();
```
mbrandonw: this output has me a bit 🤔. tracing down how's it used I see it's due to the fact that our deadline helpers require a full project instead of a date https://github.com/kickstarter/android-oss/blob/888a3746835835016fc1a1bb32d8e2a90b8bffce/app/src/main/java/com/kickstarter/libs/utils/ProjectUtils.java#L126-L129. that's a bit 🤔 🤔.

if yer up for it, I say we disentangle that so that those helpers just take a date, and then this output can be more succinctly described as `projectDeadline` or something.

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
```diff
@@ -0,0 +1,279 @@
...
+
+    public ViewModel(final @NonNull Environment environment) {
+      super(environment);
+
+      final Observable<Boolean> projectInEndedStates = this.project
+        .map(
+          p -> p.state().equals(Project.STATE_SUCCESSFUL)
+            || p.state().equals(Project.STATE_CANCELED)
+            || p.state().equals(Project.STATE_FAILED)
+            || p.state().equals(Project.STATE_SUSPENDED)
```
mbrandonw: while looking through that deadline stuff I came across this `ProjectUtils::isCompleted` helper which does precisely this! https://github.com/kickstarter/android-oss/blob/888a3746835835016fc1a1bb32d8e2a90b8bffce/app/src/main/java/com/kickstarter/libs/utils/ProjectUtils.java#L83-L89

if we statically import we could even do `this.project.map(isCompleted)` :D

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
```diff
@@ -0,0 +1,279 @@
...
+        );
+
+      this.projectOutput = this.project;
+
+      this.backersCountText = this.project
+        .map(Project::backersCount)
+        .map(NumberUtils::format);
+
+      this.backingViewGroupIsGone = this.project
+        .map(p -> p.isBacking() || p.isStarred() || p.isPotdToday() || p.isFeaturedToday());
```
mbrandonw: how does `backingViewGroup` compare to `metadataViewGroup`? just tryna understand what is being shown/hidden...

_edit_: ohhhh I think I understand! this is the backing view group _inside_ the metadata group? perhaps we could rename some stuff to make that clearer?

with that clarity, I now wonder if this output is correct. should it not be gone just when not backing? I'll look at the tests to see how it all comes out in the end...

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
```diff
@@ -0,0 +1,279 @@
...
+          !p.state().equals(Project.STATE_CANCELED) &&
+            !p.state().equals(Project.STATE_FAILED) &&
+            !p.state().equals(Project.STATE_SUSPENDED));
+
+      this.imageIsInvisible = this.project
+        .map(Project::photo)
+        .map(ObjectUtils::isNull);
+
+      this.metadataViewGroupIsGone = this.project
+        .map(p -> !p.isBacking() && !p.isStarred() && !p.isPotdToday() && !p.isFeaturedToday());
```
mbrandonw: looks like this observable is precisely `this.backingViewGroupIsGone.map(isFalse)`. any benefits from thinking of it as such?

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
```diff
@@ -0,0 +1,279 @@
...
+        .map(Project::percentageFunded)
+        .map(NumberUtils::flooredPercentage);
+
+      this.photoUrl = this.project
+        .map(Project::photo)
+        .filter(ObjectUtils::isNotNull)
+        .map(Photo::med);
+
+      this.potdViewGroupIsGone = this.project
+        .map(p -> p.isBacking() || p.isStarred() || !p.isPotdToday() || p.isFeaturedToday());
```
mbrandonw: after having seen this `backing || starred || potd || featured` logic 4 times I'm starting to think there's a better way! what if we created a lil enum to express the four types of metadata there could be, and then we can localize the cascading logic into a single place, and derive the outputs from that?

hilariously, that is precisely how @luoser did it 9 months ago in swift! https://github.com/kickstarter/ios-oss/blob/master/Library/ViewModels/DiscoveryPostcardViewModel.swift

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
```diff
@@ -0,0 +1,279 @@
...
+
+      this.projectCanceledAt = this.project
+        .filter(p -> p.state().equals(Project.STATE_CANCELED))
+        .map(Project::stateChangedAt)
+        .compose(coalesce(new DateTime()));
```
mbrandonw: im a lil worried with these separate canceled/failed/successful/suspended/etc outputs. the main worry is that we are filtering on them, which means that as view holders are reused there is a chance that we don't reset their state cause the project was filtered out, and that could lead to a card rendering with two or more states :/

if you look at the screenshots for the 4 states we are concerned with:

[canceled](https://github.com/kickstarter/ios-oss/blob/master/Screenshots/_64/Kickstarter_Framework_iOSTests.DiscoveryPageViewControllerTests/testView_Card_Project_state_canceled_lang_en_device_phone4_7inch%402x.png) • [suspended](https://github.com/kickstarter/ios-oss/blob/master/Screenshots/_64/Kickstarter_Framework_iOSTests.DiscoveryPageViewControllerTests/testView_Card_Project_state_suspended_lang_en_device_phone4_7inch%402x.png) • [failed](https://github.com/kickstarter/ios-oss/blob/master/Screenshots/_64/Kickstarter_Framework_iOSTests.DiscoveryPageViewControllerTests/testView_Card_Project_state_failed_lang_en_device_phone4_7inch%402x.png) • [successful](https://github.com/kickstarter/ios-oss/blob/master/Screenshots/_64/Kickstarter_Framework_iOSTests.DiscoveryPageViewControllerTests/testView_Card_Project_state_successful_lang_en_device_phone4_7inch%402x.png)

you'll see that it mostly boils down to just a label for the state and a label for the date. perhaps we can merge all of that into just one observable for each label?

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
```diff
@@ -0,0 +1,279 @@
...
+        .filter(p -> p.state().equals(Project.STATE_SUCCESSFUL))
+        .map(Project::stateChangedAt)
+        .compose(coalesce(new DateTime()));
+
+      this.projectSuspendedAt = this.project
+        .filter(p -> p.state().equals(Project.STATE_SUSPENDED))
+        .map(Project::stateChangedAt)
+        .compose(coalesce(new DateTime()));
+
+      this.rootCategoryName = this.project
```
mbrandonw: I believe this output is only used for that lil featured tag so perhaps we should mentioned that in the name, e.g. `rootCategoryNameForFeatured`?

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
```diff
@@ -0,0 +1,279 @@
...
+        .map(Category::root)
+        .filter(ObjectUtils::isNotNull)
+        .map(Category::name);
+
+      this.setDefaultTopPadding = this.metadataViewGroupIsGone
+        .filter(b -> b)
+        .compose(Transformers.ignoreValues());
+
+      this.setMetadataTopPadding = this.metadataViewGroupIsGone
+        .filter(b -> !b)
```
mbrandonw: instead of filter should we `map` and then in the view use the boolean in the view to determine if we add the padding or remove the padding?

same for `setDefaultTopPadding`

app/src/test/java/com/kickstarter/viewmodels/ProjectCardholderViewModelTest.java
```diff
@@ -0,0 +1,491 @@
...
+      .build();
+    setUpEnvironment(environment());
+
+    this.vm.inputs.configureWith(project);
+    this.backingViewGroupIsGone.assertValues(false);
+  }
+
+  @Test
+  public void testEmitsBlurbText() {
+    final Project project = ProjectFactory.project().toBuilder().blurb("somebody once told me").build();
```
mbrandonw: LOOOL

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
swoopej: ooo siq

swoopej: ok addressed this one in https://github.com/kickstarter/android-oss/pull/100/commits/8d5f490c88e22170295f480bcced6490ed914497

but wanna note that I'm adding `Project.STATE_SUSPENDED` to the "completed" states of a project.  That helper is only used in [one spot](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/RewardViewModel.java#L120) in the `RewardViewModel`, and it seems like it's ok to add suspended to that behavior too?  paging lisa @luoser to confirm that though

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
swoopej: nope ty ty

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
mbrandonw: perf! even better!

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
swoopej: oh I meant yep ty ty

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
swoopej: 👍

app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java
```diff
@@ -0,0 +1,337 @@
...
+        .map(Project::isFriendBacking)
+        .map(BooleanUtils::negate);
+
+      this.friendsForNamepile = this.project
+        .filter(Project::isFriendBacking)
+        .map(Project::friends);
+
+      this.fundingUnsuccessfulTextViewIsGone = this.project
+        .map(p ->
+          !p.state().equals(Project.STATE_CANCELED) &&
```
luoser: i think for style we like to put the conditionals in the front of new lines i.e.

```java
!p.state().equals(Project.STATE_CANCELED)
&& !p.state().equals(Project.STATE_FAILED)
&& !p.state().equals(Project.STATE_SUSPENDED));
```

app/src/test/java/com/kickstarter/viewmodels/ProjectCardholderViewModelTest.java
```diff
@@ -0,0 +1,494 @@
...
+import org.junit.Test;
+
+import java.util.Collections;
+import java.util.List;
+
+import rx.observers.TestSubscriber;
+
+
+public class ProjectCardholderViewModelTest extends KSRobolectricTestCase {
+  ProjectCardHolderViewModel.ViewModel vm;
```
luoser: can make this `private`

PR #101, opened 2017-05-09
Search result view tweaking

luoser:
## what
Slipped my 👀 that we want our project names to not have bottom gravity, and that our image size should be 80x45dp as per Maggie's mocks

before & after on a 19 moto:
 

on a 22 nexus 6:

Takeaways? Make sure to take a few mins to do the visual QA on older devices!!!!! And ~ one day ~ may we will have screenshot tests to catch this for us.


luoser:
gonna "rush"-merge this since it's only XML tweaks that I have already run by @maggieKSR !

PR #102, opened 2017-05-10
Reward Surveys in Android

coryroy:
# What
Allow users to see unanswered surveys in the Android client
Put this feature behind a feature flag since it will be a WIP for a little while

# Why
Sometimes Creators need information from Backers in order to fulfill their rewards.  Android users currently have to use the web in order to see and take these surveys, which is not a good workflow for them.

# How
Will modify the ActivityFeedViewModel and add a couple new ViewHolders and associated sections in the Adapter

# See 👀
Will try to track down some designs to show you, for WIP screenshots, look down the PR.


app/src/main/java/com/kickstarter/ui/adapters/ActivityFeedAdapter.java
```diff
@@ -15,25 +16,31 @@
...
   public interface Delegate extends FriendBackingViewHolder.Delegate, ProjectStateChangedPositiveViewHolder.Delegate,
-    ProjectStateChangedViewHolder.Delegate, ProjectUpdateViewHolder.Delegate, EmptyActivityFeedViewHolder.Delegate {}
+    ProjectStateChangedViewHolder.Delegate, ProjectUpdateViewHolder.Delegate, EmptyActivityFeedViewHolder.Delegate, UnansweredSurveyViewHolder.Delegate {}
 
   public ActivityFeedAdapter(final @NonNull Delegate delegate) {
     this.delegate = delegate;
 
     insertSection(SECTION_LOGGED_IN_EMPTY_VIEW, Collections.emptyList());
     insertSection(SECTION_LOGGED_OUT_EMPTY_VIEW, Collections.emptyList());
+    insertSection(SECTION_SURVEYS_HEADER_VIEW, Collections.singletonList(null));
```
mbrandonw: I think this is the reason you are seeing extra padding at the top. you have inserted a blank row for the header view, and even when survey count is zero it will not get cleared because of the `if` gating.

there are a few ways you could fix it, but perhaps write some tests first for the expectations and then we can discuss refactorings to fix it.

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyHeaderViewHolder.java
```diff
@@ -0,0 +1,48 @@
...
+  public UnansweredSurveyHeaderViewHolder(final @NonNull View view) {
+    super(view);
+    ButterKnife.bind(this, view);
+
+    ksString = environment().ksString();
+  }
+  @Override
+  public void bindData(final @Nullable Object data) throws Exception {
+
+    // Null data means we should hide the view
```
mbrandonw: instead of `null` I think we should require that `data` be a positive integer. we would never configure this view holder with `0` or `null`, cause we would just omit the view in that case.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -1,38 +1,50 @@
...
+
 public interface ActivityFeedViewModel {
 
-  interface Inputs extends ActivityFeedAdapter.Delegate {
+  interface Inputs extends Delegate {
     /** Invoke when pagination should happen. */
     void nextPage();
 
+    /** Invoke when activity's onResume runs */
+    void resume();
```
mbrandonw: so I suggested we create this input and hit it from the activity so that we could better test lifecycle methods, much like we do in ios.

this works, but one simplification we could do is to leverage the fact that all of our VM's have a base class, and they all have the lifecycle methods that activities go through, and so we could just make these lifecycle observables available for free! wanna try creating a `onResume` observable (let's call it `onResume` cause it won't conflict with the method `onResume()`) in the base class that is automatically pinged on resume, and then leverage that here? should mean this input goes away and the activity code goes away, but we get the same effect.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -61,180 +64,240 @@
...
+        )
+        .compose(Transformers.combineLatestPair(surveyFeatureEnabled))
+        .switchMap(lifecycleAndLoggedInAndEnabled ->
+          lifecycleAndLoggedInAndEnabled.second
+            ? this.client.fetchUnansweredSurveys()
+            : Observable.just(Collections.emptyList())
+        );
+
+      responses
+        .subscribe(this.surveys::onNext);
```
mbrandonw: let's just attach this to the observable above and get rid of the local. only reason to break into a local is when you need to user it elsewhere.

coryroy: Ok. I'll just add it to the chain! 👍

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
coryroy: Awesome! I will refactor it to the base.

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyHeaderViewHolder.java
coryroy: Sure. I'll make that change. 👍

app/src/main/java/com/kickstarter/ui/adapters/ActivityFeedAdapter.java
coryroy: Cool! Thanks for the tips. I will look into that and see how I can test the adapter.

coryroy: You were exactly right! I changed the line to be:
`insertSection(SECTION_SURVEYS_HEADER_VIEW, Collections.emptyList());`
and made the view holder require NonNull and it worked!

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyHeaderViewHolder.java
```diff
@@ -26,23 +26,19 @@ public UnansweredSurveyHeaderViewHolder(final @NonNull View view) {
...
+    if (unansweredSurveyCount > 0) {
+      headingTextView.setText(
+        ksString.format(
+          "Reward_Surveys",
+          unansweredSurveyCount,
+          "reward_survey_count",
+          String.valueOf(unansweredSurveyCount)));
+      headingTextView.setVisibility(View.VISIBLE);
+    } else {
+      headingTextView.setVisibility(View.GONE);
```
mbrandonw: I don't think we need to mess with the visibility of the text view. we will simply not render this view holder if the survey count is 0

app/src/main/java/com/kickstarter/libs/ActivityViewModel.java
```diff
@@ -23,8 +23,9 @@
   private final CompositeSubscription subscriptions = new CompositeSubscription();
 
   private final PublishSubject<ActivityResult> activityResult = PublishSubject.create();
-
   private final PublishSubject<Intent> intent = PublishSubject.create();
+  protected final PublishSubject<ActivityEvent> onResume = PublishSubject.create();
```
mbrandonw: sorry, this was my suggestion, but after seeing it IRL now I think we should revert and go back to the explicit `resume` input you had before. I really don't wanna push more functional into this VM base class. there's a lot of benefit to making our VM's very explicit in what they require, so it's nice to see this VM say it has functionality built on top of the activity resuming. 

do you mind reverting this commit? you can just do a `git revert` to do it cleanly. 

also you can make the resume observable just a simple `<Void>` observable since we no longer need the activity lifecycle event.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -107,22 +107,20 @@ public ViewModel(final @NonNull Environment environment) {
...
-        .switchMap(lifecycleAndLoggedInAndEnabled ->
-          lifecycleAndLoggedInAndEnabled.second
+        .switchMap(resumedAndLoggedInAndEnabled ->
+          resumedAndLoggedInAndEnabled.second
             ? this.client.fetchUnansweredSurveys()
-            : Observable.just(Collections.emptyList())
-        );
-
-      responses
+            : Observable.just(new ArrayList<SurveyResponse>())
```
mbrandonw: any reason not to use `Collections.emptyList()`?

app/src/main/java/com/kickstarter/ui/adapters/ActivityFeedAdapter.java
```diff
@@ -42,6 +49,14 @@ public void takeActivities(final @NonNull List<Activity> activities) {
     notifyDataSetChanged();
   }
 
+  public void takeSurveys(final @NonNull List<SurveyResponse> surveyResponses) {
+    if (surveyResponses.size() > 0) {
+      setSection(SECTION_SURVEYS_HEADER_VIEW, Collections.singletonList(surveyResponses.size()));
+      setSection(SECTION_SURVEYS_VIEW, surveyResponses);
+    }
```
mbrandonw: we need an `else` to this `if` so that if surveys later load 0 results we will clear the area.

a good test case to have too.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -1,35 +1,44 @@
...
 import rx.Observable;
 import rx.subjects.BehaviorSubject;
 import rx.subjects.PublishSubject;
 
+import static com.kickstarter.libs.utils.ObjectUtils.coalesce;
+
 public interface ActivityFeedViewModel {
 
-  interface Inputs extends ActivityFeedAdapter.Delegate {
+  interface Inputs extends Delegate {
```
mbrandonw: let's not import the delegate like this. from a code reviewer perspective it's nice to see where that delegate interface is coming from.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -61,86 +61,123 @@
...
+    Observable<Boolean> loggedInEmptyStateIsVisible();
+
+    /** Emits a list of unanswered surveys to be shown in the user's activity feed */
+    Observable<List<SurveyResponse>> surveys();
   }
 
-  final class ViewModel extends ActivityViewModel<ActivityFeedActivity> implements ActivityFeedAdapter.Delegate,
-    Inputs, Outputs {
+  final class ViewModel extends ActivityViewModel<ActivityFeedActivity> implements
+    ActivityFeedViewModel.Inputs, ActivityFeedViewModel.Outputs {
```
mbrandonw: on the other hand, these make sense to just be `Inputs` and `Outputs` because they are defined right above and it works the exact same way in every view model.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -152,89 +189,105 @@ public ViewModel(final @NonNull Environment environment) {
...
+    private final Observable<SurveyResponse> goToSurvey;
+    private final BehaviorSubject<Boolean> isFetchingActivities= BehaviorSubject.create();
+    private final BehaviorSubject<Boolean> loggedInEmptyStateIsVisible = BehaviorSubject.create();
+    private final BehaviorSubject<Boolean> loggedOutEmptyStateIsVisible = BehaviorSubject.create();
+    private final BehaviorSubject<List<SurveyResponse>> surveys = BehaviorSubject.create();
 
-    public final Inputs inputs = this;
-    public final Outputs outputs = this;
+    public final ActivityFeedViewModel.Inputs inputs = this;
+    public final ActivityFeedViewModel.Outputs outputs = this;
```
mbrandonw: same here...

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -61,86 +61,123 @@
...
+        .filter(resumedAndLoggedIn ->
+          resumedAndLoggedIn.first == ActivityEvent.RESUME && resumedAndLoggedIn.second
+        )
+        .compose(Transformers.combineLatestPair(surveyFeatureEnabled))
+        .switchMap(resumedAndLoggedInAndEnabled ->
+          resumedAndLoggedInAndEnabled.second
+            ? this.client.fetchUnansweredSurveys()
+            : Observable.just(new ArrayList<SurveyResponse>())
+        )
+        .subscribe(this.surveys::onNext);
```
mbrandonw: looks like we are missing a `.compose(this.bindToLifecycle())` on this.

app/src/main/java/com/kickstarter/viewmodels/UnansweredSurveyHolderViewModel.java
```diff
@@ -0,0 +1,98 @@
...
+    /** Emits the survey description */
+    Observable<List<String>> surveyDescription();
+
+    /** Emits the survey url */
+    Observable<SurveyResponse> loadSurvey();
+  }
+
+  final class ViewModel extends ActivityViewModel<UnansweredSurveyViewHolder> implements
+    UnansweredSurveyHolderViewModel.Inputs,
+    UnansweredSurveyHolderViewModel.Outputs {
```
mbrandonw: can drop these fully qualified names

app/src/main/java/com/kickstarter/viewmodels/UnansweredSurveyHolderViewModel.java
```diff
@@ -0,0 +1,98 @@
...
+    private final PublishSubject<SurveyResponse> configData = PublishSubject.create();
+    private final PublishSubject<Void> surveyClicked = PublishSubject.create();
+
+    private final Observable<String> creatorAvatarImage;
+    private final Observable<String> creatorName;
+    private final Observable<List<String>> surveyDescription;
+    private final Observable<SurveyResponse> goToSurvey;
+
+    public final UnansweredSurveyHolderViewModel.Inputs inputs = this;
+    public final UnansweredSurveyHolderViewModel.Outputs outputs = this;
```
mbrandonw: here too

app/src/main/java/com/kickstarter/viewmodels/UnansweredSurveyHolderViewModel.java
```diff
@@ -0,0 +1,98 @@
...
+      return this.creatorName;
+    }
+    @Override public Observable<List<String>> surveyDescription() {
+      return this.surveyDescription;
+    }
+    @Override public Observable<SurveyResponse> loadSurvey() {
+      return this.configData;
+    }
+
+    private List<String> getSurveyDescription(final @NonNull SurveyResponse surveyResponse) {
```
mbrandonw: a few things:

* all private helper functions should be `static` because they shouldn't rely on anything in the view model. they should just be lil pure functions that take all their input and spit out output.

* it is dangerous to use a list to represent two pieces of data that you later `get(0)` and `get(1)` on. we have no compiler time guarantee that that contract will continue to hold in the future. 

There are two options to make it safer. 

* We can make it a pair and name the variable explicitly as `creatorNameAndProjectName` to make it clear what `first` and `second` are.

* we can pass the `Project` back to the view and let the view just pluck out the data it needs to render it's string. then the output could be called something more like `projectForSurveyDescription`.

coryroy:
I started working on the tests last night and may need some help. I will
attach screenshots when I get into the office 🏢. 👍

coryroy:
device-2017-05-11-092519
With Surveys

device-2017-05-11-092345
Without surveys

app/src/main/java/com/kickstarter/libs/ActivityViewModel.java
coryroy: 👍 I have reverted the commit.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
coryroy: Right!

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
coryroy: Good point. I've made the change. 💯

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
coryroy: Ok. 👍

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
coryroy: 👍

app/src/main/java/com/kickstarter/viewmodels/UnansweredSurveyHolderViewModel.java
coryroy: 👍

app/src/main/java/com/kickstarter/services/ApiClient.java
```diff
@@ -233,6 +233,23 @@ public ApiClient(final @NonNull ApiService service, final @NonNull Gson gson) {
       .subscribeOn(Schedulers.io());
   }
 
+
+  @Override
+  public @NonNull Observable<SurveyResponse> fetchSurveyResponse(final @NonNull int surveyResponseId) {
```
luoser: don't need the `@NonNull` annotation for a primitive type. i see it's a vestige of an older commit but does that make sense?

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyHeaderViewHolder.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+  @Override
+  public void bindData(final @Nullable Object data) throws Exception {
+    final int unansweredSurveyCount = ObjectUtils.requireNonNull(
+      (int) data
+    );
+    if (unansweredSurveyCount > 0) {
+      headingTextView.setText(
+        ksString.format(
+          "Reward_Surveys",
```
luoser: can use `@BindString` to bind this up top

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyViewHolder.java
```diff
@@ -0,0 +1,92 @@
...
+  private final KSString ksString;
+  @Bind(R.id.survey_avatar_image) ImageView creatorAvatarImageView;
+  @Bind(R.id.survey_title) TextView surveyTitleTextView;
+  @Bind(R.id.survey_text) TextView surveyTextView;
+
+  @BindString(R.string.Creator_name_needs_some_information_to_deliver_your_reward_for_project_name) String surveyDescriptionString;
+
+  private final UnansweredSurveyHolderViewModel.ViewModel viewModel;
+
+  private final Delegate delegate;
```
luoser: can move all these `private final` variables to top of class for some organization!

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyViewHolder.java
```diff
@@ -0,0 +1,92 @@
...
+
+import java.util.List;
+
+import butterknife.Bind;
+import butterknife.BindString;
+import butterknife.ButterKnife;
+
+import static com.kickstarter.libs.rx.transformers.Transformers.observeForUI;
+
+public class UnansweredSurveyViewHolder extends KSViewHolder {
```
luoser: `final` class

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyViewHolder.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private final UnansweredSurveyHolderViewModel.ViewModel viewModel;
+
+  private final Delegate delegate;
+
+  public interface Delegate {
+    void surveyClicked(UnansweredSurveyViewHolder viewHolder, SurveyResponse surveyResponse);
+  }
+
+  public UnansweredSurveyViewHolder(final View view, final Delegate delegate) {
```
luoser: annotations

app/src/main/java/com/kickstarter/viewmodels/UnansweredSurveyHolderViewModel.java
coryroy: I agree! I'll go with option 2 for now and let's see how that looks.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -61,180 +64,238 @@
...
     @Override public void refresh() {
-      refresh.onNext(null);
+      this.refresh.onNext(null);
+    }
+
+    @Override public void resume() {
+      this.resume.onNext(ActivityEvent.RESUME);
+    }
+
+    @Override @NonNull public Observable<List<Activity>> activities() {
```
luoser: can you please audit these output functions for annotation order? as a reminder we like to do `public @NonNull` so that the annotation is right by the type it's annotating!

app/src/main/java/com/kickstarter/viewmodels/UnansweredSurveyHolderViewModel.java
```diff
@@ -0,0 +1,98 @@
...
+    public final UnansweredSurveyHolderViewModel.Outputs outputs = this;
+
+    @Override public void configureWith(final @NonNull SurveyResponse surveyResponse) {
+      this.configData.onNext(surveyResponse);
+    }
+    @Override public void surveyClicked() {
+      this.surveyClicked.onNext(null);
+    }
+
+    @Override public Observable<String> creatorAvatarImage() {
```
luoser: `@NonNull` for all these output funcs

app/src/main/res/layout/unanswered_survey_view.xml
```diff
@@ -0,0 +1,78 @@
...
+  xmlns:card_view="http://schemas.android.com/apk/res-auto"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="match_parent"
+  android:layout_height="wrap_content"
+  android:layout_marginBottom="@dimen/grid_new_2"
+  android:focusable="true"
+  card_view:cardCornerRadius="0dp"
+  card_view:cardUseCompatPadding="true">
+
+  <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
luoser: is this a duplicate `xmlns:android="http://schemas.android.com/apk/res/android"` ?

app/src/main/res/layout/unanswered_survey_view.xml
```diff
@@ -0,0 +1,78 @@
...
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <ImageView
+      android:id="@+id/survey_avatar_image"
+      android:layout_width="@dimen/discovery_activity_photo_width"
+      android:layout_height="@dimen/discovery_activity_photo_height"
+      android:layout_alignParentStart="true"
```
luoser: should have a matching `layout_alignParentLeft` for older devices

app/src/main/res/layout/unanswered_survey_view.xml
```diff
@@ -0,0 +1,78 @@
...
+      tools:ignore="ContentDescription" />
+
+    <TextView
+      android:id="@+id/survey_title"
+      style="@style/SubheadPrimary"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:layout_alignTop="@id/survey_avatar_image"
+      android:layout_marginBottom="@dimen/activity_padding_x"
+      android:layout_toEndOf="@id/survey_avatar_image"
```
luoser: should have matching `layout_toRightOf`

app/src/main/res/layout/unanswered_survey_view.xml
```diff
@@ -0,0 +1,78 @@
...
+    <TextView
+      android:id="@+id/survey_text"
+      style="@style/Headline"
+      android:layout_width="match_parent"
+      android:layout_height="wrap_content"
+      android:layout_below="@id/survey_avatar_image"
+      android:layout_margin="@dimen/grid_new_2" />
+
+    <View
+      android:id="@+id/separator"
```
luoser: we actually do have a `horizontal_line_1dp_view` you can reuse here with a `<layout/>` tag

app/src/main/res/layout/unanswered_survey_view.xml
```diff
@@ -0,0 +1,78 @@
...
+      android:layout_width="17dp"
+      android:layout_height="17dp"
+      android:layout_alignTop="@id/respond_button"
+      android:layout_toStartOf="@id/respond_button"
+      android:tint="@color/green"
+      android:src="@drawable/ic_survey_respond_icon"
+      tools:ignore="ContentDescription" />
+
+  </RelativeLayout>
+</android.support.v7.widget.CardView>
```
luoser: newline

app/src/main/res/layout/unanswered_surveys_header_view.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<TextView
+  android:id="@+id/heading"
+  style="@style/SubheadPrimary"
+  xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools"
```
luoser: again we do like to have our `xmlns:` tags up top in our layout files

app/src/main/java/com/kickstarter/ui/adapters/ActivityFeedAdapter.java
coryroy: Right!

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
```diff
@@ -1,31 +1,70 @@
...
+  final TestSubscriber<Void> goToDiscovery = new TestSubscriber<>();
+  final TestSubscriber<Void> goToLogin = new TestSubscriber<>();
+  final TestSubscriber<Project> goToProject = new TestSubscriber<>();
+  final TestSubscriber<Activity> goToProjectUpdate = new TestSubscriber<>();
+  final TestSubscriber<SurveyResponse> goToSurvey = new TestSubscriber<>();
+  final TestSubscriber<Boolean> loggedOutEmptyStateIsVisible = new TestSubscriber<>();
+  final TestSubscriber<Boolean> loggedInEmptyStateIsVisible = new TestSubscriber<>();
+  final TestSubscriber<List<SurveyResponse>> surveys = new TestSubscriber<>();
+
+  private void setUpEnvironment(@NonNull final Environment environment) {
```
luoser: `final @NonNull`

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyHeaderViewHolder.java
coryroy: Yes. That makes sense. 👍

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
```diff
@@ -1,31 +1,70 @@
...
+    vm = new ViewModel(environment);
+
+    vm.outputs.activities().subscribe(this.activities);
+    vm.outputs.goToDiscovery().subscribe(this.goToDiscovery);
+    vm.outputs.goToLogin().subscribe(this.goToLogin);
+    vm.outputs.goToProject().subscribe(this.goToProject);
+    vm.outputs.goToProjectUpdate().subscribe(this.goToProjectUpdate);
```
luoser: let's be consistent with our use of `this.` here, and throughout this file!

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
```diff
@@ -126,4 +141,203 @@ public void testLoginFlow() {
...
+    final TestScheduler scheduler = new TestScheduler();
+
+    final List<SurveyResponse> surveyResponses = Arrays.asList(
+      SurveyResponseFactory.surveyResponse(),
+      SurveyResponseFactory.surveyResponse()
+    );
+
+    final MockApiClient apiClient = new MockApiClient() {
+      @Override public @NonNull
+      Observable<List<SurveyResponse>> fetchUnansweredSurveys() {
```
luoser: check your formatting here

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
```diff
@@ -126,4 +141,203 @@ public void testLoginFlow() {
...
```
luoser: same funky formatting here

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
```diff
@@ -126,4 +141,203 @@ public void testLoginFlow() {
...
```
luoser: and here, and one more down on line 312

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
```diff
@@ -126,4 +141,203 @@ public void testLoginFlow() {
...
+      .currentUser(currentUser)
+      .currentConfig(currentConfig)
+      .build();
+
+    setUpEnvironment(environment);
+
+    surveys.assertNoValues();
+  }
+
+
```
luoser: watch for extra newlines

app/src/main/java/com/kickstarter/services/ApiClient.java
coryroy: Yes! That makes perfect sense! An int can't be null! 👍

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyHeaderViewHolder.java
coryroy: Yes. I like those, too. Will change.

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyViewHolder.java
coryroy: Yes. Looks better! 👍

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyViewHolder.java
coryroy: I'll put in the annotations to keep out the riff-raff! 👍

app/src/main/java/com/kickstarter/viewmodels/UnansweredSurveyHolderViewModel.java
coryroy: Got it! 👍

app/src/main/res/layout/unanswered_survey_view.xml
coryroy: Yes. That's not needed. Remove. 💯

app/src/main/res/layout/unanswered_survey_view.xml
coryroy: API 17 may make these redundant. Let's look into it!

app/src/main/res/layout/unanswered_survey_view.xml
coryroy: Cool. That is much better.

app/src/main/res/layout/unanswered_surveys_header_view.xml
coryroy: I prefer that too. Changed!

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
coryroy: Ooops. Ok fixed.

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
coryroy: fixed `this` 👍

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
coryroy: Changed all of these!

app/src/test/java/com/kickstarter/viewmodels/ActivityFeedViewModelTest.java
coryroy: Got rid of the tailgater.

app/src/main/res/layout/unanswered_surveys_header_view.xml
luoser: just want to clarify too that it's style for consistency throughout our code base, not just a personal preference of mine :)

app/src/main/res/values/strings.xml
```diff
@@ -45,6 +45,8 @@
   <string name="Sha">SHA</string>
   <string name="Staging">Staging</string>
   <string name="Submit_feedback">Submit feedback</string>
+  <string name="token_reward_surveys">Reward_Surveys</string>
+  <string name="token_reward_survey_count">reward_survey_count</string>
```
luoser: strings in this file shouldn't be hard coded--let's chat about this, think we should to review strings

coryroy: Ok. I'll slack you.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -115,15 +114,15 @@ public ViewModel(final @NonNull Environment environment) {
           this.currentUser.isLoggedIn(),
           Pair::create
         )
-        .filter(resumedAndLoggedIn ->
-          resumedAndLoggedIn.first == ActivityEvent.RESUME && resumedAndLoggedIn.second
-        )
+        .map(resumedAndLoggedIn -> resumedAndLoggedIn.second)
+        .filter(loggedIn -> loggedIn)
```
mbrandonw: we've got a [`isTrue`](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/libs/utils/BooleanUtils.java#L18) utility for this kind of thing. may seem kind of silly to have, but it allows you to write this very declaratively as `.filter(isTrue)`

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -115,15 +114,15 @@ public ViewModel(final @NonNull Environment environment) {
           this.currentUser.isLoggedIn(),
           Pair::create
         )
-        .filter(resumedAndLoggedIn ->
-          resumedAndLoggedIn.first == ActivityEvent.RESUME && resumedAndLoggedIn.second
-        )
+        .map(resumedAndLoggedIn -> resumedAndLoggedIn.second)
```
mbrandonw: if you wanted to try for a stretch goal, you could create a lil pair util function so that we can do `map(second)`. we do this kind of thing quite often in swift and it's very convenient.

this is also an example of [point-free programming](https://en.wikipedia.org/wiki/Tacit_programming), which we are big into here.

app/src/main/java/com/kickstarter/libs/ActivityViewModel.java
```diff
@@ -38,6 +38,11 @@ public void activityResult(final @NonNull ActivityResult activityResult) {
     this.activityResult.onNext(activityResult);
   }
 
+  /** Returns an observable of lifecycle events for this activity. */
+  public @NonNull Observable<ActivityEvent> lifecycle() {
```
mbrandonw: oh think we can drop this now

app/src/main/res/values/strings.xml
coryroy: I've removed those strings.

app/src/main/res/layout/unanswered_surveys_header_view.xml
coryroy: @luoser Thanks for the clarification. Consistency is a personal preference of mine. :) So I am happy to comply either way.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
coryroy: I was thinking that I didn't really like it the way it is. I will convert it to use the function, which as you say, is much more declarative and clear.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
coryroy: I will take a stab at this.

app/src/main/java/com/kickstarter/libs/rx/transformers/Transformers.java
```diff
@@ -184,4 +185,11 @@ private Transformers() {}
   public static @NonNull <T> ObserveForUITransformer<T> observeForUI() {
     return new ObserveForUITransformer<>();
   }
+
+  /**
+   * Emits the second of any input pair
+   */
+  public static <R> R second(final @NonNull Pair<?, R> anyPair) {
```
mbrandonw: let's move this to a `PairUtils` classs and create the corresponding `first` cause I'm sure we will find that useful soon enough!

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -114,8 +115,8 @@ public ViewModel(final @NonNull Environment environment) {
           this.currentUser.isLoggedIn(),
           Pair::create
         )
-        .map(resumedAndLoggedIn -> resumedAndLoggedIn.second)
-        .filter(loggedIn -> loggedIn)
+        .map(Transformers::second)
+        .filter(BooleanUtils::isTrue)
```
mbrandonw: let's statically import these so that we can do `map(second).filter(isTrue)`

app/src/main/java/com/kickstarter/libs/rx/transformers/Transformers.java
mbrandonw: just to clarify, we want this in pair utils cause it's a nice lil pure function that we can use in any context, not just with observables. the `Transformers.java` file is about creating new operators on observables.

coryroy: Thanks for clarifying. I was worried that it didn't fit exactly in the Transformers class so this sounds good. I will move it and create the corresponding `first` method as well.

coryroy: Also, I will quickly write some tests for these.

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
luoser: that's a Swift-ism! don't think it's possible in Java

mbrandonw: lol :foresaken-kid:

app/src/main/java/com/kickstarter/ui/adapters/ActivityFeedAdapter.java
```diff
@@ -29,12 +30,12 @@
...
   private static final int SECTION_ACTIVITIES_VIEW = 4;
 
-  private final Delegate delegate;
+  private final @Nullable Delegate delegate;
 
   public interface Delegate extends FriendBackingViewHolder.Delegate, ProjectStateChangedPositiveViewHolder.Delegate,
     ProjectStateChangedViewHolder.Delegate, ProjectUpdateViewHolder.Delegate, EmptyActivityFeedViewHolder.Delegate, UnansweredSurveyViewHolder.Delegate {}
 
-  public ActivityFeedAdapter(final @NonNull Delegate delegate) {
+  public ActivityFeedAdapter(final @Nullable Delegate delegate) {
```
mbrandonw: I just pushed this change to the branch.

basically, I think our delegates should always be optional (much like we do in iOS) and that way we don't have to use them in tests and put in an empty implementation.

coryroy: Ok. I will put and have a look and make changes, if necessary.

app/src/test/java/com/kickstarter/ui/adapters/ActivityFeedAdapterTest.java
```diff
@@ -19,9 +19,9 @@
 import java.util.Collections;
 import java.util.List;
 
-public class ActivityFeedAdapterTest extends KSRobolectricTestCase implements ActivityFeedAdapter.Delegate {
```
coryroy: So happy to see this requirement for a delegate in a test removed! Thank you!

app/src/main/java/com/kickstarter/ui/viewholders/UnansweredSurveyHeaderViewHolder.java
coryroy: After we worked through this, it appears, we should leave this as is.

app/src/main/java/com/kickstarter/ApplicationModule.java
```diff
@@ -200,7 +200,7 @@ String provideClientId(final @NonNull ApiEndpoint apiEndpoint) {
   @Singleton
   @NonNull HttpLoggingInterceptor provideHttpLoggingInterceptor() {
     final HttpLoggingInterceptor interceptor = new HttpLoggingInterceptor();
-    interceptor.setLevel(HttpLoggingInterceptor.Level.HEADERS);
+    interceptor.setLevel(HttpLoggingInterceptor.Level.BODY);
```
mbrandonw: what was this change? we shouldn't be making changes to this in this PR

coryroy: Oh. I needed this to debug the api. It is only used in debug, but I can revert that.

PR #109, opened 2017-06-08
fre$h new back button

luoser:
## what
Updating our back button style, everywhere, from Material Design™ to our new icon set.

Before (left) and Afer (right):

device-2017-06-08-105812 device-2017-06-08-105332

## how
Moved all repeated back button attributes into a new `style` called `ToolbarIconBackButton` so from now on, when making a new toolbar with a back button, simply use:

```xml
style="@style/ToolbarIconBackButton"
```

:neckbeard:


christopherwright:
If we start using more of the other icons from the new icon set, we could also bring in the full typeface and reference each icon using ligatures/codes, as we do with the material icons. The upside is we won't have to manually create pngs for each icon.

PR #110, opened 2017-06-08
Modernize ViewPledgeVM

luoser:
## what
Bringing our `ViewPledgeViewModel` up to speed, as well as its tests, and some activity and view cleanup in prep for messages, and to be nice to our future selves. 

Added a feature flagged `View messages` button (temp design).

device-2017-06-08-172818

## what's next
Adding logic to our `MessagesActivity` to accept a `Project` and a `Backing` will render these backing messages properly!


app/src/test/java/com/kickstarter/viewmodels/ViewPledgeViewModelTest.java
```diff
@@ -293,43 +273,73 @@ public void testShipping_withShippingLocation() {
...
+
+    this.viewMessagesButtonIsGone.assertValues(false);
   }
 
-  private @NonNull ViewPledgeViewModel vm(final @NonNull Backing backing) {
-    return new ViewPledgeViewModel(environment(backing));
+  /**
+   * Returns an environment with a backing and the backing's current user.
+   */
+  private @NonNull Environment envWithBacking(final @NonNull Backing backing) {
```
luoser: this is a unique test file in which we pull away a method to create an environment with an api client that has a `backing` and a CurrentUser which is the `backing.backer()` user...decided to condense and keep this logic at the bottom of the file

app/src/main/java/com/kickstarter/ui/activities/ViewPledgeActivity.java
```diff
@@ -68,124 +73,140 @@ public void onCreate(final @Nullable Bundle savedInstanceState) {
...
-      .observeOn(mainThread())
-      .subscribe(__ -> viewModel.inputs.projectClicked());
+    this.ksString = environment.ksString();
 
-    viewModel.outputs.backerNameTextViewText()
+    this.viewModel.outputs.backerNameTextViewText()
       .compose(bindToLifecycle())
-      .observeOn(mainThread())
-      .subscribe(backerNameTextView::setText);
+      .compose(observeForUI())
```
luoser: `Transformers.observeForUI` is an transformer we've slowly been moving towards for rendering our UI components more efficiently..see https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/libs/rx/transformers/Transformers.java#L184 for more documentation :neckbeard:

app/src/test/java/com/kickstarter/viewmodels/ViewPledgeViewModelTest.java
swoopej: noice makes sense

PR #111, opened 2017-06-14
Use gradle 3.5

onmyway133:
First of all, I really like your approach on ViewModel input and output, and I also love Rx. I used it on iOS too 🚀 
I don't know if this happens to other people, but I'm using Android Studio 3 Canary 3, and I get 

```sh
Gradle version 3.3 does not support forTask() method on BuildActionExecuter
```

I hope this will fix it. Thanks 😇


luoser:
heya thanks for this suggestion, but since you're using a version from the canary channel, a gradle 3.3+ is required. we're gonna stay with 2.3 in the stable channel for now so you'll have to keep those changes on your local env until we catchup!

PR #113, opened 2017-06-14
[Messages] Add backing & project logic to messages

luoser:
## what
Yet Another Messages PR! This time more infrastructure work to hook up messages for a Backing thread. Includes:

* refactored `MessagesVM` to be initialized with _either_ a `MessageThread` (tap on messages icon from the profile screen) or a `Pair<Backing, Project>` (tap on Message creator button ) via the `intent()`
*  If user **views message** from **view pledge**, then `view pledge` button in message is hidden, and `x` is a `back arrow`
* If user **views pledge** from **view message**, then `view message` button is hidden
* some new 🐨 events

## what's next
Cleaning up toolbar UX, adding timestamps, instrumentation and a11y sweeps


app/src/main/java/com/kickstarter/ui/data/MessagesData.kt
```diff
@@ -0,0 +1,14 @@
+package com.kickstarter.ui.data
+
+import com.kickstarter.libs.Either
+import com.kickstarter.models.Backing
+import com.kickstarter.models.MessageThread
+import com.kickstarter.models.Project
+import com.kickstarter.models.User
+
+data class MessagesData(
```
luoser: hello kotlin :shipit:

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -75,129 +103,211 @@ public ViewModel(final @NonNull Environment environment) {
...
+        .map(Either::right)
+        .filter(ObjectUtils::isNotNull)
+        .map(PairUtils::second);
+
+      final Observable<MessageThread> configThread = configData
+        .map(Either::left)
+        .filter(ObjectUtils::isNotNull);
+
+      final Observable<Either<Backing, MessageThread>> backingOrThread = Observable.merge(
+        configBacking.map(backing -> new Either.Left<>(backing)),
```
luoser: this can be replaced with a method reference, as can the `thread -> new Either.Right<>(thread)` below, but I find this syntax more understandable than

```java
(Func1<Backing, Either.Left<Backing, MessageThread>>) Either.Left::new
```
if that's okay

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -75,129 +103,211 @@ public ViewModel(final @NonNull Environment environment) {
...
 
-      final Observable<Notification<MessageThreadEnvelope>> envelopeNotification = Observable.merge(
-        messageThread,
-        messageThread.compose(takeWhen(messageSent))
+      final Observable<MessageThreadEnvelope> messageThreadEnvelope = Observable.merge(
+        backingOrThread,
+        backingOrThread.compose(takeWhen(messageSent))
       )
-        .switchMap(thread -> this.client.fetchMessagesForThread(thread).materialize())
+        .switchMap(bOrT -> bOrT.either(this.client::fetchMessagesForBacking, this.client::fetchMessagesForThread))
```
luoser: i'm in ❤️ this use of Either's `.either()`

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -75,129 +103,211 @@ public ViewModel(final @NonNull Environment environment) {
...
 
-      final Observable<MessageThreadEnvelope> messageThreadEnvelope = envelopeNotification.compose(values());
+      final Observable<Project> project = configData
+        .map(data -> data.either(MessageThread::project, projectAndBacking -> projectAndBacking.first));
 
-      final Observable<Notification<Backing>> backingNotification = Observable.combineLatest(
-        messageThread,
-        this.currentUser.observable(),
-        Pair::create
+      final Observable<User> participant = Observable.merge(
```
luoser: this logic grabs the project's creator if the VM was not initialized with a MessageThread...I'm not sure if this logic is a little too buried let me know if there are ways to improve!

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -75,129 +103,211 @@ public ViewModel(final @NonNull Environment environment) {
...
+        .compose(takeWhen(messageSent))
+        .compose(bindToLifecycle())
+        .subscribe(pc -> this.koala.trackSentMessage(pc.first, pc.second));
     }
 
+    private static @NonNull Observable<Pair<Backing, Project>> backingAndProjectFromData(final @NonNull MessagesData data,
+      final @NonNull ApiClientType client) {
+
+      return data.getBackingOrThread().either(
+        backing -> Observable.just(Pair.create(backing, data.getProject())),
```
luoser: FYI this `getSomething()` syntax is because `MessagesData` is a kotlin data class 🎈

app/src/main/res/layout/messages_layout.xml
```diff
@@ -69,37 +69,36 @@
         app:layout_scrollFlags="scroll|enterAlways">
 
         <RelativeLayout
-          android:layout_marginStart="@dimen/grid_new_3"
-          android:layout_marginEnd="@dimen/grid_new_3"
-          android:layout_marginLeft="@dimen/grid_new_3"
-          android:layout_marginRight="@dimen/grid_new_3"
+          android:paddingStart="@dimen/grid_new_3"
+          android:paddingEnd="0dp"
```
luoser: took the padding off the parent so the `View pledge` button's background ripple goes all the way to the edge

app/src/main/res/layout/messages_toolbar.xml
```diff
@@ -21,9 +21,13 @@
     android:orientation="horizontal">
 
     <com.kickstarter.ui.views.IconButton
+      android:id="@+id/messages_toolbar_back_button"
+      style="@style/ToolbarIconBackButton" />
```
luoser: I had to make two `IconButton`s, one for the `X` close button and one for `back arrow` button since it's very hairy to apply a `style` to a view programmatically :(

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
swoopej: agree

app/src/main/java/com/kickstarter/viewmodels/ViewPledgeViewModel.java
```diff
@@ -248,8 +255,18 @@ public ViewModel(final @NonNull Environment environment) {
...
-      this.viewMessagesButtonIsGone = this.currentConfig.observable()
-        .map(config -> !ObjectUtils.coalesce(config.features().get(FeatureKey.ANDROID_MESSAGES), false));
+      Observable.zip(
+        isFromMessagesActivity,
+        featureFlagIsEnabled,
+        Pair::create
+      )
+        .map(fromMessagesAndFlagEnabled -> fromMessagesAndFlagEnabled.first || !fromMessagesAndFlagEnabled.second)
+        .compose(bindToLifecycle())
+        .subscribe(this.viewMessagesButtonIsGone::onNext);
```
swoopej: just curious why we're switching from an `Observable` to a `BehaviorSubject` here and below?

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -75,129 +103,211 @@ public ViewModel(final @NonNull Environment environment) {
...
+          );
+        });
+
+      final Observable<KoalaContext.Message> koalaContext = intent()
+        .map(i -> i.getSerializableExtra(IntentKey.KOALA_CONTEXT))
+        .ofType(KoalaContext.Message.class);
+
+      final Observable<Backing> configBacking = configData
+        .map(Either::right)
+        .filter(ObjectUtils::isNotNull)
```
luoser: 👍 agreed!

luoser: ahh, but for places where we want to keep point free notation, we can't also use a static import

app/src/main/java/com/kickstarter/viewmodels/ViewPledgeViewModel.java
luoser: ah yes you have identified a weirdness/subtlety with our outputs that we're still trying to figure out-- 

our `isFromMessagesActivity` is a boolean that comes from our `intent()`, and we need the output to be a `BehaviorSubject` because for some reason an `Observable` doesn't emit immediately, due to our base `intent()` being a `PublishSubject`

tl;dr whenever you have an output that relies on data coming from the `intent()` it needs to be a `BehaviorSubject`

PR #114, opened 2017-06-15
Upgrade AS to 2.3.3

luoser:
## what
Bumping Android Studio to 2.3.3. Release notes here:
https://developer.android.com/studio/releases/index.html

Subsequently upgrading our Kotlin plugin to match our AS version.


swoopej:
bitmoji

PR #116, opened 2017-06-16
[Messages] Mark as read

luoser:
## what
Hello it's me again this time here to implement our `/read` endpoint for when you open a MessageThread and read the convo.

Also refactored some outputs in our MessageThreadsVM to clean up some UX (i.e. properly showing / hiding unread message counts).

## how
We use `SharedPreferences` similar to our cache in iOS to cache our `MessageThread` `hasUnreadMessages` boolean value, which is what determines the UI state (indicator, elevation, bolded text, etc) so we can immediately fetch from preferences what the unread state of a thread is. Pretty neato!

## todo
- [ ] There is a weird little bug that occurs sometimes when the MessageThreads view refreshes its current user and has to change the `(unread count)` visibility on the toolbar...will look into this more and fix Monday or in a separate bugs PR


app/src/main/java/com/kickstarter/ui/SharedPreferenceKey.java
```diff
@@ -0,0 +1,13 @@
+package com.kickstarter.ui;
+
+public final class SharedPreferenceKey {
```
luoser: technically we don't need this class of static key values since these keys are only used once...but it is nice to consolidate them in one place so we know what values we're storing in SharedPreferences at a glance!

app/src/main/java/com/kickstarter/viewmodels/MessageThreadsViewModel.java
```diff
@@ -58,9 +66,12 @@ public ViewModel(final @NonNull Environment environment) {
       this.client = environment.apiClient();
       this.currentUser = environment.currentUser();
 
-      final Observable<User> freshUser = this.client.fetchCurrentUser()
+      final Observable<User> freshUser = intent()
+        .compose(takeWhen(this.onResume))
```
luoser: this is the first place in which we've used `onResume()` as a lifecycle input!! 

the reasoning was that I wanted a signal similar to iOS's `viewDidAppear` to refresh the current user appropriately...the downside is that we'll be refreshing the user on rotation even, but I figured that if we tie many of our api calls to lifecycle events such as `viewWillAppear` and `viewDidAppear` in iOS this may be fine?

other option is to store a `BooleanPreference` on the environment which i'm chill with but wanted to try this approach and see how y'all feel

app/src/main/java/com/kickstarter/ui/SharedPreferenceKey.java
swoopej: 👌

app/src/main/java/com/kickstarter/viewmodels/MessageThreadHolderViewModel.java
```diff
@@ -88,7 +99,36 @@ public ViewModel(final @NonNull Environment environment) {
...
+      editor.putBoolean(cacheKey(messageThread), false);
+      editor.apply();
+    }
+
+    private static void setHasUnreadMessagesPreference(final @NonNull MessageThread messageThread,
+      final @NonNull SharedPreferences sharedPreferences) {
+
+      final SharedPreferences.Editor editor = sharedPreferences.edit();
+      editor.putBoolean(cacheKey(messageThread), messageThread.unreadMessagesCount() > 0);
+      editor.apply();
```
swoopej: just wanna confirm that we don't want to add `MESSAGE_THREAD_HAS_UNREAD_MESSAGES` to the initialization of other `SharedPreferences` in the ApplicationModule? Seems like we're just using the `sharedPreferences` here as key/value store for these threads, but wanted to confirm that's...ok.

app/src/main/java/com/kickstarter/viewmodels/MessageThreadsViewModel.java
swoopej: makes sense to me!

app/src/main/java/com/kickstarter/viewmodels/MessageThreadHolderViewModel.java
luoser: this is a very good question, i was looking at the way we do `sharedPreferences` in our `ProjectViewModel` to store ref tag cookies and figured it'd be fine to not make a new preference in the `ApplicationModule`, but seems like we do have a `provideCookieManager` and thus we should add this to the ApplicationModule to make sure it's provided correctly

good catch!!! 🐦

luoser: actually, scratch that for now. seems like we don't need to set this in the AppModule for message thread's unread messages, since it's a preference specific to each message thread id rather than a global preference. 

but this is still a valuable solution for setting a `hasUnreadMessages` boolean preference for the user that may avoid having to refresh the user onResume each  time so will explore that

swoopej: word

PR #117, opened 2017-06-19
[Messages] Enable/disable Send button

luoser:
## what
Enable the `Send` button when there is a message body present. Disable when the message body is empty and when the message is posting.

Bonus cleanups to CommentsVM and xml, since it uses similar logic.


app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java
```diff
@@ -82,7 +83,7 @@ public CommentsViewModel(final @NonNull Environment environment) {
       .share();
 
     final Observable<Boolean> commentHasBody = commentBodyChanged
-      .map(body -> body.length() > 0);
+      .map(StringUtils::isPresent);
```
swoopej: noice

PR #118, opened 2017-06-21
[Messages] Timestamps and UI fixes

luoser:
## what
* Grouping messages by start of day with a date timestamp.
* Added some basic logic for our `Delivered` indicator on a successfully sent message. (`Not delivered` is coming v2, for now we just show an error toast and the message does not get posted to the RecyclerView)
* Bonus fix / restructuring of how `message_view`s are displayed in the `MessageVH`: previously there was a bug where recipient/sender messages would anchor to the wrong sides of the screen upon recycling (scrolling through), but adding show/hide logic for permanently start/end anchored views keeps them in their correct places

## how
* Added new time formatters to our `DateUtils` to get correct DateTime formats
* Added a new `MessageCenterTimestampViewHolder.java` to hold the center timestamp
  * refactored `MessagesAdapter` to handle the grouping logic...pretty neato!
* anything else interesting i'll point out in the PR


app/src/main/java/com/kickstarter/ui/adapters/MessagesAdapter.java
```diff
@@ -7,30 +7,68 @@
...
+      return R.layout.message_center_timestamp_layout;
+    } else if (objectFromSectionRow(sectionRow) instanceof Message) {
+      return R.layout.message_view;
+    }
+    return R.layout.empty_view;
   }
 
   public void messages(final @NonNull List<Message> messages) {
-    setSection(0, messages);
+    // Group messages by start of day.
```
luoser: coolest thing about this chunk of code? i just refered to [our swift grouping](https://github.com/kickstarter/ios-oss/blob/master/Kickstarter-iOS/DataSources/ProjectActivitiesDataSource.swift#L23-L44) to do the same thing!!

app/src/main/java/com/kickstarter/ui/adapters/MessagesAdapter.java
```diff
@@ -7,30 +7,68 @@
...
+    return getLayoutId(sectionRow);
+  }
+
+  @Override
+  public void onBindViewHolder(final @NonNull KSViewHolder holder, final int position,
+    final @NonNull List<Object> payloads) {
+    super.onBindViewHolder(holder, position, payloads);
+
+    if (holder instanceof MessageViewHolder) {
+      ((MessageViewHolder) holder).isLastPosition(position == getItemCount() - 1);
```
luoser: we usually dont override `onBindViewHolder` in our `KSViewHolder`s, but i'm doing this here to let the VH know if it's in the last position or not to display the `Delivered` text view

app/src/main/java/com/kickstarter/ui/viewholders/MessageViewHolder.java
```diff
@@ -18,66 +16,59 @@
...
 import butterknife.Bind;
-import butterknife.BindDimen;
 import butterknife.ButterKnife;
 
 import static com.kickstarter.libs.rx.transformers.Transformers.observeForUI;
 
 public final class MessageViewHolder extends KSViewHolder {
   private final MessageHolderViewModel.ViewModel viewModel;
 
-  protected @Bind(R.id.message_body_card_view) CardView messageBodyCardView;
```
luoser: all of the changes, except setting the `deliveryStatusTextView` output, are related to the message view refactoring that correctly shows/hides recipient/sender message views rather than changing the alignment with an output

app/src/main/java/com/kickstarter/ui/viewholders/MessageViewHolder.java
```diff
@@ -86,19 +77,8 @@ public void bindData(final @Nullable Object data) throws Exception {
     this.viewModel.inputs.configureWith(message);
   }
 
-  private void alignMessageBodyCardView(final boolean alignParentEnd) {
-    final RelativeLayout.LayoutParams layoutParams = (RelativeLayout.LayoutParams) this.messageBodyCardView.getLayoutParams();
```
luoser: L89-94 didn't play well with the recycler view :(

app/src/main/java/com/kickstarter/libs/utils/DateTimeUtils.java
```diff
@@ -49,14 +49,14 @@ public static boolean isEpoch(final @NonNull DateTime dateTime) {
   }
 
   /**
-   * e.g.: Dec 17, 2015.
+   * e.g.: Tuesday, June 20, 2017
```
swoopej: oh chill ty for these updates

app/src/main/java/com/kickstarter/ui/adapters/MessagesAdapter.java
swoopej: maybe add a lil comment?

PR #120, opened 2017-06-22
Modernize PlaygroundVM

luoser:
## what
Giving our internal PlaygroundActivity and its VM some love, including a VMTest file.

For posterity: @christopherwright added this playground to our debug drawer a while back, let's keep it up to date to make it easier for us to test out new things!


luoser:
Hmmm so this failing test is due to `PlaygroundVMTest` living in the `/test` directory, while its `PlaygroundVM` lives in our `internal` package. Anyone have thoughts on how to easily re-package the test or vm file so that internal view models can have tests?

cc @christopherwright

christopherwright:
@luoser Ah, so the `internal` code stems from our [build flavors](https://github.com/kickstarter/android-oss/blob/master/app/build.gradle#L87-L92). Looking at [docs](http://tools.android.com/tech-docs/new-build-system/user-guide#TOC-Testing1), it seems like we may be able to add tests specific to a flavor – want to see if you can get it working, and ping me if it doesn't work? You may need to adapt those instructions for unit tests.

luoser:
not a real solution, removing the instantiation of the PlaygroundVM from tests, but don't want this to stay open forever. 

plus in reality writing playground tests will be mostly for our TDD / sanity checks and we won't be checking in actual playground code ever

PR #121, opened 2017-06-23
Dont show featured projects in subcategories

swoopej:
we are currently showing the featured project for a parent category as the first project in all of it's child, more narrow categories.  This has been determined to be a bad discovery experience!  sad!

small little fix to check if the category has no parent so we don't grab the featured project from api.  should there be a test for this? maybe! I'll try to figure that out.

cc @ktman 👋


luoser:
dope! re: test you can add a test to `DiscoveryParamsTest.java` that'd be swell

swoopej:
@luoser yeah ty figured it out in https://github.com/kickstarter/android-oss/pull/121/commits/9f98703197ef8d82f1e0254c79197d37f638bff2

PR #122, opened 2017-06-26
Fix project updates pagination

swoopej:
scrolling to the second page of project updates is currently broken - we make a request for the second page, but we don't actually end up showing new data.

I'm mostly just following the pattern that we use in the [checkout activity](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/ui/activities/CheckoutActivity.java#L190), but I think what's happening is we get multiple values emitted from [this merge](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/ProjectUpdatesViewModel.java#L70) in our ProjectUpdatesViewModel, so we don't actually load the data that is returned in the first emission.


swoopej:
oh good call liza added in https://github.com/kickstarter/android-oss/pull/122/commits/5e92db69204e8b5597e5a44d69b4bed54fa86d24

PR #124, opened 2017-06-30
Mark features as nullable

luoser:
## what
We have a bunch of crashes since our 1.4.5 release that I thought I had fixed via using:

* `coalesce` to coalesce `config.features().get` with `false` in case a config had not been refreshed properly
* [A default Builder value for features to avoid using Nullable](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/libs/Config.java#L55)

But after some debugging it turns out that the default AutoParcel initialization isn't doing what we expected, i.e. if `features` is null for a non refreshed config we see the crashes in our feature flagged places. This is a hotfix for now so we can move on with making external builds, but for future notice we need to upgrade our AutoParcel to AutoValue.

bitmoji

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/Config.java
+++ b/app/src/main/java/com/kickstarter/libs/Config.java
@@ -2,6 +2,7 @@ package com.kickstarter.libs;
 
 import android.os.Parcelable;
 import android.support.annotation.NonNull;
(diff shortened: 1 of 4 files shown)


luoser:
bitmoji

PR #126, opened 2017-07-06
[Messages] Push notifications

luoser:
## what
Woohoo push notifications for messages. Tapping on the notification opens the message thread.

image

## todo
I'd like to do some more IRL device testing (I think my notifications are getting batched right now)...but some review on the general gist of this would be great, since I just followed along with our other push infra to do this 😁


luoser:
for posterity: [this change](https://github.com/kickstarter/android-oss/pull/126/commits/78f3c9d94b4a89323ac8de234b1541d36fb72e7c#diff-24426560fc743e20683b8e3cf9204e5eR46) was made because we weren't serializing our new `PushNotificationEnvelope.Message` correctly without it, which means we also caught the same for our `PushNotificationEnvelope.Project` reminder pushes 😟 , so i've made a card for the future to make this class more testable, but merging for now!

swoopej:
cool cool

PR #129, opened 2017-07-11
[Messages] More bugfixes

luoser:
## what
* Preserve toolbar expanded state on rotation, done by setting the id of its CoordinatorLayout
* Open keyboard when there are no messages in a thread
* Add `MessageSubject` & refactor MessagesVM to use it! (this VM is where most of the surgery is done)
  * Previously, I was sending messages to a Backing thread as a backer and thus received 422s, since backers should be sending to a Project thread. The overall flow of our messages via message subjects is
  ```
  MessageSubject.Backing : creator -> backer
  MessageSubject.Project : backer / user -> creator
  MessageSubject.MessageThread : user to a message thread, navigated from Inbox
  ```

I missed this logic the first time through because in Swift, we have a `MessagesViewModel` as well as a `MessageDialogViewModel` since our editor for writing messages is its own view controller.

## todo
It's been a journey and what's left now is ironing out the expanded toolbar UX for MessageThreads, then more rounds of bugfixes if necessary 😎

### DIFF
--- a/app/src/main/java/com/kickstarter/services/ApiClient.java
+++ b/app/src/main/java/com/kickstarter/services/ApiClient.java
@@ -388,14 +388,16 @@ public final class ApiClient implements ApiClientType {
 
   @Override
   public @NonNull Observable<Message> sendMessage(final @NonNull MessageSubject messageSubject, final @NonNull String body) {
+    final MessageBody messageBody = MessageBody.builder().body(body).build();
+
     return messageSubject
       .value(
         backing -> this.service
-          .sendMessageToBacking(backing.projectId(), backing.backerId(), MessageBody.builder().body(body).build()),
+          .sendMessageToBacking(backing.projectId(), backing.backerId(), messageBody),
         messageThread -> this.service
-          .sendMessageToThread(messageThread.id(), MessageBody.builder().body(body).build()),
+          .sendMessageToThread(messageThread.id(), messageBody),
         project -> this.service
-          .sendMessageToProject(project.id(), MessageBody.builder().body(body).build())
+          .sendMessageToProject(project.id(), messageBody)
       )
       .lift(apiErrorOperator())
       .subscribeOn(Schedulers.io());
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/kickstarter/services/ApiClient.java
```diff
@@ -386,17 +387,16 @@ public ApiClient(final @NonNull ApiService service, final @NonNull Gson gson) {
...
-
-  @Override
-  public @NonNull Observable<Message> sendMessageToThread(final @NonNull MessageThread messageThread, final @NonNull String body) {
-    return service
-      .sendMessageToThread(messageThread.id(), MessageBody.builder().body(body).build())
+  public @NonNull Observable<Message> sendMessage(final @NonNull MessageSubject messageSubject, final @NonNull String body) {
+    return messageSubject
+      .value(
+        backing -> this.service
+          .sendMessageToBacking(backing.projectId(), backing.backerId(), MessageBody.builder().body(body).build()),
```
stephencelis: Can `MessageBody.builder().body(body).build()` go above to make these three cases shorter?

app/src/main/java/com/kickstarter/ui/data/MessageSubject.kt
```diff
@@ -0,0 +1,17 @@
+package com.kickstarter.ui.data
+
+sealed class MessageSubject {
+  class Backing(val backing: com.kickstarter.models.Backing) : MessageSubject()
+  class Project(val project: com.kickstarter.models.Project) : MessageSubject()
+  class MessageThread(val messageThread: com.kickstarter.models.MessageThread) : MessageSubject()
+
+  fun <A> value(
```
stephencelis: Fun!

app/src/main/java/com/kickstarter/services/ApiClient.java
luoser: yesssss

app/src/main/java/com/kickstarter/ui/data/MessageSubject.kt
luoser: image

PR #130, opened 2017-07-13
Creator tools v01

swoopej:
same tools, now with a fre$h new look and a structure using viewholders and an adapter that will allow us to more easily add the sections below the headline project stats.  also, a new `ProjectStats` model.

new look:
screen shot 2017-07-13 at 3 20 04 pm


app/src/main/java/com/kickstarter/viewmodels/inputs/CreatorDashboardHeaderHolderViewModelInputs.java
```diff
@@ -0,0 +1,10 @@
+package com.kickstarter.viewmodels.inputs;
+
+import com.kickstarter.models.Project;
+import com.kickstarter.models.ProjectStats;
+
+
+public interface CreatorDashboardHeaderHolderViewModelInputs {
```
luoser: ahh yes we're trying to move away from creating separate files for Inputs and Outputs which is why our ViewModels are interfaces, you can move these into the same file as `CreatorDashboardHeaderHolderViewModel`

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModel.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private final PublishSubject<Pair<Project, ProjectStats>> projectAndStats = PublishSubject.create();
+    private final PublishSubject<Void> projectViewClicked = PublishSubject.create();
+
+    private final BehaviorSubject<String> percentageFunded = BehaviorSubject.create();
+    private final BehaviorSubject<Project> currentProject = BehaviorSubject.create();
+    private final BehaviorSubject<String> projectBackersCountText = BehaviorSubject.create();
+    private final BehaviorSubject<String> projectNameTextViewText = BehaviorSubject.create();
+    private final BehaviorSubject<Pair<Project, RefTag>> startProjectActivity = BehaviorSubject.create();
+    private final BehaviorSubject<String> timeRemainingText = BehaviorSubject.create();
```
luoser: since these outputs aren't dependent on a value coming through from the `intent()`, these outputs can actually all be `Observable`s. can chat IRL about this since it's a confusing lil gotcha about our codebase

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModel.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Override
+    public void projectViewClicked() {
+      this.projectViewClicked.onNext(null);
+    }
+
+    @Override
+    public void projectAndStats(final Project project, final ProjectStats projectStats) {
```
luoser: these params can be marked `@NonNull`

app/src/main/java/com/kickstarter/ui/viewholders/CreatorDashboardHeaderViewHolder.java
```diff
@@ -0,0 +1,131 @@
...
+    this.viewModel.inputs.projectViewClicked();
+  }
+
+  @Override
+  public void bindData(final @Nullable Object data) throws Exception {
+    // coerce to projectstats and project
+    final Pair<Project, ProjectStats> projectAndProjectStats = requireNonNull((Pair<Project, ProjectStats>) data);
+    final Project project = requireNonNull(projectAndProjectStats.first, Project.class);
+    final ProjectStats projectStats = requireNonNull(projectAndProjectStats.second, ProjectStats.class);
+    viewModel.inputs.projectAndStats(project, projectStats);
```
luoser: since we're using these values as a pair anyways in the VM, we can just go ahead and make the `projectAndStats` input a `Pair<Project, ProjectStats>` type so we're not destructuring / restructuring the pair!

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModel.java
```diff
@@ -0,0 +1,112 @@
...
+
+  final class ViewModel extends ActivityViewModel<CreatorDashboardHeaderViewHolder> implements
+    CreatorDashboardHeaderHolderViewModelInputs, CreatorDashboardHeaderHolderViewModelOutputs {
+
+
+    public ViewModel(final @NonNull Environment environment) {
+      super(environment);
+
+      projectAndStats
+        .map(ps -> ps.first)
```
luoser: we have a PairUtils class so we can go point free with this, e.g. `.map(PairUtils::first)`

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardViewModel.java
```diff
@@ -65,31 +59,25 @@ public ViewModel(final @NonNull Environment environment) {
       final Observable<Project> latestProject = projects
         .map(ListUtils::first);
 
+      final Observable<ProjectStats> projectStats = latestProject
+        .switchMap(this.client::fetchProjectStats);
```
luoser: sweeet. for our api requests, we do have to handle the error state and call `.share()` to prevent too many requests from happening. we can handle the error by calling `.materialize()` after the request is made, which then gives us an `Observable<Notification<ProjectStats>>` that then holds either `values()` or `error()`

heres an example of [materialize and share in action](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java#L95-L97) and here's how we can [extract the values](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java#L99-L100) and the [errors](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.java#L139-L142)

app/src/main/java/com/kickstarter/viewmodels/outputs/CreatorDashboardHeaderHolderViewModelOutputs.java
```diff
@@ -0,0 +1,29 @@
...
+
+
+import android.util.Pair;
+
+import com.kickstarter.libs.RefTag;
+import com.kickstarter.models.Project;
+
+import rx.Observable;
+
+public interface CreatorDashboardHeaderHolderViewModelOutputs {
```
luoser: same with these outputs, can be moved to the ViewModel file directly

app/src/main/res/layout/dashboard_funding_view.xml
```diff
@@ -0,0 +1,123 @@
...
+  xmlns:tools="http://schemas.android.com/tools">
+
+  <TextView
+    android:id="@+id/creator_dashboard_project_name"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:paddingTop="@dimen/grid_new_2"
+    android:paddingLeft="@dimen/grid_new_2"
+    style="@style/Title3"
+    tools:text="ayyyy it's shrek" />
```
luoser: lol

app/src/main/res/layout/dashboard_funding_view.xml
```diff
@@ -0,0 +1,123 @@
...
+    android:gravity="center_horizontal"
+    android:layout_marginBottom="@dimen/grid_new_3"
+    style="@style/BodyPrimary"
+    tools:text="pledged of $12,000"/>
+
+  <LinearLayout
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:orientation="horizontal"
+    android:paddingLeft="@dimen/grid_new_5"
```
luoser: in generale we're moving away from `Left` and `Right` to use `Start` and `End` instead since L and R are now deprecated

app/src/main/res/layout/dashboard_funding_view.xml
```diff
@@ -0,0 +1,123 @@
...
+    android:id="@+id/creator_view_project_button"
+    style="@style/BlueRoundedButton"
+    android:layout_width="match_parent"
+    android:layout_height="@dimen/grid_new_7"
+    android:layout_marginTop="@dimen/grid_new_2"
+    android:layout_marginRight="@dimen/grid_new_5"
+    android:layout_marginLeft="@dimen/grid_new_5"
+    android:text="@string/View"
+    tools:text="Go to project" />
+</LinearLayout>
```
luoser: NEWLIEN

app/src/test/java/com/kickstarter/factories/ProjectStatsFactory.java
```diff
@@ -0,0 +1,122 @@
...
+      .rewardStats()
+      .toBuilder()
+      .build();
+
+    final ProjectStats.VideoStats videoStats = VideoStatsFactory
+      .videoStats()
+      .toBuilder()
+      .build();
+
+    final List<ProjectStats.RewardStats> rewardStatsList = Collections.unmodifiableList(new ArrayList<ProjectStats.RewardStats>(Arrays.asList(rewardStats)));
```
luoser: oooooo why an `unmodifiableList` here?

app/src/test/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModelTest.java
```diff
@@ -0,0 +1,85 @@
...
+  private CreatorDashboardHeaderHolderViewModel.ViewModel vm;
+
+  private final TestSubscriber<String> percentageFunded = new TestSubscriber<>();
+  private final TestSubscriber<String> projectBackersCountText = new TestSubscriber<>();
+  private final TestSubscriber<String> projectNameTextViewText = new TestSubscriber<>();
+  private final TestSubscriber<String> timeRemainingText = new TestSubscriber<>();
+  private final TestSubscriber<Pair<Project, RefTag>> startProjectActivity = new TestSubscriber<>();
+
+  protected void setUpEnvironment(final @NonNull Environment environment) {
+    this.vm = new CreatorDashboardHeaderHolderViewModel.ViewModel(environment);
```
luoser: in our `setUpEnvironment` is where we hook up our `TestSubscribers` to our view model outputs, i.e.

```java
this.vm.outputs.backButtonIsGone().subscribe(this.backButtonIsGone);
```

app/src/test/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModelTest.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Test
+  public void testProjectBackersCountText() {
+    final Project project = ProjectFactory.project().toBuilder().backersCount(10).build();
+    final ProjectStats projectStats = ProjectStatsFactory.projectStats();
+    this.vm = new CreatorDashboardHeaderHolderViewModel.ViewModel(environment());
```
luoser: rather than setting the vm in each test, let's call `setUpEnvironment(environment())` this instantiates the VM and also our test subscribers

app/src/test/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModelTest.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Test
+  public void testPercentageFunded() {
+    final Project project = ProjectFactory.project();
+    final ProjectStats projectStats = ProjectStatsFactory.projectStats();
+    this.vm = new CreatorDashboardHeaderHolderViewModel.ViewModel(environment());
+    this.vm.outputs.percentageFunded().subscribe(this.percentageFunded);
+    this.vm.inputs.projectAndStats(project, projectStats);
+    final String percentageFundedOutput = NumberUtils.flooredPercentage(project.percentageFunded());
```
luoser: general syntax in tests is to declare all our variables at the top. makes it a lil more straightforward to read through a test script's assertions

app/src/main/java/com/kickstarter/services/ApiService.java
```diff
@@ -137,6 +138,9 @@
   @GET
   Observable<Response<DiscoverEnvelope>> projects(@Url String paginationUrl);
 
+  @GET("/v1/projects/{project_id}/stats")
+  Observable<Response<ProjectStats>> projectStats(@Path("project_id") String projectId);
```
christopherwright: Given the param is a string, it may be more accurate to refer to it as a `project_param` rather than a `project_id`. Am assuming we could either use the slug or pid.

app/src/main/java/com/kickstarter/ui/viewholders/CreatorDashboardHeaderViewHolder.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  private void setPledgedOfGoalString(final @NonNull Project latestProject) {
```
christopherwright: I assume the `latestProject` name is emphasizing that we don't have a project switcher in place yet, so it's grabbing the creator's most recent project?

I suspect for a lot of this code, the source of the project isn't super important though – we could just call the variable `project`. When we have a switcher in place, none of this code would need to be renamed.

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModel.java
christopherwright: Could add some `@NonNull` annotations here.

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModel.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+
+    public final CreatorDashboardHeaderHolderViewModelInputs inputs = this;
+    public final CreatorDashboardHeaderHolderViewModelOutputs outputs = this;
+
+    private final PublishSubject<Pair<Project, ProjectStats>> projectAndStats = PublishSubject.create();
+    private final PublishSubject<Void> projectViewClicked = PublishSubject.create();
+
+    private final BehaviorSubject<String> percentageFunded = BehaviorSubject.create();
```
christopherwright: Could these `BehaviorSubject` types be replaced with `Observable`?

app/src/main/java/com/kickstarter/ui/viewholders/CreatorDashboardHeaderViewHolder.java
```diff
@@ -0,0 +1,131 @@
...
+  public CreatorDashboardHeaderViewHolder(final @NonNull View view) {
+    super(view);
+
+    viewModel = new CreatorDashboardHeaderHolderViewModel.ViewModel(environment());
+    ButterKnife.bind(this, view);
+
+    this.ksCurrency = this.environment().ksCurrency();
+    this.ksString = this.environment().ksString();
+
+    viewModel.outputs.currentProject()
```
christopherwright: I wonder if some of this code needs to be updated or deleted? We're subscribing to the project rather than `timeRemainingText`, seems like we're not using the correct output. Same applies for the pledged of goal output below.

app/src/main/java/com/kickstarter/ui/viewholders/CreatorDashboardHeaderViewHolder.java
```diff
@@ -0,0 +1,131 @@
...
+      .compose(bindToLifecycle())
+      .compose(observeForUI())
+      .subscribe(this::setPledgedOfGoalString);
+
+    viewModel.outputs.currentProject()
+      .compose(bindToLifecycle())
+      .compose(observeForUI())
+      .subscribe(this::setPledgedOfGoalString);
```
christopherwright: Looks like we have a duplicated subscription here (same as lines 62-65?)

app/src/main/java/com/kickstarter/viewmodels/inputs/CreatorDashboardHeaderHolderViewModelInputs.java
christopherwright: I believe @luoser has been moving to a new pattern of nesting the inputs and outputs inside the viewmodel, e.g.: https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/MessageHolderViewModel.java#L19-L48

app/src/main/res/layout/dashboard_funding_view.xml
christopherwright: So sad we're no longer defaulting to `SKULL GRAPHIC TEE`

app/src/main/res/layout/dashboard_funding_view.xml
christopherwright: Wonder if there's some new line setting that we should apply for consistency? I believe we used to ensure there was a new line at the end of every file, maybe there's a linter rule for this we could add

app/src/main/java/com/kickstarter/ui/viewholders/CreatorDashboardHeaderViewHolder.java
swoopej: ah yeah those are just helper methods down below that take a project.  I would make the output more specific but `ksCurrency.format` and `ProjectUtils.deadlineCountdownDetail` take a project object unless I'm missing an easier way

app/src/main/java/com/kickstarter/ui/viewholders/CreatorDashboardHeaderViewHolder.java
swoopej: word ty ty

swoopej:
alright friends all the feedback should be addressed in https://github.com/kickstarter/android-oss/pull/130/commits/4e8091418a0f28826d7813115b79c10723ce075c

thanks for the thorough reviewing!

app/src/main/java/com/kickstarter/ui/adapters/CreatorDashboardAdapter.java
```diff
@@ -0,0 +1,33 @@
...
+
+  protected @LayoutRes int layout(final @NonNull SectionRow sectionRow) {
+    return R.layout.dashboard_funding_view;
+  }
+
+  protected @NonNull KSViewHolder viewHolder(final @LayoutRes int layout, final @NonNull View view) {
+    return new CreatorDashboardHeaderViewHolder(view);
+  }
+
+
```
luoser: lil newline

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModel.java
```diff
@@ -0,0 +1,128 @@
...
+    private final Observable<String> timeRemainingText;
+
+    @Override
+    public void projectViewClicked() {
+      this.projectViewClicked.onNext(null);
+    }
+
+    @Override
+    public void projectAndStats(final @NonNull Project project, final @NonNull ProjectStatsEnvelope ProjectStatsEnvelope) {
+      this.projectAndStats.onNext(Pair.create(project, ProjectStatsEnvelope));
```
luoser: since we're just making a new pair here, we can change the `projectAndStats` input type above on line 24 to just be a `Pair<Project, Stats>`, and input it directly from the [ViewHolder](https://github.com/kickstarter/android-oss/pull/130/files#diff-f8896b2518d63fc0f3f5fbad2e11e15eR102) since we have the 🍐 there

app/src/main/res/layout/dashboard_funding_view.xml
```diff
@@ -0,0 +1,123 @@
...
+  android:layout_width="match_parent"
+  android:orientation="vertical"
+  xmlns:tools="http://schemas.android.com/tools">
+
+  <TextView
+    android:id="@+id/creator_dashboard_project_name"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:paddingTop="@dimen/grid_new_2"
+    android:paddingStart="@dimen/grid_new_2"
```
luoser: we should add a `android:paddingEnd="@dimen/grid_new_2"` here too for long project names, so there's even spacing on both ends

app/src/main/res/layout/dashboard_funding_view.xml
```diff
@@ -0,0 +1,123 @@
...
+        android:id="@+id/creator_dashboard_time_remaining_text"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        tools:text="secs to go"/>
+
+    </LinearLayout>
+
+  </LinearLayout>
+
+
```
luoser: lil newline

PR #132, opened 2017-07-20
[Messages] Render unread state correctly on refresh

luoser:
## what
There was a lil bug with rendering the correct `unread` state for messages on resume.

## how
Turns out since we weren't passing down the correctly refreshed `List<MessageThread>` data from the parent `MessageThreadsViewModel` to its `MessageThreadHolderViewModel`s, so despite caching, the previous value would be fed through again on rotation and recycling (i.e. `onResume`). Wahoo!

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/MessageThreadsViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/MessageThreadsViewModel.java
@@ -77,8 +77,10 @@ public interface MessageThreadsViewModel {
       this.client = environment.apiClient();
       this.currentUser = environment.currentUser();
 
+      final Observable<Void> startOverWith = Observable.merge(this.onResume, this.refresh);
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/viewmodels/MessageThreadsViewModel.java
```diff
@@ -77,8 +77,10 @@ public ViewModel(final @NonNull Environment environment) {
       this.client = environment.apiClient();
       this.currentUser = environment.currentUser();
 
+      final Observable<Void> startOverWith = Observable.merge(this.onResume, this.refresh);
```
luoser: for posterity `startOverWith` is a signal that we typically use for our paginators that indicate when we should fetch new data

swoopej: oh nice

PR #137, opened 2017-07-25
[Messages] QA round dos

luoser:
## What
Feedback from our QA session, fixed
1. Hide backing info banner if no backing [card](https://trello.com/c/oYTOEtzX/738-p2-view-pledge-cta-shows-in-thread-view-even-if-you-arent-a-backer)
2. Rend sent message to a new message thread [card](https://trello.com/c/G1jTflJY/737-p2-if-you-send-a-new-message-to-a-thread-from-your-backing-info-message-content-doesnt-show-in-the-screen-youre-on)
3. Let reply expand to >1 line [card](https://trello.com/c/5PWLTfJY/736-when-message-goes-onto-2nd-line-see-last-line-of-text-and-only-bottom-of-characters-on-line-above)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
@@ -1,6 +1,7 @@
 package com.kickstarter.viewmodels;
 
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.util.Pair;
 
 import com.kickstarter.libs.ActivityViewModel;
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -250,18 +251,20 @@ public ViewModel(final @NonNull Environment environment) {
...
 
       final Observable<List<Message>> newMessages = sentMessageThreadEnvelope
         .map(MessageThreadEnvelope::messages);
 
+      // Concat distinct messages to initial message list. Return just the new messages if
+      // initial list is null, i.e. a new message thread.
       final Observable<List<Message>> updatedMessages = initialMessages
         .compose(takePairWhen(newMessages))
-        .map(mm -> ListUtils.concatDistinct(mm.first, mm.second));
+        .map(mm -> mm.first == null ? mm.second : ListUtils.concatDistinct(mm.first, mm.second));
```
swoopej: is the null check here redundant now that you're filtering for `isNotNull` on `initialMessages` below?

luoser: ah, the null check here is for the `concatDistinct`; `mm.first` is the `initialMessages` list so it's checking for a null message list

if that's not clear i can update my comment above or make `mm` better!

swoopej: oooooh ya makes sense 👌

PR #138, opened 2017-07-25
Use setCurrentMillisFixed for timing test????

swoopej:
joda time has [a method to freeze time in it's current state](http://www.joda.org/joda-time/apidocs/org/joda/time/DateTimeUtils.html#setCurrentMillisFixed-long-).  Maybe this will solve our flaky timing test? Maybe there is a better place to do this like in setup/teardown blocks, but we only need it for one test right now.

anyway, here's a dance about millys

giphy

### DIFF
--- a/app/src/test/java/com/kickstarter/KSRobolectricTestCase.java
+++ b/app/src/test/java/com/kickstarter/KSRobolectricTestCase.java
@@ -13,6 +13,9 @@ import com.kickstarter.services.MockWebClient;
 
 import junit.framework.TestCase;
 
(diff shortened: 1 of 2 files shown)


christopherwright:
It'd be neat to try this for all tests – same thing that we do in the backend with Timecop. Otherwise I can see this issue popping up again in the future..

swoopej:
@christopherwright got a link to what we do with Timecop?

swoopej:
@christopherwright added the time freeze to the global setup/teardown in our base test case.

PR #139, opened 2017-07-26
Consolidate toolbar offsetlisteners

luoser:
## what
Got [this crash in HockeyApp](https://rink.hockeyapp.net/manage/apps/239008/app_versions/81/crash_reasons/178790844?type=crashes) and while I'm not able to re-pro the bug, my hunch is that setting 2 offset listeners on the same `AppBarLayout` in `MessagesActivity` is causing a rendering race condition.

For this instance, in which we do want the toolbar fade effect that `ToolbarUtils` provides us, but also additional logic from listening to the offset changes, let's consolidate the listeners for peace of mind.

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/MessagesActivity.java
+++ b/app/src/main/java/com/kickstarter/ui/activities/MessagesActivity.java
@@ -22,7 +22,6 @@ import com.kickstarter.libs.KSCurrency;
 import com.kickstarter.libs.KSString;
 import com.kickstarter.libs.qualifiers.RequiresActivityViewModel;
 import com.kickstarter.libs.utils.DateTimeUtils;
-import com.kickstarter.libs.utils.ToolbarUtils;
 import com.kickstarter.libs.utils.TransitionUtils;
 import com.kickstarter.libs.utils.ViewUtils;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/kickstarter/ui/activities/MessagesActivity.java
```diff
@@ -236,6 +230,20 @@ private void requestFocusAndOpenKeyboard() {
...
   }
 
+  /**
+   * Sets an OffsetChangedListener for the view's AppBarLayout to:
+   * 1. determine the toolbar's alpha based on scroll range
+   * 2. adjust the view's bottom padding via inputs
+   */
+  private void setAppBarOffsetChangedListener(final @NonNull AppBarLayout appBarLayout) {
+    appBarLayout.addOnOffsetChangedListener((layout, offset) -> {
+      this.projectNameToolbarTextView.setAlpha(Math.abs(offset) / layout.getTotalScrollRange());
```
luoser: FYI this line is just the same as [what we use in our ToolbarUtils](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/libs/utils/ToolbarUtils.kt#L15)

PR #141, opened 2017-07-27
Remove flag from backing screen

luoser:
Finishing what I started in #140, since we also show a `Message creator` button in our Backing info view, which would be nice to release since it's ya know already there

Refresher: we still do need to hide the button if we're navigating to the view from Messages, to avoid a weird loop of Backing -> Messages -> Backing -> Messages


luoser:
rush merging --- 100% my responsibility if something goes terribly

PR #142, opened 2017-07-31
[Messages] Loading indicators

luoser:
## what
Added a progress spinner to our message thread view since longer threads can take a while to load.

indicators

Had one on the threads Inbox view as well until I realized that having that spinner as well as a swipe-refresh spinner was odd UX. We'll revisit that one.


app/src/main/java/com/kickstarter/ui/activities/ProfileActivity.java
```diff
@@ -136,7 +136,7 @@ protected void onDestroy() {
 
   @OnClick(R.id.messages_button)
   public void messagesButtonClicked() {
-    this.viewModel.inputs.messsagesButtonClicked();
+    this.viewModel.inputs.messagesButtonClicked();
```
swoopej: 😏

PR #149, opened 2017-08-09
this, part 3

luoser:
## what
Implement [RequireThisCheck](http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/RequireThisCheck.html) in our checkstyle, catch final errors. Can use actual 👀  on this one to make sure I didn't mess anything up!

## interesting things

1.
`RequireThisCheck` isn't smart enough to sweep for point-free invoked methods in RxJava, so for a few outputs where we have the same output name as, say, an envelope's values e.g.:

```
this.messages
```

is an output in a file where we also invoke

```
MessageThreadEnvelope::messages
```

the checkstyle would flag the second line for linting.

For these instances, I simply renamed `messages` to `messageList`, as it's more descriptive of the data structure anyways.

2.
We're also not allowed to annotate instance variables with `this` (even though they are still instance variables) in new data structure declaration, e.g.

```
return new HashMap<String, Object>() {
   put("hello", this.someObject)
};
```

doesn't fly so I've chosen to make seemlessly pointless intermediate vars to satisfy the check. If there's a better way lmk but seems fine and I don't see it happening often.


script/style/checkstyle.xml
```diff
@@ -40,6 +40,9 @@
     <module name="NoFinalizer" />
     <module name="OneStatementPerLine" />
     <module name="ParameterAssignment" />
+    <module name="RequireThisCheck">
+      <property name="checkMethods" value="false"/>
```
luoser: leaving the `checkMethods` part of this as false for now, we can flip to true and audit again in the future but this took enough busy time 😬

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
```diff
@@ -53,7 +53,7 @@ public void onActivityResumed(final @NonNull Activity activity) {
       this.client.config()
         .compose(Transformers.pipeApiErrorsTo(this::handleConfigApiError))
         .compose(Transformers.neverError())
-        .subscribe(this.config::config);
+        .subscribe(c -> this.config.config(c));
```
swoopej: is this one of the edge cases that RequireThisCheck doesn't handle or is it just some debuggering

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModel.java
```diff
@@ -56,9 +56,7 @@ public ViewModel(final @NonNull Environment environment) {
 
       this.percentageFunded = this.projectAndStats
         .map(PairUtils::first)
-        .map(Project::percentageFunded)
-        .map(NumberUtils::flooredPercentage)
-        .compose(bindToLifecycle());
+        .map(p -> NumberUtils.flooredPercentage(p.percentageFunded()));
```
swoopej: still want `compose(bindToLifeCycle)` here?

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
luoser: it's an edge check because there's a `this.config` and a `config()` refresh method that it gets confused by :/

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardHeaderHolderViewModel.java
luoser: nope, we don't need to `bindToLifecycle` when it's not a subscription to a subject i.e. when the output is an observable !

PR #151, opened 2017-08-28
Surveys

luoser:
## what
Wrapping up core surveys feature. Challenge here was figuring out how to properly handle our redirect from `/surveys` to `/project` on a successful form submit, since we don't have (easy) access to navigation types in android WebViews 😭. Turns out the solution was actually not bad at all, since our `Request` gives us `tag` info. 

Now for some slow web view / emulator action:

_Unsuccessful submit:_
surveys-bad

_Successful submit:_
surveys-good

## todo before feature release
- [ ] Push Notifications
- [x] Refresh on swipe refresh from activity feed


app/src/main/java/com/kickstarter/viewmodels/SurveyResponseViewModel.java
```diff
@@ -0,0 +1,155 @@
...
+        .subscribe(p -> this.startProjectActivity.onNext(Pair.create(p, RefTag.survey())));
+
+      this.goBack = this.okButtonClicked;
+    }
+
+    /**
+     * Parses a project request for project params.
+     */
+    private @NonNull String extractProjectParams(final @NonNull Request request) {
+      return request.url().encodedPathSegments().get(2);
```
swoopej: you probably already thought about this, but could the list returned by `encodedPathSegments` ever be `null` or a length shorter than two?

swoopej: adding idk if there is anything to be done about it even if there is.

app/src/main/java/com/kickstarter/ui/activities/SurveyResponseActivity.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  private boolean handleProjectUriRequest(final @NonNull Request request, final @NonNull WebView webView) {
+    this.viewModel.inputs.projectUriRequest(request);
+    return true;
+  }
+
+  private boolean handleProjectSurveyUriRequest(final @NonNull Request request, final @NonNull WebView webView) {
+    this.viewModel.inputs.projectSurveyUriRequest(request);
+    return false;
```
swoopej: doesn't need to be addressed here but these two booleans along with the ones in `KSWebViewClient.handle_request` are still confusing to me.

app/src/main/java/com/kickstarter/viewmodels/SurveyResponseViewModel.java
luoser: you are quite right, this definitely is not an ideal way to handle parsing project params. fwiw it's a todo on our UpdateViewModel as well, to build a Navigation helper for safer extraction that i'd like to make eventually. for now it's ok because our request handler pretty much guarantees that a project request is coming through but seeing this `.get(2)` is 😨

app/src/main/java/com/kickstarter/ui/activities/SurveyResponseActivity.java
luoser: returning `false` here tells the web view client that we haven't handled the survey uri request yet, which allows for re-handling the request e.g. reloading the url on an unsuccessful form submission...but yeah it is kinda confusing we should talk irl and debate intelligently about it on our vision quest

PR #156, opened 2017-09-05
attach new fragment when project switcher is clicked in dashboard

swoopej:
hello, attaching an onclick and updating the fragment when the project switcher is clicked.  here's a look:

project_switcher

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/CreatorDashboardActivity.java
+++ b/app/src/main/java/com/kickstarter/ui/activities/CreatorDashboardActivity.java
@@ -6,9 +6,9 @@ import android.support.annotation.Nullable;
 import android.support.design.widget.BottomSheetBehavior;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentTransaction;
-import android.util.Pair;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
+import android.util.Pair;
 
 import com.kickstarter.R;
 import com.kickstarter.libs.BaseActivity;
@@ -41,7 +41,7 @@ public final class CreatorDashboardActivity extends BaseActivity<CreatorDashboar
     ButterKnife.bind(this);
 
     // Set up the bottom sheet recycler view.
-    this.bottomSheetAdapter = new CreatorDashboardBottomSheetAdapter();
+    this.bottomSheetAdapter = new CreatorDashboardBottomSheetAdapter(this.viewModel.inputs);
     this.bottomSheetRecyclerView.setAdapter(this.bottomSheetAdapter);
     this.bottomSheetRecyclerView.setLayoutManager(new LinearLayoutManager(this)); // todo: reuse LayoutManager?
     this.bottomSheetBehavior = BottomSheetBehavior.from(this.bottomSheetRecyclerView);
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/kickstarter/ui/activities/CreatorDashboardActivity.java
```diff
@@ -41,31 +41,41 @@ protected void onCreate(final @Nullable Bundle savedInstanceState) {
...
+    this.viewModel.refreshProject(project);
+    this.bottomSheetBehavior.setState(BottomSheetBehavior.STATE_COLLAPSED);
+  }
+
+  private void createProjectDashboardFragment(final @NonNull Pair<Project, ProjectStatsEnvelope> projectAndStats) {
     final FragmentManager fragmentManager = getSupportFragmentManager();
     final FragmentTransaction fragmentTransaction = fragmentManager.beginTransaction();
     final CreatorDashboardFragment fragment = CreatorDashboardFragment.newInstance(projectAndStats);
-    fragmentTransaction.add(R.id.creator_dashboard_coordinator_view, fragment);
+    fragmentTransaction.replace(R.id.creator_dashboard_coordinator_view, fragment);
```
swoopej: this works even if there is no fragment in the transaction fortunately

app/src/main/java/com/kickstarter/ui/activities/CreatorDashboardActivity.java
```diff
@@ -41,31 +41,41 @@ protected void onCreate(final @Nullable Bundle savedInstanceState) {
...
+      .compose(observeForUI())
+      .subscribe(this::newProjectClicked);
   }
 
   private void setProjectsForDropdown(final @NonNull List<Project> projects) {
     this.bottomSheetAdapter.takeProjects(projects);
   }
 
-  private void createProjectDashboardFragments(final @NonNull Pair<Project, ProjectStatsEnvelope> projectAndStats) {
+  private void newProjectClicked(final @NonNull Project project) {
```
luoser: mayhaps `projectSwitcherProjectClicked` would be a lil more specific (like your output name!)

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardBottomSheetHolderViewModel.java
```diff
@@ -25,23 +28,35 @@ public ViewModel(final @NonNull Environment environment) {
       super(environment);
 
       this.projectNameText = this.currentProject.map(Project::name);
+
+      this.projectSwitcherProject = this.currentProject
+        .compose(Transformers.takeWhen(this.projectSwitcherClicked));
```
luoser: can do a static import of `takeWhen`

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardViewModel.java
```diff
@@ -72,21 +73,25 @@ public ViewModel(final @NonNull Environment environment) {
...
+      this.projectSwitcherProjectClickOutput = this.projectSwitcherClicked;
+
+      this.startProjectActivity = currentProject
         .compose(takeWhen(this.projectViewClicked))
         .map(p -> Pair.create(p, RefTag.dashboard()));
     }
 
     private final PublishSubject<Void> projectViewClicked = PublishSubject.create();
+    private final PublishSubject<Project> projectSwitcherClicked = PublishSubject.create();
+    private final BehaviorSubject<Project> currentProject = BehaviorSubject.create();
```
luoser: since `currentProject` is just a local variable used in the ViewModel constructor u can remove this and replace L76 with:

```java
final Observable<Project> currentProject = projects.map(ListUtils::first);
```

app/src/test/java/com/kickstarter/viewmodels/CreatorDashboardViewModelTest.java
```diff
@@ -58,9 +58,11 @@ public void testStartProjectActivity() {
   }
 
   @Test
-  public void testLatestProjectAndStats() {
+  public void testProjectAndStats() {
+    final Project projectOne = ProjectFactory.project();
+    final Project projectTwo = ProjectFactory.project();
```
luoser: maybe do a builder with these and just change something like the `id()` to `id(1)` and `id(2)` or something to make the validity of this test stronger (since `projectOne` and `projectTwo`) are technically identical 👯

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardViewModel.java
swoopej: yeah but we also need to call `onNext` from the [new input](https://github.com/kickstarter/android-oss/pull/156/files#diff-e77a405946e815e7f08dcae820a52984R112) so not sure that would work?  Not sure this logic is as clear as it could be though.  Like I'm making a round trip from new input -> new output -> new input -> update `currentProject`.

app/src/test/java/com/kickstarter/viewmodels/CreatorDashboardViewModelTest.java
swoopej: 👌 I think the `public static synchronized int` magic in the `IdFactory` does this for us but also fine with being more explicit.

app/src/main/java/com/kickstarter/viewmodels/CreatorDashboardViewModel.java
luoser: ahhhh gotcha, i missed that the view model input `this.viewModel.refreshProject(project);` is invoked in the `projectSwitcherProjectClickOutput` call. seems like we should have just one `projectAndStats` output that handles the switching logic in the view model, we can pair on it!

app/src/test/java/com/kickstarter/viewmodels/CreatorDashboardViewModelTest.java
luoser: ahhh yeah you're right! forgot about that nice lil helper, fine with either way, ty ty

PR #157, opened 2017-09-06
Fix survey project request bug

luoser:
## what
Previous `handleProjectUriRequest(...)` logic in the SurveyResponseActivity was duplicating `startProjectActivity` logic and completing the request handling prematurely (i.e. tap project header -> go back -> tap project header again -> nothing happens). 

The [change made here](https://github.com/kickstarter/native-secrets/pull/14) fixes the bug, in conjunction with returning `false` in `handleProjectUriRequest` so that we continue to handle project uris. Other changes include code cleanup.

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/SurveyResponseActivity.java
+++ b/app/src/main/java/com/kickstarter/ui/activities/SurveyResponseActivity.java
@@ -1,24 +1,19 @@
 package com.kickstarter.ui.activities;
 
-import android.content.Intent;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v7.app.AlertDialog;
-import android.util.Pair;
 import android.view.View;
 import android.webkit.WebView;
 
 import com.kickstarter.R;
 import com.kickstarter.libs.BaseActivity;
-import com.kickstarter.libs.RefTag;
 import com.kickstarter.libs.qualifiers.RequiresActivityViewModel;
 import com.kickstarter.libs.utils.AnimationUtils;
-import com.kickstarter.models.Project;
 import com.kickstarter.services.KSUri;
 import com.kickstarter.services.KSWebViewClient;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/ui/activities/SurveyResponseActivity.java
```diff
@@ -64,16 +59,11 @@ protected void onCreate(final @Nullable Bundle savedInstanceState) {
...
-    this.viewModel.outputs.startProjectActivity()
-      .compose(bindToLifecycle())
-      .compose(observeForUI())
-      .subscribe(this::startProjectActivity);
   }
 
   private boolean handleProjectUriRequest(final @NonNull Request request, final @NonNull WebView webView) {
     this.viewModel.inputs.projectUriRequest(request);
-    return true;
+    return false;
```
luoser: returning false here lets the request handler know to continue handling project urls

app/src/main/java/com/kickstarter/viewmodels/SurveyResponseViewModel.java
```diff
@@ -70,41 +56,22 @@ public ViewModel(final @NonNull Environment environment) {
...
-        .compose(bindToLifecycle())
-        .subscribe(p -> this.startProjectActivity.onNext(Pair.create(p, RefTag.survey())));
-
       this.goBack = this.okButtonClicked;
     }
 
     /**
-     * Parses a project request for project params.
-     */
-    private @NonNull String extractProjectParams(final @NonNull Request request) {
```
luoser: it's nice that we can remove this logic here, i missed at first that `startProjectActivity`, with project uri request handlers, is baked into our [webview client]!!

swoopej: love [webview clients]

luoser:
> a test for our new input

just checked and the `projectRequestUri` input just moved cause of alphabetizing. sorry for the noise!!

Issue #64, opened 2017-03-03
background in View is better than in RelativeLayout?


ronanhardiman:
`<RelativeLayout
    android:id="@+id/checkout_loading_indicator"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    android:layout_below="@id/checkout_toolbar">

    <View
      android:layout_width="match_parent"
      android:layout_height="match_parent"
      android:background="@color/white_alpha_60"/>

    <ProgressBar
      android:id="@+id/loading_indicator"
      android:layout_width="wrap_content"
      android:layout_height="wrap_content"
      android:layout_centerVertical="true"
      android:layout_centerHorizontal="true"/>

  </RelativeLayout>`

background in **View** is better than in **RelativeLayout**?
thank you.

luoser:
Can you point me to the layout file which contains this snippet? Thanks!

ronanhardiman:
https://github.com/kickstarter/android-oss/blob/8c655fa0ecad478250d530f5c3dad0bfa0d49aec/app/src/main/res/layout/help_layout.xml
![image](https://cloud.githubusercontent.com/assets/1968388/24182933/ff5494ac-0eff-11e7-9307-0042c24ef553.png)

why not this
![image](https://cloud.githubusercontent.com/assets/1968388/24182978/2c7d6cba-0f00-11e7-80e2-eb9adff319bc.png)

thank you

luoser:
sorry this has taken so long, but you are correct, setting this background in a separate `View` tag is redundant when it can be done in the enclosing `RelativeLayout`. feel free to submit a PR, otherwise I'll go ahead an make the changes

PR #158, opened 2017-09-08
Remove view from help-layout

jorgecasariego:
This task (#64) remove **<View..>** because it was used just for add a background. Now, this background is set in the enclosing RelativeLayout.


luoser:
thanks!

PR #159, opened 2017-09-13
[Dashboard] Creator messages

luoser:
## what
Infrastructure refactors to support creator messages, accessed via the creator dashboard.

Main structural changes:
* Adding member data (e.g. permissions, unread msgs count, unseen activity count) to project model to display unread messages counts
* Refactoring `MessageThreadsViewModel` to refresh threads properly when configured with a `Project`
* Fix the reply EditText's hint to properly display participant (vs. project creator when viewing a project thread as a non-creator)

## bonus 
* Modernized `ProjectViewModelTests`
* Refactored `BackingActivity` to accept a `Project` and a `User?` to discern whether a creator is viewing a backer's pledge info vs. a backer viewing their own pledge info

Messages logic is confusing af so happy to chat through this PR irl or in a squash to logic 'splain :)


app/src/main/java/com/kickstarter/services/ApiClient.java
```diff
@@ -248,11 +248,6 @@ public ApiClient(final @NonNull ApiService service, final @NonNull Gson gson) {
   }
 
   @Override
-  public @NonNull Observable<MessageThreadsEnvelope> fetchMessageThreads(final @NonNull Mailbox mailbox) {
```
luoser: this was a weird temp method i made rather than just passing in a null project so we dont need it anymore / never really needed it oh well

app/src/main/java/com/kickstarter/viewmodels/MessageThreadsViewModel.java
```diff
@@ -151,14 +187,14 @@ public ViewModel(final @NonNull Environment environment) {
...
+    private final BehaviorSubject<Boolean> hasNoUnreadMessages = BehaviorSubject.create();
     private final BehaviorSubject<Boolean> isFetchingMessageThreads = BehaviorSubject.create();
     private final BehaviorSubject<List<MessageThread>> messageThreadList = BehaviorSubject.create();
-    private final Observable<Integer> unreadCountTextViewColorInt;
-    private final Observable<Integer> unreadCountTextViewTypefaceInt;
+    private final BehaviorSubject<Integer> unreadCountTextViewColorInt = BehaviorSubject.create();
+    private final BehaviorSubject<Integer> unreadCountTextViewTypefaceInt = BehaviorSubject.create();
     private final Observable<Boolean> unreadCountToolbarTextViewIsGone;
-    private final Observable<Integer> unreadMessagesCount;
+    private final BehaviorSubject<Integer> unreadMessagesCount = BehaviorSubject.create();
```
luoser: these all became `BehaviorSubject`s cause our `unreadMessagesCount` now also relies on a `project` from the `intent()`

app/src/test/java/com/kickstarter/viewmodels/ProjectViewModelTest.java
```diff
@@ -1,91 +1,127 @@
...
 import com.kickstarter.libs.MockCurrentUser;
 import com.kickstarter.models.Project;
+import com.kickstarter.models.User;
 import com.kickstarter.ui.IntentKey;
 
 import org.junit.Test;
 
 import rx.observers.TestSubscriber;
 
 public class ProjectViewModelTest extends KSRobolectricTestCase {
```
luoser: most noise here is from modernizing, few tests added tho!

luoser:
ahh sorry for the failing checks, forgot to refactor the BackingVMTests to take in a backer as well, fixing 😅

app/src/main/java/com/kickstarter/viewmodels/MessageThreadsViewModel.java
swoopej: cool.

app/src/main/java/com/kickstarter/viewmodels/MessagesViewModel.java
```diff
@@ -472,8 +485,8 @@ public ViewModel(final @NonNull Environment environment) {
     @Override public @NonNull Observable<List<Message>> messageList() {
       return this.messageList;
     }
-    @Override public @NonNull Observable<String> participantNameTextViewText() {
-      return this.participantNameTextViewText;
+    @Override public @NonNull Observable<String> creatorNameTextViewText() {
+      return this.creatorNameTextViewText;
```
swoopej: we don't need a version of this if the participant is not a creator? assume this was intentional but just checking because I can't remember what this will do in the UI.

luoser: oooh this is actually a good call / edge case I don't think I tested, thanks for catching will audit

luoser: ah--so this creator name text view text field is displayed under the project name up top, e.g.

image

so since a message thread will always have a project as a context, the creator name displayed there will always be the creator's name regardless of participants

PR #166, opened 2017-10-06
signup refactor

eoji:
## what
Refactored SignupViewModel to use new pattern.
Testing all outputs.


app/src/main/java/com/kickstarter/viewmodels/SignupViewModel.java
```diff
@@ -13,153 +13,188 @@
...
-      .doOnSubscribe(() -> this.formSubmitting.onNext(true))
-      .doAfterTerminate(() -> this.formSubmitting.onNext(false));
+  interface Outputs {
+    /** Emits a string to display when signup fails. */
+    Observable<String> errorString();
+
+    /** Emits a boolean that determines if the sign up button is enabled. */
+    Observable<Boolean> formIsValid();
+
+    /** Emits a boolean that determines if the sign up button is disabled. */
```
christopherwright: Does this comment line up with the name of the output? (sign up button disabled vs form submitting)

eoji: I had a question about output javadoc. Should they be what's happening in the Activity or what the ViewModel is doing?

luoser: chatted irl, and for posterity: in general an output should be doing the same thing in the ViewModel and the Activity so the javadoc would be representative of both

in this situation since we use both `formIsValid` and `formSubmitting` to set the enabled / disabled state of the signup button, [with funny lil helper methods](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/ui/activities/SignupActivity.java#L114-L120), we should ideally refactor to merge these outputs into one, e.g.

```java
Observable<Boolean> signupButtonIsEnabled();
```

and change the logic a bit in the VM, but this is a non-blocking change for purposes of this PR

PR #172, opened 2017-10-30
Refactor all Update activity helpers

luoser:
## what
In #167 we refactored the activity feed to patch its Update logic to use the appropriate `UpdateActivity.class` with `Project` and `Update` extras rather than the legacy `WebViewActivity.class` with a `Url`. 

Well...turns out there were a couple more spots we needed to change: the push notification, and the activity sample from the discovery feed. This PR:
* Refactors PushNotifications and DiscoveryFragment to open the appropriate `UpdateActivity` with proper data
* Refactors `UpdateViewModel` to accept either a Project or its Param 
* Renames an output to align with modern styles

A takeaway from this is to audit more thoroughly when implementing new features that replace "temporary" placeholder features.


luoser:
@swoopej good call thanks for helping me keep our Testing Integrity 💪

PR #173, opened 2017-11-02
Minimize dependency injections

luoser:
## what
Went through some android / architecture stuff w/ dear @christopherwright, was inspired by the eaze of simplifying our dependency graph, had some fidget time on the airplane so here is a PR cleaning our dependency injections via:
* Adding `environment()` to base UI classes when possible, so that components such as `Toolbar`s have access to an `environment` much like an activity or view holder
* Adding once-injected elements (e.g. `WebClient`) to our Environment and making sure there's a `@Provides` in our `ApplicationModule`
* Removing unused injections from `ApplicationGraph` for clarity and a Cleaner Codebase™

### DIFF
--- a/app/src/main/java/com/kickstarter/ApplicationGraph.java
+++ b/app/src/main/java/com/kickstarter/ApplicationGraph.java
@@ -1,8 +1,6 @@
 package com.kickstarter;
 
 import com.kickstarter.libs.Environment;
-import com.kickstarter.libs.KSCurrency;
-import com.kickstarter.libs.Koala;
 import com.kickstarter.libs.KoalaTrackingClient;
 import com.kickstarter.libs.utils.ApplicationLifecycleUtil;
 import com.kickstarter.services.gcm.MessageService;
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/kickstarter/ApplicationGraph.java
```diff
@@ -18,15 +13,10 @@
   Environment environment();
   void inject(ApplicationLifecycleUtil __);
```
luoser: the things that remain in `ApplicationGraph` will probably be here to stay so long as we use Dagger (until the end of time) since they are base classes which cannot have an environment built in

app/src/main/java/com/kickstarter/services/gcm/RegisterService.java
```diff
@@ -35,7 +36,7 @@ public void onCreate() {
   }
 
   @Override
-  protected void onHandleIntent(final @NonNull Intent intent) {
+  protected void onHandleIntent(final @Nullable Intent intent) {
```
luoser: the inherent overriden method has a `@Nullable` intent so made this change as per lint warning

PR #180, opened 2017-11-21
[Modernize] ResetPassword ViewModel

miga9:
This is a PR for modernizing ResetPasswordViewModel according to your latest VM architecture doc.


miga9:
Please look at #184 instead.

PR #183, opened 2017-11-25
Modernize ProjectNotificationSettings ViewModel

miga9:
This is a PR for modernizing ProjectNotificationSettingsViewModel according to your latest VM architecture doc.


app/src/main/java/com/kickstarter/viewmodels/ProjectNotificationSettingsViewModel.java
```diff
@@ -8,39 +8,47 @@
...
-    super(environment);
-
-    final ApiClientType client = environment.apiClient();
+  interface Outputs {
+    Observable<List<ProjectNotification>> projectNotifications();
+  }
 
-    this.projectNotifications = client.fetchProjectNotifications()
-      .compose(Transformers.pipeErrorsTo(this.unableToFetchProjectNotificationsError));
+  interface Errors {
```
luoser: aha, so we're also moving away from having `Errors` as a separate interface, you can move this Observable into `Outputs` since this error is technically just another output for the UI!

miga9: Oh I see 👍 I will make that change! :)

PR #186, opened 2017-12-03
Upgrade tools while building with latest Android Studio

tuyen-vuduc:
- Gradle v3.0.1
- ButterKnife v8.8.1
- BuildTools v26.0.2

Remove Retrolambda


eoji:
Thanks for this! We did it internally here: #267

PR #181, opened 2017-11-23
facepile ( •_•) ( •_•)>⌐■-■ (⌐■_■)

eoji:
## what
Implementing ✨ facepile ✨ designs showing up to 3 friends' names and avatars.
`SocialUtils` got hairy. Open to any and all suggestions.

PR #187, opened 2017-12-05
[Modernize] ProjectViewHolder

luoser:
## what
Updating our ProjectViewHolder by:
* making it a ViewModel
* writing tests for the ViewModel

Started doing this after brand refresh, upon realizing that yet another crucial part of our app (the project page) didn't have any test coverage. Now it does, and this will streamline future development.

NB: there are still a few helper functions (landscape and a11y related) that I haven't quite figured out the best solution for, so open to suggestions / doing that in another pass since this PR is already too big (most of it is javadoc / boilerplate tho!).


app/src/main/java/com/kickstarter/ui/viewholders/ProjectViewHolder.java
```diff
@@ -143,82 +132,333 @@
...
+      .subscribe(this::setGoalTextView);
+
+    this.viewModel.outputs.locationTextViewText()
+      .compose(bindToLifecycle())
+      .compose(observeForUI())
+      .subscribe(this.locationTextView::setText);
+
+    this.viewModel.outputs.projectOutput()
+      .subscribe(p -> {
+        // todo: break down these helpers
```
luoser: this is what i meant re: not being sure how to handle landscape and a11y specific outputs where it's kinda inevitable to just pass along the same project output to...

app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java
```diff
@@ -0,0 +1,497 @@
...
+      this.potdViewGroupIsGone = projectMetadata
+        .map(ProjectUtils.Metadata.POTD::equals)
+        .map(BooleanUtils::negate);
+
+      this.projectDisclaimerGoalReachedDateTime = project
+        .filter(Project::isFunded)
+        .map(Project::deadline);
+
+      this.projectDisclaimerGoalNotReachedString = project
+        .filter(p -> p.deadline() != null && p.isLive() && !p.isFunded())
```
swoopej: what's the case where a deadline would be null?

app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java
```diff
@@ -0,0 +1,497 @@
...
+
+      this.projectMetadataViewGroupBackgroundDrawableInt = projectMetadata
+        .filter(ProjectUtils.Metadata.BACKING::equals)
+        .map(__ -> R.drawable.rect_green_grey_stroke);
+
+      this.projectMetadataViewGroupIsGone = projectMetadata
+        .map(m -> m != ProjectUtils.Metadata.POTD
+          && m != ProjectUtils.Metadata.CATEGORY_FEATURED
+          && m != ProjectUtils.Metadata.BACKING
+        );
```
swoopej: can we simplify this like we did in the `ProjectCardHolderViewModel`: 

https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.java#L118

that might be true for some other outputs in this file - it seems like a lot of them are shared.  idk if there's any way to consolidate that logic across the two view models or if it would be over abstraction...

app/src/main/java/com/kickstarter/ui/viewholders/ProjectViewHolder.java
luoser: yeah agreed that's a little cleaner than invoking a chain of helpers

app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java
luoser: since the deadline comes from the `_mini` and not the `_compact`, serializer, it's a nullable field. also for projects that are in draft mode (fwiw i think we exclude projects in draft or "submitted" state from the api, but this also makes `deadline` technically nullable) :/

app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java
luoser: oooo i think that line is now pointed at a diff place cause #181 merged, but agreed that there is a bunch of repeated project logic here as in the project card :/

PR #190, opened 2017-12-08
[Modernize] 2FA

eoji:
## what
Modernizing the `TwoFactorViewModel` i.e. using interfaces, removing I/O class files, consolidating `Errors` with `Outputs`, etc.

Also fixed 2 broken tests.


app/src/test/java/com/kickstarter/viewmodels/TwoFactorViewModelTest.java
```diff
@@ -27,21 +34,20 @@ public void testTwoFactorViewModel_FormValidation() {
...
+    this.vm.inputs.code("444444");
+    this.formIsValid.assertValue(true);
 
-    vm.inputs.code("");
-    formIsValid.assertValues(true, false);
+    this.vm.inputs.code("");
+    this.formIsValid.assertValues(true, false);
 
-    koalaTest.assertValue("Two-factor Authentication Confirm View");
+    this.koalaTest.assertValue("Two-factor Authentication Confirm View");
```
luoser: we've started adding these event strings to [KoalaEvent.java](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/libs/KoalaEvent.java) to avoid re-typing / pasting the same string over and over again if you'd like to add these now!

app/src/main/java/com/kickstarter/libs/KoalaEvent.java
```diff
@@ -30,6 +32,8 @@ private KoalaEvent() {}
   public static final String SENT_MESSAGE = "Sent Message";
   public static final String SHOWED_SHARE_SHEET = "Showed Share Sheet";
   public static final String STARRED_PROJECT = "Starred Project";
+  public static final String TWO_FACTOR_AUTH_CONFIRM_VIEW = "Two-factor Authentication Confirm View";
+  public static final String TWO_FACTOR_AUTH_RESEND_CODE = "Two-factor Authentication Resend Code";
```
luoser: sweet!! forgot to mention that you can now also use these static var events in Koala.java to for the actual tracking event too :neckbeard:

selmanon:
`TwoFactorAuthenticationViewModel` is better that `TwoFactorViewModel` I guss

PR #192, opened 2017-12-11
Unique id tracking

swoopej:
just copyin this jazz over to android for the most part: https://github.com/kickstarter/ios-oss/pull/82/files

minus the test cases but I don't see anywhere that we have that in android for environment things? correct if I'm wrong though pls.

https://trello.com/c/5i1ntLPO/26-distinctids-shouldnt-reset-on-rails-event-in-android

### DIFF
--- a/app/src/main/java/com/kickstarter/KSApplication.java
+++ b/app/src/main/java/com/kickstarter/KSApplication.java
@@ -20,6 +20,8 @@ import net.danlew.android.joda.JodaTimeAndroid;
 import net.hockeyapp.android.CrashManager;
 import net.hockeyapp.android.CrashManagerListener;
 
+import org.joda.time.DateTime;
+
 import java.net.CookieHandler;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/kickstarter/KSApplication.java
```diff
@@ -60,6 +66,16 @@ public void onCreate() {
       .build();
     component().inject(this);
 
+    /*
+      setting cookie for unique id tracking, hardcoding prod
+     */
+    final HttpCookie cookie = new HttpCookie("vis", UUID.randomUUID().toString());
+    cookie.setMaxAge(DateTime.now().plusYears(100).getMillis());
```
luoser: geologically speaking, is 100 years into the future really that distant? :trollface:

swoopej: should we go into _the next millenium_

PR #193, opened 2017-12-11
[Modernize] Discovery

eoji:
## what
Modernizing the `DiscoveryViewModel` i.e. using interfaces, removing I/O class files, etc. ✨


app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.java
```diff
@@ -45,270 +44,315 @@
...
+    /** Emits a list of pages that should be cleared of all their content. */
+    Observable<List<Integer>> clearPages();
+
+    /** Emits when a newer build is available and an alert should be shown. */
+    Observable<InternalBuildEnvelope> showBuildCheckAlert();
+
+    Observable<Void> showLoginTout();
+    Observable<Void> showProfile();
+    Observable<Void> showSettings();
+    Observable<Void> showInternalTools();
```
luoser: can alpha and add a lil javadoc :)

PR #197, opened 2017-12-15
Create `isWebViewUri` helper to allow GraphQL reqs via Webviews

the-robear:
# What + Why 
This PR adds two helper methods, `isWebViewUri` and `isKSGraphQLUri` to determine whether the URI of a HTTP request is for a WebView i.e. a GET request to get the contents of a WebView. Here, we define a WebView URI as one that is a Kickstarter URI, but not to our GraphQL endpoint.

### DIFF
--- a/app/src/main/java/com/kickstarter/services/KSUri.java
+++ b/app/src/main/java/com/kickstarter/services/KSUri.java
@@ -42,6 +42,14 @@ public final class KSUri {
     return uri.getHost().equals(Uri.parse(webEndpoint).getHost());
   }
 
+  public static boolean isKSGraphQLUri(final @NonNull Uri uri, final @NonNull String webEndpoint) {
+    return isKickstarterUri(uri, webEndpoint) && uri.getPath().equals("/graph");
+  }
+
+  public static boolean isWebViewUri(final @NonNull Uri uri, final @NonNull String webEndpoint) {
+    return isKickstarterUri(uri, webEndpoint) && !isKSGraphQLUri(uri, webEndpoint);
+  }
+
   public static boolean isNewGuestCheckoutUri(final @NonNull Uri uri, final @NonNull String webEndpoint) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kickstarter/services/KSUri.java
```diff
@@ -42,6 +42,14 @@ public static boolean isKickstarterUri(final @NonNull Uri uri, final @NonNull St
     return uri.getHost().equals(Uri.parse(webEndpoint).getHost());
   }
 
+  public static boolean isKSGraphQLUri(final @NonNull Uri uri, final @NonNull String webEndpoint) {
+    return isKickstarterUri(uri, webEndpoint) && uri.getPath().equals("/graph");
```
eoji: Jw, do we need the first conditional here? Or do we not need it in `isWebViewUri`? 🤔

the-robear: Great question. I originally had this as `isGraphQLUri`, where it just checks if the path is `/graph`. But, if we were to use this method in isolation, it might not be exactly what they're thinking - e.g. we might try to use `isGraphQLUri` without checking if it is also a kickstarter uri.

Maybe I'm being a bit overprotective - I'm happy to change it back! And as I'm explaining it, I'm feeling that maybe I should... what do you think?

eoji: Overprotective is good! I was jw if I was missing some context.

app/src/test/java/com/kickstarter/services/KSUriTest.java
```diff
@@ -44,6 +44,29 @@ public void testKSUri_isKickstarterUri() {
     assertFalse(KSUri.isKickstarterUri(uri, this.webEndpoint));
   }
 
+  @Test
```
eoji: tests ✨

the-robear: I ❤️ TESTS

PR #199, opened 2017-12-21
Update circle config to CI2.0 and xlarge resource

talaris:
# What

- Update build syntax to Circle2.0
- Use `xlarge` resource class so we don't run out of memory
- Use a private docker image with oracle java 

# Why
Switching to CircleCI 2.0 allows us to use larger instances so we don't run into out of memory issues.

# Who
@eoji + @frewsxcv


talaris:
[Update "android-oss" CI config to 2.0 format, use "medium+" resource class.](https://trello.com/c/TJPvLtuw/1399-update-android-oss-ci-config-to-20-format-use-medium-resource-class)

.circleci/config.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-android-{{ .Branch }}
+            - v1-android-d
+
+      - run: $ANDROID_HOME/tools/android list target
+      - run: make bootstrap-circle
+      - run: echo y | android update sdk --no-ui --all --filter android-25
```
frewsxcv: > `echo y`

nice

talaris: yep its a fun trick, there is also `yes | some_command`, yes will repeatedly send `y` so its nice if you need to accept multiple licenses or similar.

talaris:
@eoji seems like the system is picky about the caches, so maybe something is getting included that shouldn't

eoji:
@talaris and i discussed trying this out as any issues usually resolve themselves when you rebuild without cache. if it becomes a consistent problem, we will revert.

PR #200, opened 2017-12-22
settings help bug

eoji:
# what
Using production endpoint from Secrets for displaying help links.

# before

# after

cc @lovehandle


eoji:
[P3: Help webviews not working](https://trello.com/c/L6qIIiAM/325-p3-help-webviews-not-working)

PR #202, opened 2018-01-05
Checkout rotation 404 bug

eoji:
# What
No longer intercepting the favicon which caused a 404 when rotating the `CheckoutActivity`.

# Proof in the pudding


eoji:
[Android checkout webview 404s when view changes from portrait to landscape](https://trello.com/c/pvPBq7XA/300-android-checkout-webview-404s-when-view-changes-from-portrait-to-landscape)

PR #203, opened 2018-01-09
Flipping converted pledge and goal for non US projects.

eoji:
# what
We had a bug where we flipped the converted goal and pledged amount of non US projects. 🙃 

# views
old v new


eoji:
[P2: Goal and pledged are swapped when currency converted on Android project page](https://trello.com/c/LFFCZjL1/383-p2-goal-and-pledged-are-swapped-when-currency-converted-on-android-project-page)

PR #211, opened 2018-02-08
creator tools collaborator access

eoji:
# what 
Adds `memberProjectsCount` to the `User` object so we show creator dashboard icon for collaborators, not just creators.
Fixes error when registering push token (can't call first() on an `Empty`).
Tests!


eoji:
[Creator tools: collaborator permissions](https://trello.com/c/FQ0cZwqw/391-creator-tools-collaborator-permissions)

PR #214, opened 2018-02-14
creator tools design sweep

eoji:
# what
Went through v1 of creator tools with @dannyalright:
-UI changes (spacing and text size)
-added a new style, `Title2Medium`

# before and after


eoji:
[Creator tools design sweep](https://trello.com/c/zGfaIjH9/417-creator-tools-design-sweep)

PR #220, opened 2018-03-06
Creator tools ref chart

eoji:
# what
Removing circle style chart and adding the ConstraintLayout support library to implement the new creator tools referral chart.

# see

PR #221, opened 2018-03-07
Creator tools ref color

eoji:
Please review #220 first, as this is branched off of it.

# what
Now that we have a new snazzy referrers chart, we need to display what the colors correspond to in our table.

# see


eoji:
[Creator dashboard: referrers table color-coding](https://trello.com/c/om9aKWVP/447-creator-dashboard-referrers-table-color-coding)

PR #224, opened 2018-03-13
refresh-user

eoji:
# what
Refreshing the user on app resumes.

# why
We are running into a bug where creators won't see the button to the dashboard because their user object is stale. `memberProjectsCount` on the `User` object was previously unused so for all users it's 0 until they are refreshed. Previously, we only refreshed when they logged in or out or they viewed settings, messages or profile. 

# how
This happens in the `onActivityResumed`(when the app is foregrounded) method of `ApplicationLifecycleUtil` where we are currently tracking app opens and refreshing the config. We have the access token (a nullable string) and the user observable. I make the token an observable by calling `.just`. Then, I use `zipWith` to combine them in a pair/tuple (You don’t want to subscribe to the current user observable because it would emit every time the user is refreshed and then you'd kick off never ending network requests to refresh the user ). Then I call refresh using the current user.

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
@@ -7,22 +7,28 @@ import android.content.res.Configuration;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.util.Pair;
 
 import com.facebook.appevents.AppEventsLogger;
 import com.kickstarter.KSApplication;
 import com.kickstarter.libs.CurrentConfigType;
+import com.kickstarter.libs.CurrentUserType;
 import com.kickstarter.libs.Koala;
 import com.kickstarter.libs.Logout;
 import com.kickstarter.libs.rx.transformers.Transformers;
+import com.kickstarter.models.User;
 import com.kickstarter.services.ApiClientType;
 import com.kickstarter.services.apiresponses.ErrorEnvelope;
 
 import javax.inject.Inject;
 
+import rx.Observable;
+
 public final class ApplicationLifecycleUtil implements Application.ActivityLifecycleCallbacks, ComponentCallbacks2 {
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
```diff
@@ -55,6 +61,16 @@ public void onActivityResumed(final @NonNull Activity activity) {
...
 
+      // Refresh the user
+      final Observable<User> user = this.currentUser.observable();
+
+      final String accessToken = this.currentUser.getAccessToken();
+
+      Observable.just(accessToken)
+        .filter(ObjectUtils::isNotNull)
+        .zipWith(user, Pair::create)
+        .subscribe(tokenUserPair -> this.currentUser.refresh(tokenUserPair.second));
```
justinswart: Hey as far as I can tell (but I might be missing something in the Java) is that `this.currentUser.refresh()` will just refresh with whatever is passed to it. So I think this will just take the `currentUser` and "refresh" it again. At which point do we make a new call to the API to fetch a fresh user?

Also, in this particular scenario it might not even be necessary to use RxJava because we're just imperatively lifting the token into FRP land in order to `subscribe` to it (with the user) and then to refresh the user with the user 🤔 

I think I might need a little more insight into how the lifecycle stuff works in Android but maybe we should pair on this 💪

eoji: 😅 look for my calendar invite!

justinswart: Aah got it! I missed the _fetch_ part of that 😄 

Then actually this is probably fine but couldn't it just be:

```
this.client.fetchCurrentUser()
  .filter(accessToken != null) //pseudeocode
  .subscribe(freshUser -> this.currentUser.refresh(freshUser))
```

Think that just removes the need to lift the accessToken into RxJava and to zip it with the new user.

PR #227, opened 2018-03-16
creator-tools-clipped-text

eoji:
# what
Checking if external stats are to the left of the breakdown view,  :sparkles: including padding :sparkles: so it doesn't get clipped

# before & after


eoji:
[P2: Android bar chart label cut off](https://trello.com/c/m9kDWeGx/510-p2-android-bar-chart-label-cut-off)

PR #229, opened 2018-03-23
hide ref chart :(

eoji:
# what
Hiding the referrer breakdown chart.
With tests ✨ 

# why
The `/v1/projects/{project_param}/stats` endpoint only returns so the first 30 stats so our chart is inaccurate and we need to hide it.

# before and after 😢


eoji:
[Hide chart section on Android dashboard](https://trello.com/c/Q1Pb0oJf/517-hide-chart-section-on-android-dashboard)

PR #234, opened 2018-03-30
marking rewards as received 

eoji:
# what 
Backers should be able to mark their rewards as received.
Cleaning up the layout.

# see

# jw
Am I writing too many tests?


eoji:
[Android: mark as received](https://trello.com/c/NQsTkVCE/423-android-mark-as-received)

app/src/main/java/com/kickstarter/viewmodels/BackingViewModel.java
```diff
@@ -275,6 +285,36 @@ public ViewModel(final @NonNull Environment environment) {
...
+        .map(ObjectUtils::isNotNull)
+        .compose(bindToLifecycle())
+        .subscribe(this.markAsReceivedIsChecked);
+
+      final Observable<Pair<Project, Backing>> projectAndBacking = Observable
+        .combineLatest(project, backing, Pair::create);
+
+      projectAndBacking
+        .compose(takePairWhen(this.markAsReceivedSwitchChecked))
+        .switchMap(triple -> this.client.postBacking(triple.first.first, triple.first.second, triple.second))
```
Scollaco: Is there a way to make this line more readable? I think that using `triple`, `first` and `second` is a little bit too generic to understand what's going on.

PR #235, opened 2018-04-02
[wip] backend funnel fix

eoji:
# what
We should be setting the visitor cookie on web and api. The value is `InstanceId.getInstance().getId()`. Read more about it here -> https://developer.android.com/training/articles/user-data-ids.html.

I also added our first instrumentation test ✨ but I'm doing some gross stuff with optionals so I'm open to feedback. (Cookies are returned as a list and not a map because 2 cookies can have the same name, party time.)

Shout out to @ryanpeterson for pairing on this with me for many moons.


eoji:
[Write instrumentation tests for Ruby fix and implement](https://trello.com/c/wkw38VnD/363-write-instrumentation-tests-for-ruby-fix-and-implement)

app/src/main/java/com/kickstarter/KSApplication.java
```diff
@@ -111,4 +103,18 @@ public boolean shouldAutoUploadCrashes() {
       }
     });
   }
+
+
+  private void setVisitorCookie() {
+    final String deviceId = InstanceID.getInstance(this).getId();
+    final String uuid = TextUtils.isEmpty(deviceId) ? UUID.randomUUID().toString() : deviceId;
```
eoji: rename to `uniqueIdentifier`

app/src/androidTest/java/VisitorCookieTest.java
```diff
@@ -0,0 +1,63 @@
...
+    Assert.assertNotNull(deviceId);
+
+    Assert.assertTrue(webCookie.getValue().equals(deviceId));
+    Assert.assertFalse(webCookie.getValue().equals("beep"));
+    Assert.assertTrue(apiCookie.getValue().equals(deviceId));
+    Assert.assertFalse(apiCookie.getValue().equals("boop"));
+  }
+
+  private Optional<HttpCookie> getOptionalVisitorCookieForURI(final @NonNull CookieStore cookieStore, final @NonNull URI webUri) {
+    // it's problematic, i know
```
eoji: it's fine

app/src/androidTest/java/VisitorCookieTest.java
```diff
@@ -0,0 +1,63 @@
...
+    final CookieManager cookieManager = environment.cookieManager();
+    final CookieStore cookieStore = cookieManager.getCookieStore();
+    final URI webUri = URI.create(Secrets.WebEndpoint.PRODUCTION);
+    final URI apiUri = URI.create(ApiEndpoint.PRODUCTION.url());
+    final Optional<HttpCookie> webVisCookie = getOptionalVisitorCookieForURI(cookieStore, webUri);
+    final Optional<HttpCookie> apiVisCookie = getOptionalVisitorCookieForURI(cookieStore, apiUri);
+    final HttpCookie webCookie = webVisCookie.get();
+    final HttpCookie apiCookie = apiVisCookie.get();
+
+    Assert.assertNotNull(webCookie);
```
eoji: instead of asserting they aren't null, assert that `cookie.isPresent()`

PR #236, opened 2018-04-04
name copy update

eoji:
# what
Updating sign up name copy and updating strings.

# before & after
 

# cc
@miki-nakata


eoji:
[Android app asks for user's full name on sign up page](https://trello.com/c/L1tklQfj/541-android-app-asks-for-users-full-name-on-sign-up-page)

PR #241, opened 2018-04-23
[bug] NPE in DiscoveryFragment

eoji:
# what
We have a few `NullPointerException`s because we attempt to access the `ViewModel` after the fragment is detached.

# how
Now we check if the `ViewModel` is `null`.

# bug
https://rink.hockeyapp.net/manage/apps/239014/app_versions/51/crash_reasons/215537957?type=overview


eoji:
[DiscoveryFragment crash](https://trello.com/c/2cCZtyQY/19-discoveryfragment-crash)

PR #242, opened 2018-04-24
[GDPR] Sign up newsletter opt out

eoji:
# what
Opting out of the newsletter on sign up and Facebook confirmation screens.
Updating tests.

# why
GIF

### DIFF
--- a/app/src/test/java/com/kickstarter/viewmodels/FacebookConfimationViewModelTest.java
+++ b/app/src/test/java/com/kickstarter/viewmodels/FacebookConfimationViewModelTest.java
@@ -91,27 +91,9 @@ public class FacebookConfimationViewModelTest extends KSRobolectricTestCase {
   }
 
   @Test
-  public void testToggleSendNewsLetter_ForUSUser() {
+  public void testToggleSendNewsLetter_isNotChecked() {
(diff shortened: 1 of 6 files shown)


eoji:
[Opting out of newsletter on sign up](https://trello.com/c/zCXy70vf/578-opting-out-of-newsletter-on-sign-up)

app/src/test/java/com/kickstarter/viewmodels/FacebookConfimationViewModelTest.java
```diff
@@ -98,12 +98,12 @@ public void testToggleSendNewsLetter_ForUSUser() {
     this.vm = new FacebookConfirmationViewModel.ViewModel(environment);
 
     this.vm.outputs.sendNewslettersIsChecked().subscribe(this.sendNewslettersIsChecked);
-    this.sendNewslettersIsChecked.assertValue(true);
+    this.sendNewslettersIsChecked.assertValue(false);
```
Scollaco: I would change the name of the test to something like `testToggleSendNewsLetter_isNotChecked`, since now we don't differenciate the toggle status for US/Non-US users.

PR #244, opened 2018-05-03
[GDPR] Hiding recommendations

eoji:
# what
Hiding recommendations when user has setting off.

# why
GIF

# 👀 wax on, wax off


eoji:
[Hide Recommendations when settings are turned off](https://trello.com/c/5lu9XPEC/603-hide-recommendations-when-settings-are-turned-off)

PR #253, opened 2018-05-21
[GDPR] messages opt out

eoji:
# what 
opting out of receiving message push notifications

# why
GIF

# how
We have 2 new properties on the `User` model, `notifyMobileOfMessages` and `notifyOfMessages`.

# see


eoji:
[Implement Message Settings Android](https://trello.com/c/szccAdGg/692-implement-message-settings-android)

PR #255, opened 2018-05-23
Recommendations info dialog

eoji:
# what 
Showing recommendations info, with tests (•̀ᴗ•́)و ̑̑

# how
Updated strings.
Added the new info icon.
On user click, we show the alert.
Test ✨

# why
GIF

# c-c-c-check it out

cc @dannyalright


dannyalright:
minor thing: I'd expect the _Got it._ to be all caps on Android.

1. would you, too?
2. is this consistent with how we label dialog buttons elsewhere on the app?

eoji:
@dannyalright 
1. I actually would expect it to be in caps, per https://material.io/design/components/dialogs.html#alert-dialog
2. Yes, our other dialog buttons have sentence case 🤔

dannyalright:
cool. I'd prefer it to be _GOT IT_, without the period at the end. Should we do it here, one-off, and I'll make a card to fix others another time? Or just keep as is and do all at another time?

eoji:
@dannyalright Let's do it now and I'll open a PR with that new string!

eoji:
Here's the strings PR: https://github.com/kickstarter/kickstarter/pull/11028

eoji:
New string 😎 @dannyalright

dannyalright:
awesome, thank you!

eoji:
[Recommendations and Privacy UI in Settings](https://trello.com/c/qr4OXlu1/606-recommendations-and-privacy-ui-in-settings)

PR #256, opened 2018-05-25
v1 of data export

eoji:
# what
Linking externally to data export feature.

# why
GDPR is still coming for my neck

# see


eoji:
[Export Data external link - Android](https://trello.com/c/zOn1M6SY/713-export-data-external-link-android)

PR #259, opened 2018-06-07
Toggling following in settings

eoji:
# what
Users should be able to toggle whether or not they can be following or `social` via settings. This change also updates whether the `Backed by people you follow` section is visible in the drawer.
This is a destructive change so we need to confirm they definitely want to do this.

# why
G
D
P
R

# how 🤔 
The `optIntoFollowing` input determines whether user has checked the switch on or off. The `optOutOfFollowing` input determines whether user has confirmed turning Following off.

When `optIntoFollowing` is `true`, we just turn Following on.
When `optIntoFollowing` is `false`, we show the confirmation dialog. This dialog is not cancelable, I ~~expect~~ demand, they make a choice!

When `optOutOfFollowing` is `true`, that's a confirmation, we turn Following off.
When `optOutOfFollowing` is `false`, we change the switch back to on!

# see 👀 
First, I'm `social` and my drawer shows  `Backed by people you follow`. I attempt to turn it off, cancel. I attempt to turn it off again, and confirm. My drawer no longer shows  `Backed by people you follow`.

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/SettingsViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/SettingsViewModel.java
@@ -64,7 +64,10 @@ public interface SettingsViewModel {
     void notifyOfUpdates(boolean checked);
 
     /** Call when the user toggles the Following switch. */
-    void optIntoFollowing();
+    void optIntoFollowing(boolean checked);
+
+    /** Call when the user confirms or cancels opting out of Following. */
+    void optOutOfFollowing(boolean optOut);
 
     /** Call when the user toggles the Recommendations switch. */
     void optedOutOfRecommendations(boolean checked);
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/viewmodels/SettingsViewModel.java
```diff
@@ -161,6 +179,26 @@ public ViewModel(final @NonNull Environment environment) {
...
+        .subscribe(__ -> this.showConfirmFollowingOptOutPrompt.onNext(null));
+
+      this.optOutOfFollowing
+        .compose(bindToLifecycle())
+        .filter(optOut -> optOut)
+        .subscribe(__ -> this.userInput.onNext(this.userOutput.getValue().toBuilder().social(false).build()));
+
+      this.optOutOfFollowing
+        .compose(bindToLifecycle())
+        .filter(optOut -> !optOut)
```
ifbarrera: I'm not sure if it might make sense to consolidate these two subscribers into one:

```
.subscribe(optOut -> 
   if (optOut) {
     this.userInput.onNext(this.userOutput.getValue().toBuilder().social(false).build())
   } else {
     this.hideConfirmFollowingOptOutPrompt.onNext(null)
   })
```

eoji: I'm glad you called this out because I had it that way and changed it. What do you think is best? I was trying to avoid doing an `if/else` because it's not very functional 😭but then this way is not very readable.

eoji:
[Following consent UI](https://trello.com/c/hzdseeEx/568-following-consent-ui)

eoji:
[Following Consent Implement Android](https://trello.com/c/TOhaa2MN/682-following-consent-implement-android)

PR #260, opened 2018-06-07
fixing broken videos and images on 19 and below

eoji:
# what, why, how
TLS 1.0 and 1.1 is no longer supported on our image host starting 5/18/18.  https://blog.imgix.com/2018/05/18/tls-deprecation.html

Android versions 19 and below don't have TLS 1.2 enabled by default :neutral_face: So videos and images are currently broken for all users on 19 and below 😢 

This change uses Google Play Services to patch Android's security provider, it's lit 👏 
Shout out to this StackOverflow answer: https://stackoverflow.com/questions/29916962/javax-net-ssl-sslhandshakeexception-javax-net-ssl-sslprotocolexception-ssl-han/36892715#36892715

I'm removing the work Lisa did to enable TLS on 19 > devices because it was only on Retrofit, we have external dependencies that make network calls including Picasso (I think uses Okhttp) and ExoPlayer. This needs to be a global change.

# before & after

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,9 +38,4 @@ secrets:
 	mkdir -p app/src/main/assets/www/
 	cp vendor/native-secrets/android/WebViewJavascript.html app/src/main/assets/www/WebViewJavascript.html || true
 
-	# Copy TLS helpers over.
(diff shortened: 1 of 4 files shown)


eoji:
[[ZD] [Android] App not loading pictures or videos when browsing projects](https://trello.com/c/fDPOwW0b/4524-zd-android-app-not-loading-pictures-or-videos-when-browsing-projects)

PR #269, opened 2018-07-03
Enhancement - Setup new settings screen layout

Rcureton:
## Why
* We're upgrading our overloaded settings and moving into a new place 😏

## What
* Added `settings_layout_new` to implement the new designs from Danny.

## See
screen shot 2018-07-05 at 1 32 10 pm


Rcureton:
[Re-factored notification settings](https://trello.com/c/40eteppL/843-re-factored-notification-settings)

PR #272, opened 2018-07-07
Saved empty view

eoji:
# what
When users don't have any saved projects, show an empty view with a heart animation. They can also click the heart to see the animation again.

# how
Added `fadeOutAndScale`, `fadeInAndScale`, and `crossFade` to `AnimationUtils`.
Removed some unused drawables.
Updated `DiscoveryFragmentViewModel` to new style with `Inputs` and `Outputs` in the same file.
Updated `DiscoveryFragmentViewModelTest` to have `setUpEnvironment` helper method.

# see


app/src/main/java/com/kickstarter/libs/utils/AnimationUtils.kt
```diff
@@ -20,4 +26,52 @@ object AnimationUtils {
     animation.fillAfter = true
     return animation
   }
+
+  @JvmOverloads
+  fun fadeInAndScale(view: View, duration: Long = 500L, startDelay: Long = 500L,
```
eoji: Open to renaming suggestions.

app/src/main/java/com/kickstarter/libs/utils/AnimationUtils.kt
```diff
@@ -20,4 +26,52 @@ object AnimationUtils {
...
+                     interpolator: Interpolator = AccelerateDecelerateInterpolator()): ObjectAnimator {
+    val scaleX = PropertyValuesHolder.ofFloat("scaleX", 0f, 1f)
```
eoji: Should I make these strings constants?

app/src/main/java/com/kickstarter/libs/utils/AnimationUtils.kt
```diff
@@ -20,4 +26,52 @@ object AnimationUtils {
...
+    return animator
+  }
+
+  @JvmOverloads
+  fun crossFade(visibleView: View, hiddenView: View, crossDuration: Long = 500L, startDelay: Long = 500L,
+                      interpolator: Interpolator = AccelerateDecelerateInterpolator()): AnimatorSet {
+    val crossFadeAnimatorSet = AnimatorSet()
+    val fadeOutAndScale = fadeOutAndScale(visibleView, crossDuration, startDelay, interpolator)
+    val fadeInAndScale = fadeInAndScale(hiddenView, crossDuration, startDelay, interpolator)
+
```
eoji: What's our new line policy?

app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
```diff
@@ -44,214 +44,295 @@
...
+          this.projectList.onNext(Collections.emptyList());
+        });
+
+      this.paramsFromActivity
+        .compose(combineLatestPair(userIsLoggedIn))
+        .map(pu -> isOnboardingVisible(pu.first, pu.second))
+        .compose(bindToLifecycle())
+        .subscribe(this.shouldShowOnboardingView);
+
+      this.paramsFromActivity
```
eoji: Lines 169-181 is what I actually changed.

eoji:
[Update null state for no saved projects (v2)](https://trello.com/c/NJ7p8Hh0/560-update-null-state-for-no-saved-projects-v2)

app/src/main/java/com/kickstarter/libs/utils/AnimationUtils.kt
Rcureton: Do we see this being used elsewhere? If so then the name is fine but if it's just going to be used for the heartbeat animation I would name it something related to it.

app/src/main/java/com/kickstarter/libs/utils/AnimationUtils.kt
Rcureton: Yes seeing that we're using them in many places. Good call 😉

app/src/main/java/com/kickstarter/libs/utils/AnimationUtils.kt
Rcureton: For me, as long as it doesn't run the length of the screen it's good with me. What are your thoughts?

Rcureton:
Gif

PR #273, opened 2018-07-09
[Enhancement] - Added new activities to new SettingsActivity

Rcureton:
## What
- Created and wired up all the click listeners to the new activities in the SettingsNewViewModel. 

## Why
- To finish the Settings screen so that we can implement the other features.

## See
jul-09-2018 14-59-49

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -10,10 +10,10 @@ def internalPublicVersion = new File(project.rootDir.path + "/app/internal_versi
 def isCircle = "true" == System.getenv("CIRCLECI")
 
 ext.signing = [
-    storeFilePath: "path/to/keystore",
-    storePassword: "keystore password",
-    keyAlias     : "key alias",
-    keyPassword  : "key password"
+        storeFilePath: "path/to/keystore",
+        storePassword: "keystore password",
+        keyAlias     : "key alias",
+        keyPassword  : "key password"
 ]
 
 if (file('signing.gradle').exists()) {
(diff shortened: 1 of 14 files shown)


app/build.gradle
```diff
@@ -249,3 +249,5 @@ gradle.taskGraph.beforeTask { Task task ->
 }
 
 apply plugin: 'com.google.gms.google-services'
+
+apply plugin: 'kotlin-android-extensions'
```
eoji: Does this belong down here? I'm worried about any weird errors.

app/src/main/java/com/kickstarter/viewmodels/SettingsNewViewModel.kt
```diff
@@ -1,16 +1,33 @@
...
+import com.kickstarter.ui.activities.NewsletterActivity
+import com.kickstarter.ui.activities.NotificationsActivity
+import com.kickstarter.ui.activities.PrivacyActivity
 
-class SettingsNewViewModel(private val context: Context) : BaseObservable() {
+class SettingsNewViewModel(private val context: Context, private val environment: Environment) : BaseObservable() {
 
     fun getVersion() = BuildConfig.VERSION_NAME
 
-    fun rateUsClick() {
```
eoji: Let's....alphabetize
( • _•)
 these 
( • _•)>⌐■-■
functions 
(⌐■_■)

app/build.gradle
Rcureton: It randomly generated there somehow but I'll move it!

PR #288, opened 2018-08-03
Creator Notifications v1?  👀

eoji:
# what
There are settings for creator notifications that never lived in the Android app because we don't send Android creator pushes (but we do send emails).
Those settings are:
`backings`, `comments`, `creator_edu`, and `post_likes`
I also fixed the row animation getting cut off (cc @dannyalright)

# how
I refactored how we update the buttons because it was hard to keep track of 8.
I gave the Notifications rows their own style.
Testzzzzz.

# see still & gif


app/src/main/java/com/kickstarter/viewmodels/NotificationsViewModel.kt
```diff
@@ -126,19 +188,21 @@ interface NotificationsViewModel {
...
-        override fun user(): Observable<User> {
-            return this.userOutput
-        }
+        override fun creatorNotificationsAreGone(): Observable<Boolean> = this.creatorNotificationsAreGone
 
-        override fun unableToSavePreferenceError(): Observable<String> {
-            return this.unableToSavePreferenceError
+        override fun user(): Observable<User> = this.userOutput
+
+        override fun unableToSavePreferenceError(): Observable<String> =  this.unableToSavePreferenceError
```
Rcureton: Nice use of type inference!

app/src/main/res/values/styles.xml
```diff
@@ -528,4 +527,15 @@
     <item name="android:layout_marginEnd">@dimen/grid_1</item>
   </style>
 
+  <style name="SettingsNotificationRow">
+    <item name="android:paddingStart">@dimen/activity_horizontal_margin</item>
```
Rcureton: Nice cleanup into a style!

eoji:
[Creator Notifications on Android](https://trello.com/c/gEsPClnT/873-creator-notifications-on-android)

PR #289, opened 2018-08-03
No longer consuming project preview links

eoji:
# what
No longer showing a blank project page when a user navigates to a project preview. Instead, taking the user to their external browser.

# how
Checking if the URL looks like a project but contains the param `token` 🦆 
WITH TESTS

# seeing is believing

cc @roblum @willmanduffy thanks for your help!


eoji:
[Project Preview links broken on Android](https://trello.com/c/MobE5MZ8/902-project-preview-links-broken-on-android)

PR #296, opened 2018-08-09
[Refactor] - Refactored all classes associated with LoginActivity to Kotlin

Rcureton:
## What
- Refactored `LoginActivity`, `LoginViewModel` and `LoginViewModelTest` to Kotlin
- Added Edit Text extension function to retrieve the string from an EditText. Also documented what the methods do.

## Why
- Want to get our Kotlin percentage up in our codebase!


app/src/main/java/com/kickstarter/extensions/EditTextExt.kt
```diff
@@ -15,3 +16,6 @@ fun EditText.onChange(input: (String) -> Unit) {
         override fun onTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) {}
     })
 }
+
+/** This is an Extension function to get the string input for EditText */
+fun EditText.string() = this.text.toString()
```
eoji: I was wondering if we should call this `text` vs `string`. I first thought it was a shortcut for `toString` and I was like 👁 why would we want that value?

app/src/main/java/com/kickstarter/extensions/EditTextExt.kt
```diff
@@ -4,6 +4,7 @@ import android.text.Editable
 import android.text.TextWatcher
 import android.widget.EditText
 
+/** This is an Extension function to implement a TextChangedListener for EditText */
```
eoji: This is ✨ SO COOL ✨ and very useful!

app/src/main/java/com/kickstarter/ui/activities/LoginActivity.java
```diff
@@ -1,181 +0,0 @@
-package com.kickstarter.ui.activities;
```
eoji: 👋 java

app/src/main/java/com/kickstarter/extensions/EditTextExt.kt
Rcureton: I think text would be good I'll make the change.

PR #297, opened 2018-08-09
[Bug Fix] -added check to only enable fabric when it's not OSS

Rcureton:
## What 
- Added check to ensure that the app won't enable Fabric when it's OSS.


Rcureton:
Added the checks there as well!

PR #302, opened 2018-08-13
added bottom slide up animation for settings

Rcureton:
## What
- added bottom slide up animation for settings
- Added extension function for the animation.

## Why
- Recommended by @dannyalright plus it looks better.

## See
aug-13-2018 15-50-53

### DIFF
--- a/app/src/main/java/com/kickstarter/extensions/ActivityExt.kt
+++ b/app/src/main/java/com/kickstarter/extensions/ActivityExt.kt
@@ -1,6 +1,0 @@
-package com.kickstarter.extensions
-
-import android.app.Activity
-import com.kickstarter.R
-
-fun Activity.bottomSlideAnimation() = overridePendingTransition(R.anim.settings_bottom_slide, R.anim.fade_out)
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/extensions/ActivityExt.kt
```diff
@@ -0,0 +1,6 @@
+package com.kickstarter.extensions
+
+import android.app.Activity
+import com.kickstarter.R
+
+fun Activity.bottomSlideAnimation() = overridePendingTransition(R.anim.settings_bottom_slide, R.anim.fade_out)
```
eoji: Let's get rid of this extension and use what already exists: `BaseActivity` has a method called `exitTransition` that you would override in `SettingsActivity`.

So instead of adding `bottomSlideAnimation` every time you start a new activity, you'd just add 

```  @Override
  protected @Nullable Pair<Integer, Integer> exitTransition() {
    return slideUpFromBottom();
  } 
```

where `slideUpFromBottom()` is a Pair of animation resource IDs defined in `TransitionUtils`:

```
  public static @NonNull Pair<Integer, Integer> slideUpFromBottom() {
    return Pair.create(R.anim. settings_bottom_slide, R.anim. fade_out);
  }
```

PR #303, opened 2018-08-14
Conditional creator notifications

eoji:
I copied most of this description from https://github.com/kickstarter/ios-oss/pull/401 @ifbarrera thank youuu

# What
When pledge activity email notifications are enabled, show an "Email frequency" spinner that allows a user to select how frequently they should receive notifications. The option presents a spinner with the two frequency options. Selecting an option should update the user's notification setting. Selecting the same option, does nothing.

If pledge activity email notifications are disabled, the "Email frequency" container should not show.

# Why
It's part of creator notifications that Android didn't have.

# see

cc @dannyalright You can't change the white space around the spinner 😞 This is the default style. But maybe we could get it in when we make the reusable component?


app/src/main/java/com/kickstarter/ui/activities/NotificationsActivity.kt
```diff
@@ -180,9 +193,31 @@ class NotificationsActivity : BaseActivity<NotificationsViewModel.ViewModel>() {
     private fun displayBackingsNotificationSettings(user: User) {
         this.notifyMobileOfBackings = isTrue(user.notifyMobileOfBackings())
         this.notifyOfBackings = isTrue(user.notifyOfBackings())
+        this.notifyOfCreatorDigest = isTrue(user.notifyOfCreatorDigest())
+
+        val frequencyIndex = when {
```
Rcureton: Nice use of `when`!  I see you 👀

app/src/main/java/com/kickstarter/viewmodels/NotificationsViewModel.kt
```diff
@@ -196,6 +218,8 @@ interface NotificationsViewModel {
             this.userInput.onNext(this.userOutput.value.toBuilder().notifyOfUpdates(checked).build())
         }
 
+        override fun creatorDigestFrequencyIsGone(): Observable<Boolean> = this.creatorDigestFrequencyIsGone
```
Rcureton: Little nit but nothing big you can just remove the return type since it's a `single expression function`

dannyalright:
looks good! No worries about white space. Good idea about reusable component in the future. 

Could we have the green underline remain until the user makes a selection? Or is what we have now 'we got for free'?

eoji:
@dannyalright Yeah, that's the style; it only turns green in the `pressed` state.

eoji:
[Conditional changes depending on whether the user (a creator) has chosen to receive Pledge activity by email or not](https://trello.com/c/69Lkz0b5/896-conditional-changes-depending-on-whether-the-user-a-creator-has-chosen-to-receive-pledge-activity-by-email-or-not)

PR #304, opened 2018-08-16
Updated strings.

eoji:


eoji:
[Export data not translated](https://trello.com/c/Fds8kaKK/919-export-data-not-translated)

PR #305, opened 2018-08-16
👋🏾 bye data binding

eoji:
# what
Removed data binding.
Renamed `HelpNewViewModel` to `HelpSettingsViewModel`.
Refactored `HelpSettingsViewModel` to extend our base `ActivityViewModel`.
Fixed broken tests.
Added tests for `HelpSettingsViewModel`.
Removed `How Kickstarter Works`.

image


app/src/main/java/com/kickstarter/ui/activities/HelpSettingsActivity.kt
```diff
@@ -0,0 +1,85 @@
...
+        privacy_policy.setOnClickListener {
+            startActivityWithSlideLeftTransition(Intent(this, HelpActivity.Privacy::class.java))
+        }
+
+        terms_of_use.setOnClickListener {
+            startActivityWithSlideLeftTransition(Intent(this, HelpActivity.Terms::class.java))
+        }
+    }
+
+    private fun composeContactEmail(user: User?) {
```
Rcureton: Since we know that `user` can't be null do we should remove the `?`. That way the code will be much cleaner.

app/src/main/java/com/kickstarter/ui/activities/SettingsActivity.kt
```diff
@@ -71,10 +79,8 @@ class SettingsActivity : BaseActivity<SettingsViewModel.ViewModel>() {
         ApplicationUtils.startNewDiscoveryActivity(this)
     }
 
-    override fun onBackPressed() {
-        super.onBackPressed()
-        val exit = slideInFromLeft()
-        overridePendingTransition(exit.first, exit.second)
+    override fun exitTransition(): Pair<Int, Int>? {
```
Rcureton: I think we can use a single expression function here 
```
override fun exitTransition() = slideInFromLeft()
```

eoji:
[Refactored Help settings Android](https://trello.com/c/Phs2zHyo/824-refactored-help-settings-android)

PR #318, opened 2018-08-28
!! Reset password crash

eoji:
# what 
First pass of fixing crash when users reset their password.
```
Fatal Exception: java.lang.RuntimeException: Unable to start activity 
ComponentInfo{com.kickstarter.kickstarter/com.kickstarter.ui.activities.LoginActivity}: 
rx.exceptions.OnErrorNotImplementedException: 
lateinit property confirmResetPasswordSuccessDialog has not been initialized
```

# why'd i make this WIP?
Because I am wondering if I should write a UI test because then we'd have caught this bug.

# 8/30
I wrote the tests!


eoji:
[[P2] Crash when resetting password](https://trello.com/c/Q9hL0kyJ/937-p2-crash-when-resetting-password)

PR #323, opened 2018-09-06
[Bug Fix] - Fixed TokenListener crash on devices 8 and higher

Rcureton:
## What
- We updated the `FirebaseInstanceIdService` to `FirebaseMessagingService`
- We added `Firebase-job-dispatcher` to the project.
- We Removed `IntentService` and replaced it with `JobService`.

## Why
- We added `Firebase-job-dispatcher` so that on devices on OS 8 and higher we can run jobs in the background. It also supports back to API level 14.

- We upgraded the `FirebaseMessaging` import to use the new `FirebaseMessagingService` class which replaced `FirebaseInstanceIdService`. We also had to change the `intent-filter` in the manifest. 
    > A service that extends FirebaseMessagingService. This is required if you want to do any message handling beyond receiving notifications on apps in the background. To receive notifications in foregrounded apps, to receive data payload, to send upstream messages, and so on, you must extend this service.


app/src/main/java/com/kickstarter/libs/DeviceRegistrar.java
```diff
@@ -29,7 +31,13 @@ public void registerDevice() {
       return;
     }
 
-    this.context.startService(new Intent(this.context, RegisterService.class));
+    final FirebaseJobDispatcher dispatcher = new FirebaseJobDispatcher(new GooglePlayDriver(this.context));
```
eoji: Looks like we're doing the same thing 3 times, which seems like a Util class would be helpful! Let's make a Util class for Jobs™. So you'd pass in a context, the class, and the tag. And then we could make these tags constants!

eoji: I'm also wondering if it's expense to keep making new `FirebaseJobDispatcher` objects and we should make it a singleton instead.

app/src/main/java/com/kickstarter/services/firebase/RegisterService.java
```diff
@@ -18,35 +17,32 @@
...
     super.onCreate();
     ((KSApplication) getApplicationContext()).component().inject(this);
   }
 
+
   @Override
-  protected void onHandleIntent(final @Nullable Intent intent) {
-    Timber.d("onHandleIntent");
+  public boolean onStartJob(final JobParameters job) {
+    FirebaseInstanceId.getInstance().getInstanceId().addOnSuccessListener(instanceIdResult -> {
```
eoji: Nice job getting rid of the try catch

app/src/main/java/com/kickstarter/services/firebase/RegisterService.java
```diff
@@ -18,35 +17,32 @@
...
-      if(ObjectUtils.isNotNull(token)) {
-        sendTokenToApi(token);
-        subscribeToGlobalTopic();
-      }
-    } catch (final Exception e) {
-      Timber.e("Failed to complete token refresh: %s", e);
-    }
+  @Override
+  public boolean onStopJob(final JobParameters job) {
+    return false;
```
eoji: what does this mean and why do we return false?

Rcureton: I guess if we needed to kill the job we would make sure we dispose of it here.

app/src/main/java/com/kickstarter/libs/DeviceRegistrar.java
Rcureton: I'll make a file-level function for it!

app/src/main/java/com/kickstarter/services/firebase/RegisterService.java
```diff
@@ -18,35 +17,32 @@
...
+      final String newToken = instanceIdResult.getToken();
+      Timber.d("newToken", newToken);
+      sendTokenToApi(newToken);
+      subscribeToGlobalTopic();
+    });
 
-    try {
-      final String token = FirebaseInstanceId.getInstance().getToken();
-      Timber.d("Token: %s", token);
+    return true;
```
eoji: >Return `true` from this method if your job needs to continue running.`

I don't think that's what we want. When do we ever stop the job?

app/src/main/java/com/kickstarter/ApplicationModule.java
```diff
@@ -177,6 +179,13 @@ static ApolloClientType provideApolloClientType(final @NonNull ApolloClient apol
     return new KSApolloClient(apolloClient);
   }
 
+  @Provides
```
eoji: Let's get rid of this.

Rcureton: I removed it in a previous PR

eoji:
[[P2] TokenListenerService service crash on Oreo and above](https://trello.com/c/hPwUQhEw/946-p2-tokenlistenerservice-service-crash-on-oreo-and-above)

PR #328, opened 2018-09-12
[Enhancement] - Removed privacy feature for creators

Rcureton:
## What
- Added logic to hide privacy rows when the user is a creator.
- Added ViewModelTest functions to test if the privacy row is hidden.

## Why
- Creators can't have private profiles.

## See
ksl
sep-12-2018 14-29-41-rashad


Rcureton:
[Hide Private profile from creators](https://trello.com/c/pYvjQ646/975-hide-private-profile-from-creators)

PR #339, opened 2018-09-19
[Enhancement] - Added dialog for currency change

Rcureton:
## What
- Added dialog for when the currency picker is changed.

## See 👀
sep-19-2018 15-40-09


eoji:
Let's close this and reopen one against master 👀

PR #344, opened 2018-10-02
Change password

eoji:
# what did i do?
TLDR; the most.

Updated strings.
Added mutation for changing password.
Added green check icon.
Deriving GraphQL endpoint from web endpoint so we don't have to manually change it for staging.
Changed staging toolbar to grey so we can green toolbar icons.
Also fixed it in XML previews.
Created `Activity` extensions for showing `Snackbar`s.
Created reusable `Snackbar` extensions for success and error messages.
The `Snackbar` extensions return themselves so you can chain them (if you wanted to inflate a success/error `Snackbar` manually).
Change password:
Showing warning when new password is less than 6 characters.
Showing warning when confirm password is not equal to new confirm.
Disabling save button until all validations pass.
Added 2 new `LoginReasons`: `CHANGE_PASSWORD` and `RESET_PASSWORD` because we want to prefill the email if it's supplied from the intent but need to check if we're showing the reset dialog or change password Snackbar.
Moved the text changed listeners to before we subscribe to the `prefillEmail` output.
When password is successfully changed, taking user to `LoginActivity`. 
LoginActivity shows success `Snackbar` letting the user know their changes were saved.
Made LoginActivity's root a `CoordinatorLayout` so the `Snackbar` is shown over the keyboard.
Tests on tests on tests.

image

# see
device-2018-10-02-183943 2018-10-02 18_43_09 device-2018-10-02-184119 2018-10-02 18_43_25

# what's left?
Final copy.


eoji:
[Change Password](https://trello.com/c/vDFqsAQc/882-change-password)

app/src/main/java/com/kickstarter/ui/activities/ChangePasswordActivity.kt
```diff
@@ -1,13 +1,116 @@
...
+import com.kickstarter.viewmodels.ChangePasswordViewModel
+import kotlinx.android.synthetic.main.activity_change_password.*
+import kotlinx.android.synthetic.main.change_password_toolbar.*
 
-class ChangePasswordActivity : AppCompatActivity() {
+
+@RequiresActivityViewModel(ChangePasswordViewModel.ViewModel::class)
+class ChangePasswordActivity : BaseActivity<ChangePasswordViewModel.ViewModel>() {
+
+    var saveEnabled = false
```
Rcureton: This can be private

app/src/main/java/com/kickstarter/viewmodels/ChangePasswordViewModel.kt
```diff
@@ -0,0 +1,176 @@
...
+                        && isNotEmptyAndAtLeast6Chars(this.newPassword)
+                        && isNotEmptyAndAtLeast6Chars(this.confirmPassword)
+                        && this.confirmPassword == this.newPassword
+            }
+
+            private fun isNotEmptyAndAtLeast6Chars(password: String) = !password.isEmpty() && password.length >= MINIMUM_PASSWORD_LENGTH
+        }
+    }
+
+    companion object {
```
Rcureton: You can define this at the top of the class I think instead of a companion object because no other class is accessing this constant

```
const val MINIMUM_PASSWORD_LENGTH = 6

interface ChangePasswordViewModel {

```

Issue #230, opened 2018-03-26
Status bar overlaps side navigation on the Essential phone


starkej2:
![screenshot_20180326-152005](https://user-images.githubusercontent.com/1320020/37935740-b10a656c-3120-11e8-92ea-80bf48ce4c11.png)

starkej2:
It looks like the layout has `fitsSystemWindows` applied properly, but it doesn't appear to be working properly on the Essential PH-1. Not sure if it's a problem w/ the phone or this app though. 

I tried to fix it by programmatically adding margin/padding equal to the status bar height to the top of the navigation drawer using `systemWindowInsetTop` provided by the `View.setOnApplyWindowInsetsListener`...didn't really help.

michaelsam94:
we can make the status bar transparent and follow material guidelines 
![material](https://user-images.githubusercontent.com/9461037/38063395-3dbe33a6-32f9-11e8-8d83-69ab243a2fb4.JPG)

eoji:
Hey Jeffrey(@starkej2)!! Thanks for the feedback. And thank you @michaelsam94 for the reference! This issue is logged in our roadmap and it will be in an upcoming release 🤓

eoji:
Hey Jeffrey 🤓 I finally got some time to address this. Thanks again for pointing it out!

starkej2:
🎃

PR #347, opened 2018-10-05
[bug] Fixing the drawer getting cut off and reformatting.

eoji:
# what
Finally had time to address https://github.com/kickstarter/android-oss/issues/230 🤓 Now the drawer won't get cut off for devices with cutouts.

# how
It was actually pretty straight-forward (as straight-forward as Android can be 😛)
The default value of `fitsSystemWindows` is `true` so we don't need it in the `DrawerLayout` but for `NavigationView` we do, so it draws below the status bar.

# before and after
  

# cool tipz
There's an option in Developer Settings to simulate a cutout. That's how I tested this out!

image


eoji:
[Close out last open issue in Android OSS repo](https://trello.com/c/oJmjl1AY/25-close-out-last-open-issue-in-android-oss-repo)

PR #350, opened 2018-10-08
[Enhancement] - Currency picker alert

Rcureton:
## What
- User can see the Currency Change Alert with the updated copy.
- Updated the strings from master.

## See 👀
oct-08-2018 16-14-11


app/src/main/res/values/styles.xml
```diff
@@ -95,6 +95,17 @@
     <item name="android:progressDrawable">@drawable/progress_bar_green_horizontal</item>
   </style>
 
+  <style name="AlertDialog" parent="Theme.AppCompat.Light.Dialog.Alert">
+    <item name="buttonBarPositiveButtonStyle">@style/AlertDialogButtonStyle</item>
+    <item name="buttonBarNegativeButtonStyle">@style/AlertDialogButtonStyle</item>
+    <item name="colorAccent">@color/text_secondary</item>
+  </style>
+
+  <style name="AlertDialogButtonStyle" parent="Widget.AppCompat.Button.ButtonBar.AlertDialog">
```
eoji: Sweet!

app/src/main/res/values/styles.xml
```diff
@@ -95,6 +95,17 @@
...
+    <item name="android:textColor">@color/ksr_green_500</item>
```
eoji: Should we refer to this is `colorAccent`?

app/src/main/res/layout/activity_account.xml
```diff
@@ -84,11 +84,13 @@
...
           <Spinner
             android:id="@+id/currency_spinner"
+            style="@style/Base.Widget.AppCompat.Spinner.Underlined"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:layout_marginEnd="@dimen/activity_horizontal_margin"
             android:layout_marginStart="@dimen/activity_horizontal_margin"
-            android:layout_marginTop="@dimen/grid_1">
+            android:layout_marginTop="@dimen/grid_1"
+            android:dropDownVerticalOffset="@dimen/grid_11">
```
eoji: I don't think you'll need this because the drop down list is drawn over the spinner.

Rcureton: Sounds good I'll remove it!

app/src/main/res/values/styles.xml
Rcureton: So it was ignoring the color when I used colorAccent so I went with TextColor

eoji: Sorry, I wasn't clear. I mean the color value you're passing in. Like this `<item name="android:textColor">@color/accent</item>`

Rcureton:
[Currency picker alert](https://trello.com/c/rXIzObr0/979-currency-picker-alert)

PR #352, opened 2018-10-10
change email validation

eoji:
# what did i do!
Switched to horizontal progress bar below toolbar.
No longer crashing when user can't connect to the server.
Added back `CustomAdapter` for `Email` because the mutation fails without it.
Added `StringUtils. isValidPassword`.
Added options menu to `ChangeEmailActivity` with save button that is enabled/disabled when input is valid or network call is in progress.
Hiding the keyboard when the user clicks save.
Created data class `ChangeEmail` so the validation is easier to follow in the view model.
Moved the start and end margin to the `TextInputLayout` instead of the `TextInputEditText` so the error is left aligned.
Showing error on email input when it loses focus and isn't valid.
Showing error snackbar. (will be restyled in a different PR)
Tests ✨ 

# see
device-2018-10-09-203325 2018-10-09 20_54_50 device-2018-10-09-203412 2018-10-09 20_55_12


eoji:
[Change email address: validation errors](https://trello.com/c/jXFssPTH/929-change-email-address-validation-errors)

PR #354, opened 2018-10-15
[Enhancement] - Currency Preferences on Project Pages

Rcureton:
## What
- User's that have selected and chosen currency can now view projects in their preferred currency.
- If the user hasn't chosen a currency then the default `US$` will be shown.
- Created the `UserCurrency` class which displays the proper currency symbol also the project's pledge and goal amounts based on the `fxRate`.

## See
oct-15-2018 10-33-59


Rcureton:
[Currency Preferences on Project Pages](https://trello.com/c/IrqRw0MZ/999-currency-preferences-on-project-pages)

app/src/main/java/com/kickstarter/libs/UserCurrency.kt
```diff
@@ -0,0 +1,125 @@
...
+
+import com.kickstarter.libs.utils.NumberUtils
+import com.kickstarter.models.Project
+import type.CurrencyCode
+import java.math.RoundingMode
+
+class UserCurrency(private val currentConfigType: CurrentConfigType) {
+
+    /**
+     * Returns a currency string appropriate to the user's locale and location relative to a project.
```
eoji: I am not sure if this is accurate to this class. Might be a case of copy pasta 🕵️‍♀️

app/src/main/java/com/kickstarter/libs/UserCurrency.kt
```diff
@@ -0,0 +1,125 @@
...
+     */
+    private fun userCurrencyOptions(value: Float, project: Project, symbol: String): KSCurrency.CurrencyOptions {
+        val fxRate = project.fx_rate()
+        val config = this.currentConfigType.getConfig()
+
+        return if (config.countryCode() == "XX") {
+            KSCurrency.CurrencyOptions.builder()
+                    .country(project.country())
+                    .currencyCode("")
+                    .currencySymbol("US$ ")
```
eoji: I'm not sure if this true for a US user looking at a US project. I think I'd see the project just as "$"

app/src/main/java/com/kickstarter/libs/UserCurrency.kt
```diff
@@ -0,0 +1,125 @@
...
+                .build()
+
+        return NumberUtils.format(currencyOptions.value(), numberOptions)
+    }
+
+    /** Show's the project in the user's preferred currency. If the user has no preferred currency the project is shown
+     * in $US as a default.
+     */
+    private fun userCurrencyOptions(value: Float, project: Project, symbol: String): KSCurrency.CurrencyOptions {
+        val fxRate = project.fx_rate()
```
eoji: what if this is null? Doing `!!` will still cause a crash

app/src/main/java/com/kickstarter/libs/UserCurrency.kt
```diff
@@ -0,0 +1,125 @@
...
+            CurrencyCode.HKD.rawValue() -> symbol = "HK$ "
+            CurrencyCode.JPY.rawValue() -> symbol = "¥ "
+            CurrencyCode.MXN.rawValue() -> symbol = "MX$ "
+            CurrencyCode.NOK.rawValue() -> symbol = "NOK "
+            CurrencyCode.NZD.rawValue() -> symbol = "NZ$ "
+            CurrencyCode.SEK.rawValue() -> symbol = "SEK "
+            CurrencyCode.SGD.rawValue() -> symbol = "S$ "
+            CurrencyCode.USD.rawValue() -> symbol = "$ "
+
+            else -> symbol = "$ "
```
eoji: Should we be returning US$ here?

app/src/main/java/com/kickstarter/models/User.java
```diff
@@ -19,6 +19,7 @@
   public abstract @Nullable Boolean artsCultureNewsletter();
   public abstract Avatar avatar();
   public abstract @Nullable Integer backedProjectsCount();
+  public abstract @Nullable String chosenCurrency();
```
eoji: I'm going to push back on this. Stay tuned for my followup in the viewmodel!

app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java
```diff
@@ -37,141 +43,243 @@
 public interface ProjectHolderViewModel {
 
   interface Inputs {
-    /** Call to configure view holder with a project and the config country. */
+    /**
```
eoji: Please revert all of these comments that got reformatted.

app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java
```diff
@@ -37,141 +43,243 @@
...
 
     public ViewModel(final @NonNull Environment environment) {
       super(environment);
+
+      this.client = environment.apiClient();
+      this.currentUser = environment.currentUser();
       this.ksCurrency = environment.ksCurrency();
+      this.userCurrency = environment.userCurrency();
+
+      this.client.fetchCurrentUser()
```
eoji: Here's the followup to my push back. We're updating the user using GraphQL, we should just query for their currencyPreference using GraphQL. The only info we want is the currency but we're pulling down the whole user object from REST. Then we won't need to update the user every time we look at a project.

app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java
```diff
@@ -350,131 +466,173 @@ public ViewModel(final @NonNull Environment environment) {
     public final Inputs inputs = this;
     public final Outputs outputs = this;
 
-    @Override public void configureWith(final @NonNull Pair<Project, String> projectAndCountry) {
+    @Override
```
eoji: Let's revert all of this reformatting.

app/src/test/java/com/kickstarter/UserCurrencyTest.kt
```diff
@@ -0,0 +1,76 @@
...
+                .countryCode(countryCode)
+                .build()
+
+        val currentConfig = MockCurrentConfig()
+        currentConfig.config(config)
+
+        return UserCurrency(currentConfig)
+    }
+
+    fun testFormatCurrency_withUserInUS() {
```
eoji: I'm not sure what these tests are testing. Is it testing that the user prefers a currency or they're in the a locale? Let's rename it to what it's actually testing.

app/src/main/java/com/kickstarter/libs/UserCurrency.kt
Rcureton: For a regular US user it won't go into that block because it's only for those in un-launched countries.

app/src/main/java/com/kickstarter/libs/UserCurrency.kt
Rcureton: It shouldn't be null we get this value with every project that we pull in.

app/src/test/java/com/kickstarter/UserCurrencyTest.kt
Rcureton: So these tests are testing where the user is and how the currency should be displayed based on the locale.

app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java
```diff
@@ -167,11 +177,22 @@
...
     public ViewModel(final @NonNull Environment environment) {
       super(environment);
+
+      this.apolloClient = environment.apolloClient();
       this.ksCurrency = environment.ksCurrency();
+      this.userCurrency = environment.userCurrency();
+
+      this.apolloClient.userPrivacy()
+        .map(currency -> Objects.requireNonNull(currency.me()).chosenCurrency())
+        .compose(Transformers.neverError())
```
eoji: I think this is causing the funded amount to be blank when a user is logged out or doesn't have service.

app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java
```diff
@@ -414,7 +444,8 @@ public ViewModel(final @NonNull Environment environment) {
     @Override public @NonNull Observable<Boolean> projectDisclaimerTextViewIsGone() {
       return this.projectDisclaimerTextViewIsGone;
     }
-    @Override public @NonNull Observable<Integer> projectMetadataViewGroupBackgroundDrawableInt() {
+    @Override
+    public @NonNull Observable<Integer> projectMetadataViewGroupBackgroundDrawableInt() {
```
eoji: let's do this all on 1 line like 
```@Override public @NonNull Observable<Integer> projectMetadataViewGroupBackgroundDrawableInt() {```

File: app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java (hunk 10 above)
Comment (eoji) [reply app/src/main/java/com/kickstarter/viewmodels/ProjectHolderViewModel.java]: I think adding tests for `ProjectHolderViewModel` will help you catch these edge cases!

File: app/src/main/java/com/kickstarter/libs/UserCurrency.kt [hunk 12]
```diff
@@ -0,0 +1,131 @@
...
+
+    }
+
+    /** Returns the proper currency symbol based on the user's chosenCurrency preference.  */
+    private fun currencySymbol(chosenCurrency: String): String {
+        val symbol: String
+        val config = this.currentConfigType.getConfig()
+
+        if (config.countryCode() == "US" && chosenCurrency == CurrencyCode.USD.rawValue()) {
+            symbol = "$ "
```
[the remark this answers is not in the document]
nnekab: @eoji correct -- remove the space between the currency symbol and amount/number. This is huge -- i'm really excited for this feature

app/src/main/java/com/kickstarter/libs/UserCurrency.kt
```diff
@@ -0,0 +1,132 @@
...
+     * $US.
+     */
+    private fun userCurrencyOptions(value: Float, project: Project, symbol: String): KSCurrency.CurrencyOptions {
+        val fxRate = project.fx_rate()
+        val config = this.currentConfigType.getConfig()
+
+        return if (config.countryCode() == "XX") {
+            KSCurrency.CurrencyOptions.builder()
+                    .country(project.country())
+                    .currencyCode("")
```
eoji: What is `currencyCode` used for and why do you pass "" for both the if and the else

app/src/main/java/com/kickstarter/models/Project.java
```diff
@@ -33,9 +33,11 @@
   public abstract User creator();
   public abstract String currency(); // e.g.: USD
   public abstract String currencySymbol(); // e.g.: $
+  public abstract String current_currency(); // e.g.: User's Preferred currency USD
```
eoji: I think this should be `currentCurrency`

app/src/main/java/com/kickstarter/models/Project.java
```diff
@@ -33,9 +33,11 @@
...
   public abstract boolean currencyTrailingCode();
   public abstract @Nullable DateTime featuredAt();
   public abstract @Nullable List<User> friends();
+  public abstract Float fx_rate();
```
eoji: I think this one should be `fxRate`

app/src/test/kotlin/UserCurrencyTest.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.kickstarter.mock.MockCurrentConfig
+import com.kickstarter.mock.factories.ConfigFactory
+import com.kickstarter.mock.factories.ProjectFactory
+import junit.framework.Assert
+import junit.framework.TestCase
+import type.CurrencyCode
+
+class UserCurrencyTest: TestCase() {
+
+    private fun createUserCurrency(countryCode: String): UserCurrency {
```
eoji: let's move private methods to the bottom of the class

app/src/test/kotlin/UserCurrencyTest.kt
```diff
@@ -0,0 +1,60 @@
...
+        val currentConfig = MockCurrentConfig()
+        currentConfig.config(config)
+
+        return UserCurrency(currentConfig)
+    }
+
+    fun testFormatCurrency_withUserInUS() {
+        val currency = createUserCurrency("US")
+        Assert.assertEquals("$100", currency.format(100.0f, ProjectFactory.project(), CurrencyCode.USD.rawValue()))
+        Assert.assertEquals("CA$100", currency.format(100.0f, ProjectFactory.caProject(), CurrencyCode.CAD.rawValue()))
```
eoji: I'm not understanding what these tests are testing. My understanding is if I prefer USD, if a project is in Canadian dollars, I'd see the conversion in USD.

For this project, I'd assume I would see US$100 because I prefer USD.

How does `UserCurrency` take in the user's currency preference?

app/src/main/java/com/kickstarter/libs/UserCurrency.kt
Rcureton: So after some deep diving, we actually we were actually never using it because it would've shown in the projects currency. Now we actually capture it and map it to the showCurrencySymbol method where we map to the display strings we want to show the user. I just removed it. Great catch!

app/src/test/kotlin/UserCurrencyTest.kt
Rcureton: So for these tests it tests the user's location and based upon the location and user's preferred currency it will show the proper currency symbol. 

So for the test above the user is in the `US` and if the user's preferred currency is `$US` or they don't have a preference it still should show as `$`.

eoji: But this test is asserting that when we call `currency.format(100.0f, ProjectFactory.caProject(), CurrencyCode.CAD.rawValue()`, it will output `"CA$100"`. Where does the user's location (`US`) come in? Shouldn't it be `US$100`?

eoji: why does the user's location matter? 👀

Rcureton: Location matters because the display of the currency like `$` or `US$` depends on the location of the user.

eoji: ok but this test is called `testFormatCurrency_withUserInUS` so i'm assuming the user is always in the US. is the test name accurate?

Rcureton: Yes isn't it? It's testing the format of the currency with the user in the US 🧐

Rcureton:
The story says that the rewards should be converted to the user's preferred currency.

eoji:
@nnekab Wanted to confirm if we prefer a currency that's different from the project's, we should see the reward in the original currency amount and the converted amount underneath.

Here's what it looks like now in the current version of the app. I'm viewing a project that's in Euros and I am defaulted to USD. The user sees the rewards in the project's currency with a conversion of the user's currency underneath

nnekab:
@dannyalright @eoji this is how we handle on web -- i think its fine as shown in the image you shared. I get that it's little confusing because we convert the project goal and then show reward amount conversion underneath. but im fine with leaving similar to web

eoji:
Will update on Trello 👍

Rcureton:
I refactored the methods to show the conversion when the project currency and user currency differ. 
screenshot-1540311255787
screenshot-1540311291540

app/src/main/java/com/kickstarter/models/Project.java
```diff
@@ -33,9 +33,11 @@
   public abstract User creator();
   public abstract String currency(); // e.g.: USD
   public abstract String currencySymbol(); // e.g.: $
+  public abstract String currentCurrency(); // e.g.: User's Preferred currency USD
   public abstract boolean currencyTrailingCode();
   public abstract @Nullable DateTime featuredAt();
   public abstract @Nullable List<User> friends();
+  public abstract Float fxRate();
```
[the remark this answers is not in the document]
Rcureton: So I confirmed with the API Team and the `fx_rate` should never be null.

https://kickstarter.slack.com/archives/C98LKNPB3/p1540405705000100

app/src/main/java/com/kickstarter/libs/KSCurrency.java
```diff
@@ -93,6 +125,70 @@ public KSCurrency(final @NonNull CurrentConfigType currentConfig) {
...
+  /** Show's the project in the user's preferred currency. If the user has no preferred currency the project is shown
+   * in $ as a default if the user is in the US. If the user is located outside of the US the default will show as
+   * $US.
+   */
+  private @NonNull CurrencyOptions userCurrencyOptions(final float value, final @NonNull Project project, final String symbol) {
+    final Float fxRate = project.fxRate();
+
+    return CurrencyOptions.builder()
+        .country(project.country())
+        .currencyCode("")
```
eoji: Do we need this?

app/src/main/java/com/kickstarter/libs/KSCurrency.java
```diff
@@ -67,6 +68,37 @@ public KSCurrency(final @NonNull CurrentConfigType currentConfig) {
...
   }
 
+  /**
+   * Returns a currency string appropriate to the user's locale and preferred currency.
+   *
+   * @param initialValue Value to display, local to the project's currency.
+   * @param project The project to use to look up currency information.
+   * @param symbol The currency symbol that should be shown next to the pledge and goal amounts.
+   */
+  public String formatWithUserPreference(final float initialValue, final @NonNull Project project, final String symbol) {
```
eoji: The only place this method is used is in tests, so we're not actually testing what we're showing the user. I don't think we need it since we are only using 
```formatWithUserPreference(final float initialValue, final @NonNull Project project, final @NonNull RoundingMode roundingMode, final String symbol)```

app/src/main/java/com/kickstarter/libs/KSCurrency.java
Rcureton: We need this because it's defined in the CurrencyOptions Model class. If we remove it, it won't build properly but also there's many other classes that use this KSCurrency class removing it altogether would affect the other classes.

eoji: I mean do we need it in this method?

Rcureton: Yes because we will get this error if it's removed:
`Caused by: java.lang.IllegalStateException: Missing required properties: currencyCode`

I think later we can actually remove it once we scope out how currency should be handled in those other classes.

PR #356, opened 2018-10-17
Fixes DiscoveryViewModelTest.testClearingPages

eoji:
# how
The `clearPages` output `Emits a list of pages that should be cleared of all their content.` which translates to all of the pages but the current one. Since we got rid of the last sort, our indices went from [0,1,2,3,4] to [0,1,2,3]. Doing `this.vm.inputs.discoveryPagerAdapterSetPrimaryPage(null, 4)` is an illegal condition since position 4 doesn't exist and it will never get "cleared" because it doesn't actually exist.


eoji:
[Remove Most Funded from sorts](https://trello.com/c/NX2J2YkZ/1015-remove-most-funded-from-sorts)

PR #366, opened 2018-10-29
[Enhance] Quick return Discovery CollapsingToolbarLayout

eoji:
### THE CHAMP IS HERE

# what
Quick returning the entire Discovery CollapsingToolbarLayout on scroll and stubbornly bringing back the elevation.
Removing the divider line with AlphaBehavior because it's no longer visually relevant when the Appbar is collapsed.

# see
device-2018-10-26-165952 2018-10-29 15_29_37

image


eoji:
[Don't hide discovery header/search/activity bar while scrolling](https://trello.com/c/sXmGqggQ/1042-dont-hide-discovery-header-search-activity-bar-while-scrolling)

PR #372, opened 2018-11-07
[Enhancement] - Send Verification email mutation and isEmailDeliverable

Rcureton:
## What
- Added the `mutation` for sendVerificaitonEmail
- Added fields for the user query `isEmailDeliverable`, `isCreator`
- If user is a creator the row will read `RE-SEND Verification Email` 
- if user is a backer the row will read `Send Verification Email` 
- If the email is undeliverable that message will show.

## Stories
 [Change Email Undeliverable](https://trello.com/c/6VkU85sS/1048-change-email-undeliverable-email-addresses)
[Resend verification email](https://trello.com/c/YjNCldjD/1035-email-update-resending-verification)


app/src/main/java/com/kickstarter/ui/activities/ChangeEmailActivity.kt
```diff
@@ -58,6 +59,28 @@ class ChangeEmailActivity : BaseActivity<ChangeEmailViewModel.ViewModel>() {
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe { showErrorSnackbar(change_email_layout, it) }
 
+        this.viewModel.outputs.isCreator()
+                .compose(bindToLifecycle())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe {
+                    verification_text_view.text = getString(R.string.Send_verfication_email)
```
eoji: what if they're already verified? we want to show this row? when does it ever hide? 🤔

app/src/main/java/com/kickstarter/viewmodels/ChangeEmailViewModel.kt
```diff
@@ -38,6 +42,15 @@ interface ChangeEmailViewModel {
         /** Emits a string to display when email update fails.  */
         fun error(): Observable<String>
 
+        /** Emits a boolean to determine if the user is a creator. */
+        fun isCreator(): Observable<Boolean>
```
eoji: But what does being a creator signify in this viewmodel? In the activity, it seems like we show some text so maybe it'd be like `shouldShowBlahBlahText`

app/src/main/java/com/kickstarter/viewmodels/ChangeEmailViewModel.kt
```diff
@@ -38,6 +42,15 @@ interface ChangeEmailViewModel {
...
+
+        /** Emits a boolean to determine if a user's email is deliverable. */
+        fun isDeliverable(): Observable<Boolean>
```
eoji: Same here, if it's not deliverable, what does that mean for the activity that subscribes.

app/src/main/java/com/kickstarter/viewmodels/ChangeEmailViewModel.kt
```diff
@@ -38,6 +42,15 @@ interface ChangeEmailViewModel {
...
+
+        /** Emits a boolean to display if the user's email is verified. */
+        fun isEmailVerified(): Observable<Boolean>
```
eoji: -whispers- same here

app/src/test/java/com/kickstarter/viewmodels/ChangeEmailViewModelTest.kt
```diff
@@ -100,6 +185,19 @@ class ChangeEmailViewModelTest : KSRobolectricTestCase() {
         this.saveButtonIsEnabled.assertValues(true, false, true, false)
     }
 
+    @Test
+    fun testSendVerificationEmail() {
```
eoji: let's also test if this fails

app/src/main/java/com/kickstarter/viewmodels/ChangeEmailViewModel.kt
```diff
@@ -38,6 +44,9 @@ interface ChangeEmailViewModel {
         /** Emits a string to display when email update fails.  */
         fun error(): Observable<String>
 
+        /** Emits a boolean to display if the user's email is verified. */
+        fun isEmailVerified(): Observable<Boolean>
```
eoji: Should we rename this to `sendVerificationIsVisible`?

PR #376, opened 2018-11-13
added register receiver to onResume

Rcureton:
## What
- Registering the receiver in onResume


app/src/main/java/com/kickstarter/libs/BaseActivity.java
```diff
@@ -117,8 +118,7 @@ protected void onStart() {
       .observeOn(AndroidSchedulers.mainThread())
       .subscribe(__ -> goBack());
 
-    final IntentFilter filter = new IntentFilter(ConnectivityManager.CONNECTIVITY_ACTION);
-    this.registerReceiver(this.connectivityReceiver, filter);
+    this.registerReceiver(this.connectivityReceiver, this.filter);
```
eoji: We should only register the receiver once, in `onResume`

PR #377, opened 2018-11-13
[(∩｀-´)⊃━☆ﾟ.*･｡ﾟ] Payment methods recycler animations

eoji:
# what
-Leveraged the new `ListAdapter` class that handles list changes more responsibly 😛 
-When a user adds or deletes a card, we now get these awesome animations for free.

Shoutout to @CDRussell for a great blog post about it: https://medium.com/@trionkidnapper/recyclerview-more-animations-with-less-code-using-support-library-listadapter-62e65126acdb

# how
-I didn't want to change `KSAdapter` because 16 classes extend it and that's not what this feature is about. Maybe for an Investment Day™, I could refactor all of them.
-So! I made a new class called `KSListAdapter` that extends `ListAdapter`. It still can handle sections and uses our existing `KSViewHolder` pattern ٩( ᐛ )و 

# look how cute

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/adapters/KSListAdapter.kt
+++ b/app/src/main/java/com/kickstarter/ui/adapters/KSListAdapter.kt
@@ -1,0 +1,193 @@
+package com.kickstarter.ui.adapters
+
+import android.support.annotation.LayoutRes
+import android.support.v7.recyclerview.extensions.ListAdapter
+import android.support.v7.util.DiffUtil
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import com.crashlytics.android.Crashlytics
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/ui/adapters/KSListAdapter.kt
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Gets the data object associated with a position.
+     */
+    protected fun objectFromPosition(position: Int): Any {
+        return objectFromSectionRow(sectionRowFromPosition(position))
+    }
+
+    private fun sectionRowFromPosition(position: Int): SectionRow {
+        val sectionRow = SectionRow()
```
Rcureton: So this is for rows?

PR #378, opened 2018-11-14
[QA FIX] - Fixed currency formatting case

Rcureton:
## What
- Fixed formatting for currency symbol when a project meets these cases:
 >Symbol for USD displays as US$ if the
project location is in the United States and
any of the following are true: • User’s currency preference is set to
currency other than USD
• User’s profile location outside of the US

## See 👀
**Scenario** :  1
- User location is Brooklyn, NY
- User currency pref is MXN
- Project is CA Based and CAD Currency

 
screen shot 2018-11-15 at 12 01 35 pm

**Scenario** :  2
- User location is Barcelona, SP
- User currency pref is HK $
- Project is US Based and USD Currency

screen shot 2018-11-15 at 12 05 55 pm

**Scenario** :  3
- User location is Brooklyn, NY
- User currency pref is MX$
- Project is US Based and USD Currency

screen shot 2018-11-15 at 3 01 10 pm

**Scenario** :  4
- User location is Brooklyn, NY
- User currency pref is USD
- Project is US Based and USD Currency


app/src/main/java/com/kickstarter/libs/KSCurrency.java
```diff
@@ -68,6 +68,25 @@ public KSCurrency(final @NonNull CurrentConfigType currentConfig) {
...
   }
 
+  /**
+   * Returns a currency string appropriate to the user's locale, chosenCurrency and project preferred currency.
+   *
+   * @param initialValue Value to display, local to the project's currency.
+   * @param project The project to use to look up currency information.
+   * @param roundingMode This determines whether we should round the values down or up.
+   */
+  public String formatWithRewardPreference(final float initialValue, final @NonNull Project project, final @NonNull RoundingMode roundingMode) {
```
eoji: I'm not sure what a "reward preference" is. And we also use this method for the converted pledge and goal amounts. I think a better name could be "formatWithProjectCurrency" since I think that's what this method is doing.

app/src/main/java/com/kickstarter/libs/KSCurrency.java
```diff
@@ -157,27 +190,76 @@ private String getSymbolForCurrency(final String chosenCurrency) {
...
-      return "US$";
+      return "US$ ";
+    } else {
+      symbol = "US$ ";
+    }
+    return symbol;
+  }
+
+  /** Returns the proper currency symbol for the reward pledge amount **/
+  private String getSymbolForRewardCurrency(final Project project) {
```
eoji: What's the difference between this method and getSymbolForCurrency. Smells like copy pasta 🍝

Rcureton: `getSymbolForRewardCurrency` takes in a project and checks against the project currency and `getSymbolForCurrency` takes in the user's preferred currency then displays the correct symbol. They do the same thing just different parameters.

Rcureton: fixed!

app/src/test/java/com/kickstarter/viewmodels/ProjectHolderViewModelTest.java
```diff
@@ -498,7 +498,8 @@ public void testProjectStatsEmit() {
 
   @Test
   public void testUsdConversionForNonUSProject() {
-    final Project project = ProjectFactory.caProject();
+    // Use a CA project with a CAD currency
```
eoji: This is a CA project with MX currency.

Rcureton: fixed

PR #380, opened 2018-11-19
[Enhancement] - Change email error on Account page

Rcureton:
## What
- The email error icon will show in the account screen when a `user` is a creator and they change their email and do not `verify` it or the email is `undeliverable`. For a backer the error icon will show when their email is `undeliverable`.

## Story
[Change email error on Account page](https://trello.com/c/I7vN4lVG/1068-change-email-error-on-account-page)


app/src/main/java/com/kickstarter/viewmodels/AccountViewModel.kt
```diff
@@ -94,10 +105,22 @@ interface AccountViewModel {
...
         }
 
+        override fun showEmailErrorIcon(): Observable<Boolean> = this.showEmailErrorIcon
+
         override fun success(): BehaviorSubject<String> {
             return this.success
         }
 
+        private fun showEmailErrorImage(userPrivacy: UserPrivacyQuery.Data?): Boolean? {
+            return if (userPrivacy?.me()?.isCreator!! && !userPrivacy.me()?.isDeliverable!!) {
```
eoji: 🐛 👀

app/src/test/java/com/kickstarter/viewmodels/AccountViewModelTest.kt
```diff
@@ -55,4 +58,17 @@ class AccountViewModelTest : KSRobolectricTestCase() {
         this.chosenCurrency.assertValues("USD", CurrencyCode.AUD.rawValue())
     }
 
+    @Test
+    fun testShowEmailErrorIcon() {
```
eoji: let's add for a backer that's undeliverable and a creator that's unverified

app/src/main/java/com/kickstarter/viewmodels/AccountViewModel.kt
```diff
@@ -94,10 +105,22 @@ interface AccountViewModel {
...
+        override fun showEmailErrorIcon(): Observable<Boolean> = this.showEmailErrorIcon
+
         override fun success(): BehaviorSubject<String> {
             return this.success
         }
 
+        private fun showEmailErrorImage(userPrivacy: UserPrivacyQuery.Data?): Boolean? {
+            return if (userPrivacy?.me()?.isCreator!! && !userPrivacy.me()?.isDeliverable!! || !userPrivacy.me()?.isEmailVerified!!) {
+                return true
+            } else if (!userPrivacy.me()?.isDeliverable!! && !userPrivacy.me()?.isEmailVerified!!) {
```
eoji: For this one, I thought for backers, we are only uninterested in if the email is undeliverable and we don't care if it's unverified.

We should also have a test for this case. That we don't show the icon for unverified backers.

app/src/main/res/layout/activity_account.xml
```diff
@@ -36,10 +36,30 @@
...
+              android:text="@string/Change_email" />
+
+            <ImageView
+              android:id="@+id/email_error_icon"
+              android:layout_width="wrap_content"
+              android:layout_height="wrap_content"
+              android:layout_marginBottom="@dimen/activity_horizontal_margin"
+              android:layout_marginEnd="@dimen/activity_horizontal_margin"
+              android:layout_marginTop="@dimen/activity_horizontal_margin"
+              android:layout_weight="0"
```
eoji: Why are we using a weight here? The default weight is 0. Also, instead of setting top and bottom layout margins, we can just give this view a gravity of `center_vertical`

app/src/main/java/com/kickstarter/viewmodels/AccountViewModel.kt
Rcureton: fixed

app/src/main/res/layout/activity_account.xml
Rcureton: fixed

app/src/main/java/com/kickstarter/ui/activities/AccountActivity.kt
```diff
@@ -57,7 +57,7 @@ class AccountActivity : BaseActivity<AccountViewModel.ViewModel>() {
         this.viewModel.outputs.showEmailErrorIcon()
                 .compose(bindToLifecycle())
                 .compose(observeForUI())
-                .subscribe { ViewUtils.setGone(email_error_icon, !it) }
+                .subscribe { ViewUtils.setGone(email_error_icon, it) }
```
eoji: nice catch 😛

app/src/main/java/com/kickstarter/viewmodels/AccountViewModel.kt
```diff
@@ -114,7 +114,7 @@ interface AccountViewModel {
         private fun showEmailErrorImage(userPrivacy: UserPrivacyQuery.Data?): Boolean? {
             return if (userPrivacy?.me()?.isCreator!! && !userPrivacy.me()?.isDeliverable!! || !userPrivacy.me()?.isEmailVerified!!) {
```
eoji: you're going to be sad but i realized the logic is simpler like this:
 ```private fun showEmailErrorImage(userPrivacy: UserPrivacyQuery.Data?): Boolean? {
            return if ( !userPrivacy.me()?.isDeliverable!!) {
                return true
            } else if (userPrivacy?.me()?.isCreator!! && !userPrivacy.me()?.isEmailVerified!!) {
                return true
            } else {
                false
            }
        }
```

File: app/src/test/java/com/kickstarter/viewmodels/AccountViewModelTest.kt [hunk 7]
```diff
@@ -82,6 +82,18 @@ class AccountViewModelTest : KSRobolectricTestCase() {
         this.showEmailErrorIcon.assertValue(true)
     }
 
+    @Test
+    fun testShowEmailErrorIconGoneForBackerUndeliverable() {
```
eoji: what's the difference between this test and `testShowEmailErrorIconForBackerUndeliverable`? This code is kind of hard to read because idk what these booleans correspond to `true, false, true`. So it's be helpful if we had variables like `val isCreator = false` for example and pass those in

Rcureton: fixed!

app/src/main/java/com/kickstarter/viewmodels/AccountViewModel.kt
Rcureton: fixed!

app/src/main/java/com/kickstarter/viewmodels/AccountViewModel.kt
```diff
@@ -94,10 +105,22 @@ interface AccountViewModel {
...
         }
 
+        override fun showEmailErrorIcon(): Observable<Boolean> = this.showEmailErrorIcon
+
         override fun success(): BehaviorSubject<String> {
             return this.success
         }
 
+        private fun showEmailErrorImage(userPrivacy: UserPrivacyQuery.Data?): Boolean? {
+            return if (!userPrivacy?.me()?.isDeliverable!!) {
```
eoji: I got some helpful feedback from @Scollaco. When you use `!!` and this value is null, it's going to crash the app. We shouldn't assume that it's not going to be null. Here's a helpful link I found: https://kotlinlang.org/docs/reference/null-safety.html

Rcureton: Fixed!

eoji: amazing

Rcureton:
[Change email error on Account page](https://trello.com/c/I7vN4lVG/1068-change-email-error-on-account-page)

PR #381, opened 2018-11-20
[📵] Updating offline copy

eoji:
# what
We got some new copy. Thanks you @vickisiolos 

# see
device-2018-11-20-123352 2018-11-20 12_35_31


eoji:
[Update copy for offline Snackbar.](https://trello.com/c/mO3HsguG/1083-update-copy-for-offline-snackbar)

PR #389, opened 2018-11-27
Sending Settings v2 events to Koala

Rcureton:
## What
- added event attributes and sending setting v2 events to koala.


Rcureton:
I'll hold off on the other stuff and only add the fabric custom event that way our work won't mix 👍🏾

app/src/main/java/com/kickstarter/extensions/AnswersExt.kt
```diff
@@ -6,3 +6,7 @@ import com.crashlytics.android.answers.CustomEvent
 fun fabricLogCustomEvent(customEvent: String) {
     Answers.getInstance().logCustom(CustomEvent(customEvent))
 }
+
+fun fabricLogCustomEventWithAttributes(customEvent: String, key: String, attribute: String) {
```
eoji: This method sends a single attribute along with an event. My understanding was that we wanted to send all of the attributes we send to Koala to Fabric/Firebase. Which is why @justinswart was running into the limit issue. I know Justin said he's going to do some more exploration with Fabric/Firebase.

Rcureton: I think the issue was that he was sending all of the default properties along with the custom event which we don't need to do because we get that from Fabric automatically. (Device, OS, etc..)

justinswart: Oh cool, @Rcureton please show me where to find that in Fabric, I could only see the custom attributes on the event that I was logging. Also - I'm not seeing any of that stuff appearing in Firebase yet 🤔

Rcureton: So for us to be able to see events in Firebase we'll have to use the Firebase SDK. Right now the only thing that transfers over are the crash reports. Also I thought `Fabric` had a set of automatically collected properties with each events but apparently `Firebase` has them [Firebase Automatically Collected Properties](https://support.google.com/firebase/answer/6317486?hl=en&ref_topic=6317484). I don't think it will be a large change for us on both platforms. What do you think?

app/src/main/java/com/kickstarter/libs/Koala.java
```diff
@@ -344,9 +345,14 @@ public void trackSavedPaymentMethod() {
...
-    this.client.track(KoalaEvent.SELECTED_CHOSEN_CURRENCY);
-    AnswersExtKt.fabricLogCustomEvent(KoalaEvent.SELECTED_CHOSEN_CURRENCY);
+  public void trackSelectedChosenCurrency(final String selectedCurrency) {
+    this.client.track(KoalaEvent.SELECTED_CHOSEN_CURRENCY, new HashMap<String, Object>() {
+      {
+        put("user_chosen_currency", selectedCurrency);
+      }
+    });
+
+    AnswersExtKt.fabricLogCustomEventWithAttributes(KoalaEvent.SELECTED_CHOSEN_CURRENCY, "currency", selectedCurrency);
```
eoji: I thought we're not doing any more Fabric events. 👀

app/src/main/java/com/kickstarter/extensions/AnswersExt.kt
```diff
@@ -6,3 +6,7 @@ import com.crashlytics.android.answers.CustomEvent
 fun fabricLogCustomEvent(customEvent: String) {
```
eoji: Should we kill this extension until we decide we're using Fabric officially?

app/src/main/java/com/kickstarter/libs/Koala.java
Rcureton: Removed!

app/src/main/java/com/kickstarter/extensions/AnswersExt.kt
Rcureton: Killed!

PR #395, opened 2018-11-29
Change Email Fix

Rcureton:
## What
- removed not null from change email and added logic for showing icon in account.


eoji:
Found one more small bug: the change email row clickability didn't include the warning icon. I fixed it 😅

PR #398, opened 2018-11-30
[🚮] MixPanel removal

eoji:
# what
Removing Mixpanel from the Android codebase.

# background
When we switched over to Koala, we continued to use the Mixpanel library to send events but we also continued to use a Mixpanel provided value here:
```
hashMap.put("device_fingerprint", this.mixpanel.getDistinctId());
hashMap.put("android_uuid", this.mixpanel.getDistinctId());
```

`getDistinctId` is defined as

>The id returned by getDistinctId is independent of the distinct id used to identify any People Analytics properties in Mixpanel.

Which is great…for Mixpanel!

# remediation
To fully set Mixpanel free, we need to change this value, which I don’t think will affect tracking too much (a big assumption) I’m just changing the value representing unique users. I was thinking of using the same value we use for the visitor cookie. Which is `FirebaseInstanceId.getInstance().getId()`. It gets reset when users log in or out. That makes the most sense to me. (I'm open to being wrong (my greatest weakness)).

I created a `KoalaService` API wrapper to post events to the `/native/track` endpoint.

I created a `KoalaBackgroundService` that uses `JobDispatcher` to send tracking events in the background. It comes with free automatic retries woohoo. Also, it doesn’t run the jobs unless there’s a network connection.

**TESTS**
Previously, we were only testing that Koala tracked an event by asserting the event's name was present. But we never tested the properties we were sending. 

A non exhaustive list of discrepancies I found:
- Android was never tracking "project_duration"
- Android was tracking project creator info like ""project_creator_uid" -> "3" vs "creator_uid" -> "3" on iOS
- Android was tracking project user info like ""project_user_is_project_creator" -> "3"" vs "user_is_project_creator" -> "3" on iOS
- We never tracked user info when tracking projects. We always passed in a null user 😧 

The Android codebase now tests every property iOS tests, except live streaming because we don't have that feature.

# note
Previously, all debug builds sent Koala events to staging, even when we were hitting the production API. After this gets merged, staging API will point to staging Koala and prod API will point to prod Koala. All the more reason to develop on staging 🙂  

You made it to the end. You are strong. HAPPY INVESTMENT DAY
image

cc @kickstarter/data


eoji:
[Removing Mixpanel library from Android codebase, part 5...(I'm tired)](https://trello.com/c/P6i1OIIv/115-removing-mixpanel-library-from-android-codebase-part-5im-tired)

app/src/main/java/com/kickstarter/libs/utils/KoalaUtils.java
```diff
@@ -101,33 +106,27 @@ private KoalaUtils() {}
...
     final Map<String, Object> properties = new HashMap<String, Object>() {
       {
         put("backers_count", project.backersCount());
+        put("comments_count", project.commentsCount());
         put("country", project.country());
+        put("duration", Math.round(ProjectUtils.timeInSecondsOfDuration(project)));
         put("currency", project.currency());
         put("goal", project.goal());
-        put("pid", project.id());
+        put("has_video", project.video() != null);
```
Rcureton: So if the `project.video = null` it won't put the value in the map?

app/src/main/java/com/kickstarter/libs/utils/KoalaUtils.java
```diff
@@ -101,33 +106,27 @@ private KoalaUtils() {}
...
+        put("hours_remaining", (int) Math.ceil(ProjectUtils.timeInSecondsUntilDeadline(project) / 60.0f / 60.0f));
         put("name", project.name());
```
Rcureton: What do you think about adding a method in `ProjectUtils` to do these calculations for `hours_remaining` and `percent_raised`?

app/src/main/java/com/kickstarter/libs/KoalaTrackingClient.java
```diff
@@ -2,100 +2,168 @@
...
-  public void track(final @NonNull String eventName, final @NonNull Map<String, Object> properties) {
-    final Map<String, Object> newProperties = new HashMap<>(properties);
-    newProperties.putAll(defaultProperties());
-
-    this.mixpanel.trackMap(eventName, MapUtils.compact(newProperties));
+  public void track(final @NonNull String eventName, final @NonNull Map<String, Object> additionalProperties) {
+    try {
+      final String trackingData = getTrackingData(eventName, combinedProperties(additionalProperties));
+      final String encodedTrackingData = Base64Utils.encodeUrlSafe(trackingData
+        .getBytes());
```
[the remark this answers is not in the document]
eoji: Yep!

app/src/main/java/com/kickstarter/libs/utils/KoalaUtils.java
eoji: No, I'm putting in a boolean called "has_video", so it's either true or false.

app/src/main/java/com/kickstarter/libs/utils/KoalaUtils.java
eoji: -hisses- no. It's only used here. 😛

terminalbae:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 @eoji

nnekab:
SO EXCITED 

image

abelsonlive:
very sick izzy!

PR #401, opened 2018-12-04
[Enhancement] - added 1 password autofill for change password, change email and login

Rcureton:
## What ❓
- added the `1Password` autofill for `ChangePassword`, `ChangeEmail` and `Login` screens.

## Story 📖
[Investigate using autofill for Change Password - Current and New Password](https://trello.com/c/MvW8qizp/1074-investigate-using-autofill-for-change-password-current-and-new-password)


dannyalright:
*Log in:* if I manually type in my email address, and then go to the password field, does the 1Password autofill come up there?

Or only in the email field?

Rcureton:
@dannyalright It comes up in both fields.

Rcureton:
I verified that this works without adding the `importantForAutofill` so we don't need to add this. Good catch!

PR #412, opened 2018-12-12
Pull Request Template

Rcureton:
Here's a basic template for our PR's! Please let me know if there's something you would like to add.


eoji:
love the emojis!

Rcureton:
Sometimes I think it's needed but in many cases it's not. Like when we implement new features we don't need a `why` but when we are changing the flow or how a feature works then a `why` is needed. Agree/Disagree?

PR #415, opened 2018-12-14
[Enhancement] - Settings and child activities transitions

Rcureton:
# What ❓
- Added `v21` anim for `slide_from_bottom` animation.
- Added the `exitTransitions` for ChangeEmail, ChangePassword, Privacy, Account, Newsletter, Notifications. 
- Added `windowIsTranslucent">true</item>` to KSTheme this is what removes the black background when animations are happening.

# Story 📖

[Settings Transitions](https://trello.com/c/NUyWtYhA/1108-some-settings-views-need-transitions?menu=filter&filter=label:Android#)

# See 👀
settings-slide


dannyalright:
yeah, I'd love to have some alpha to smooth out that disappearance. Good catch, @eoji. I think the Settings app on 27+ is a good reference for how this animation should behave, if it helps.

eoji:
Here are the 2 animations side by side. They're not the same but will defer to @dannyalright on approving the PR.
device-2019-01-03-105128 2019-01-03 10_56_27 device-2019-01-03-105409 2019-01-03 10_56_23

dannyalright:
thanks for that nifty comparison @eoji. My understanding from pairing with @Rcureton was that it wasn't possible to get the same spring/bezier curve animation as the Settings app, and what @Rcureton built is as close as possible as we could get in a reasonable amount of time.

If that's not the case or things have changed, it would be sweet to get it closer. Otherwise we can ship it as-is.

Rcureton:
Hey @eoji !!

@dannyalright and I paired on this PR this morning and he signed off on the design.

dannyalright:
just following up on this and ditto-ing what @Rcureton said: we paired on this and couldn't find a templated activity transition for the one used in the default Settings app. We're going to use the one Rashad made as our default and iterate from there.

PR #421, opened 2019-01-03
[Tech Debt] - Updated Stripe library

Rcureton:
# What ❓

- Updated the `Stripe` version to `8.1.0`

# Story 📖

[External Library Sweep](https://trello.com/c/c8nq6Sin/1147-external-library-sweep)


eoji:
Have you also tested this on the settings v3 branch? That's where we actually use Stripe

Rcureton:
Tested this on `v3` I added a new card it works!.

PR #424, opened 2019-01-04
Updated Fastlane version

Rcureton:
# What ❓

- Updated Gemlock file with latest Fastlane version


Rcureton:
Yes I think it should work and update the missing files that circle couldn't find. It didn't fail the initial build and test but let's hope it makes the alpha

PR #429, opened 2019-01-08
[UI] Snackbar style update

eoji:
# What ❓
Removing rounded corners and margin from Snackbars because they were all forced to the bottom in API 28.

# Story 📖
[Update Snackbar Styles](https://trello.com/c/ZgWvg0ok/1151-update-snackbar-styles)

# B4 & After 👀

### DIFF
--- a/app/src/main/java/com/kickstarter/extensions/SnackbarExt.kt
+++ b/app/src/main/java/com/kickstarter/extensions/SnackbarExt.kt
@@ -44,8 +44,8 @@ fun Snackbar.setTextColor(colorId: Int): Snackbar {
     return this
 }
 
-fun Snackbar.style(gravity: Int, textColorId: Int, backgroundId: Int) {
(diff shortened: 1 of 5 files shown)


eoji:
[Update Snackbar styles](https://trello.com/c/ZgWvg0ok/1151-update-snackbar-styles)

dannyalright:
Given that we had to lose the bottom margin/inset and that we're now using API 28, I think we should slowly migrate this over to the [native Snackbar pattern](https://material.io/design/components/snackbars.html) (with our primary and accent colors).

Let me know if you want me to make a card for this, or if we should just add it to our design pairing @eoji.

eoji:
Like left aligning them? Yeah, let's make a card. Let's actually make a giant API 28 theme card and everything we're updating to a list with checkboxes so we can keep track of the changes!

dannyalright:
Nice. Added [here](https://trello.com/c/PtSIgPet/654-api-28-design-updates)

PR #431, opened 2019-01-10
updated internal and external version name

Rcureton:
# What ❓

- Updated `Internal` and `External` version names.


eoji:
I was pretty busy yesterday so feel free to take other team members so I don't hold you up next time!

Rcureton:
Alright! Thanks

PR #435, opened 2019-01-17
[Bug Fix] - Android 8 + Samsung orientation theme fix

Rcureton:
# What ❓

- So with the adoption of `Android 8` we can't change the orientation for an `Activity` that has a translucent background according to several [stackoverflow](https://stackoverflow.com/a/50832408) posts. The crash was occurring in the `VideoActivity` where the screen changes from `portrait` to `landscape` and since our overall theme had the `<item name="android:windowIsTranslucent">true</item>` it was crashing on Samsung devices and some other devices running `Android 8`. So to fix this issue I removed the `<item name="android:windowIsTranslucent">true</item>` from the main parent theme and created a style for the activities in `Settings` where we're using this attribute to remove the black background during activity transitions. 

- `ChangePassword`,`ChangeEmail`, `Notifications`, `Newsletters`, `Privacy`, `Account`, `Help`

# Story 📖

[Android 8 orientation fix](https://trello.com/c/4LiHQD34/1170-android-8-orientation-crash)

# See 👀
vid
jan-17-2019 14-45-19

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -29,7 +29,8 @@
 
     <!-- ACTIVITIES -->
 
-    <activity android:name=".ui.activities.AccountActivity" />
+    <activity android:name=".ui.activities.AccountActivity"
+      android:theme="@style/SettingsActivitiesTransitions"/>
     <activity
       android:name=".ui.activities.ActivityFeedActivity"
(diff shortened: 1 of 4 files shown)


Rcureton:
[Android 8 Orientation Crash](https://trello.com/c/4LiHQD34/1170-android-8-orientation-crash)

PR #445, opened 2019-01-25
[🚮] Removing deprecated calls to pipeErrorsTo in Newsletters and Edit Profile

eoji:
# What ❓
`Transformers.pipeErrorsTo` is _deprecated_. Instead, we should use `Observable.materialize` and handle the error state and the values state.
Started with `EditProfileViewModel` and `NewsletterViewModel`.

# How to QA? 🤔 
- [ ] Newsletter screen successfully toggles newsletter preferences and shows errors.
- [ ] Edit profile screen (only available on debug builds) successfully displays profile, toggles privacy preference and shows errors.

image

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/EditProfileViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/EditProfileViewModel.kt
@@ -86,6 +86,7 @@ interface EditProfileViewModel {
 
             updateUserNotification
                     .compose(errors())
+                    .compose(bindToLifecycle())
                     .subscribe(this.unableToSavePreferenceError)
 
             this.userInput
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/viewmodels/EditProfileViewModel.kt
```diff
@@ -47,87 +50,98 @@ interface EditProfileViewModel {
...
         private fun success(user: User) {
             this.currentUser.refresh(user)
             this.updateSuccess.onNext(null)
         }
 
-        private fun updateSettings(user: User): Observable<User> {
+        private fun updateSettings(user: User): Observable<Notification<User>>? {
             return this.client.updateUserSettings(user)
-                    .compose(Transformers.pipeErrorsTo(this.unableToSavePreferenceError))
+                    .materialize()
```
Rcureton: 👍🏾

PR #449, opened 2019-01-25
[🚮] Removing deprecated calls to pipeErrorsTo in Notifications

eoji:
# What ❓
`Transformers.pipeErrorsTo` is deprecated. Instead, we should use `Observable.materialize` and handle the error state and the values state.
Removing `pipeErrorsTo` from `NotificationsViewModel`. Also displaying when there's an error (added a test). Cleaning up huge `onCreate` in `NotificationsActivity`.

# how to test ⁉️ 
Notification screen successfully:

- [ ] updates user notification preferences
- [ ] displays an error toast when a user is offline

image


app/src/main/java/com/kickstarter/ui/activities/NotificationsActivity.kt
```diff
@@ -305,6 +213,108 @@ class NotificationsActivity : BaseActivity<NotificationsViewModel.ViewModel>() {
         view.contentDescription = contentDescription
     }
 
+    private fun setUpClickListeners() {
```
Rcureton: Nice clicklistener cleanup 😉

PR #458, opened 2019-02-04
Created standard styles for TextInputLayout

eoji:
# What ❓
- Created a `BaseTextInputLayout` and then we basically have 2 types. One that lives in a form view (its parent has padding so it doesn't need start or end margin. And the ones that are used in Settings, they need `activity_horizontal_margin` on the start and end.
- Per documentation: `The hint should be set on the TextInputLayout, rather than the EditText`
- Created a `BasePasswordTextInputLayout` to handle showing the visibility icon and changing its color.
- (Bonus fun times: making payment methods row visible)

# Story 📖
[Trello](https://trello.com/c/xHWVc2aY/1178-settings-v3-sweep)

- [ ] All filled text fields: Remove gap from bottom line (and make the thickness default)
- @dannyalright, at this time centering the input is not possible https://github.com/material-components/material-components-android/issues/120

# See 👀
`FormTextInputLayout`

`SettingsTextInputLayout`


ekussberg:
The layout here is still broken, the hint should be positioned vertically center when the input field is empty. In your case it is shifted to the bottom.

eoji:
@ekussberg, yep! Here's the issue opened on the material components repo https://github.com/material-components/material-components-android/issues/120. Hope they fix it soon!

dannyalright:
> Should all of them have the same top and bottom margin

@eoji yes please. The 'same' value should be similar to the _Create Password example_ below. The Material guidelines are pretty vague as to why that much vertical space—I take it's for assistive text.

Even if we don't use assistive text, (i.e. maybe Login) we should use the same spacing for visual consistency.

I _think_ that's what you were double-checking. Let me know if you meant something else.

| Create Password example  | Material [example](https://material.io/design/components/text-fields.html#usage) |
| ------------- | ------------- |
| text field spec  | Material example  |

PR #467, opened 2019-02-07
[Bug] fixed the delay in the pledge activity digest spinner

Rcureton:
# What ❓
- Paired with @eoji on fixing the delay in the `NotificationsActivity` it was due the emissions of the `userInput` and `userOutput` being in separate observables. 

# How to QA? 🤔
> Upon deselecting 'Project activity' if I have 'Daily summary' as my chosen value, the spinner will first change from 'Daily summary' to 'Twice a day summary', then close. If I re-enable 'Project activity', it will start from 'Twice a day summary'.

- It should now just collapse the spinner without changing the values.

# See 👀
pledge-fix


dannyalright:
nice!

PR #468, opened 2019-02-07
Koala Settings Event Name Update

Rcureton:
# What ❓
- Apparently the event name we were using for Settings was deprecated so I updated it to the new event name. 
- Also added the tracking events for `Payment Methods` they were there just not wired up. Added them to the tests as well.


eoji:
I'm not sure what `deprecated` means so can you please confirm we need one and not both?

Rcureton:
@eoji So I confirmed with @peat for this PR that we can fully deprecate `Settings View` since the rest of the organization isn't using it to track anything.

PR #472, opened 2019-02-08
Drawer item updates

eoji:
# What ❓
Styled:
- top filters
- category parents
- category children

# Story 📖
[Trello](https://trello.com/c/oPMLypl6/721-improve-navigation-and-filtering-on-android-app)

# before and after 👀


app/src/main/res/values/colors.xml
```diff
@@ -40,12 +40,13 @@
...
   <color name="black_alpha_17">#2B000000</color>
   <color name="black_alpha_30">#4D000000</color>
-  <color name="black_alpha_50">#88000000</color>
+  <color name="black_alpha_50">#80000000</color>
   <color name="blue_darken_10">@color/ksr_cobalt_600</color>
   <color name="gray_alpha_50">@color/ksr_grey_300</color>
   <color name="gray_darken_10">@color/ksr_grey_500</color>
-  <color name="green_alpha_20">#33037362</color>
-  <color name="green_alpha_50">#80037362</color>
+  <color name="green_alpha_18">#2e009E74</color>
```
dannyalright: My bad; I've realised that `green_alpha_18` is not accessible with our primary green as text on top of it. Could we please make and use `green_alpha_10` instead?

dannyalright:
thanks for fixing! I'm also noticing that the green background tint doesn't have a matching margin from the right-side of the drawer. Not important enough to hold merge though, I reckon.

eoji:
It's 6dp on both sides 🕵️‍♀️

dannyalright:
my mistake—didn't take into account the gutter GitHub puts in between images

eoji:
[Improve navigation and filtering on Android app](https://trello.com/c/oPMLypl6/1214-improve-navigation-and-filtering-on-android-app)

PR #477, opened 2019-02-12
Settings transitions

eoji:
# What ❓
Created standard style `SettingsActivity` for settings transitions  so we don't need to override `exitTransition` everywhere.

Removed Discovery -> Settings transition.

# Story 📖
[Trello](https://trello.com/c/xHWVc2aY/1178-settings-v3-sweep)

# See 👀
**From Discovery** 
device-2019-02-12-162352 2019-02-12 16_26_19

**Top Level Settings**
device-2019-02-12-161633 2019-02-12 16_26_45

**Account & Payment methods**
(change email is omitted from gif on purpose)
device-2019-02-12-162915 2019-02-12 16_30_22

**Notifications**
device-2019-02-12-162118 2019-02-12 16_26_03

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -62,8 +62,9 @@
     <item name="android:windowBackground">@color/white</item>
   </style>
 
-  <style name="SettingsActivitiesTransitions" parent="KSTheme">
+  <style name="SettingsActivity" parent="KSTheme">
     <item name="android:windowIsTranslucent">true</item>
+    <item name="android:windowAnimationStyle">@style/SettingsActivityTransitions</item>
   </style>
 
   <style name="ThanksActivity" parent="KSTheme">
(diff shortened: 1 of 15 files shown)


dannyalright:
> Changed Discovery -> Settings transition to slide up.

This feels a bit jarring to me. Agreed with the sentiment, though. How about we remove the transition completely? The Drawer -> Section transition (instantaneous, in this suggestion) should occur before the drawer closes. Gmail on Android is a good example of this.

eoji:
Gmail is using Fragments, since we're using Activities, it's not possible (we're going from 1 screen with a drawer to a new one). Here's what it looks like with no transition:
device-2019-02-12-170149 2019-02-12 17_06_10

dannyalright:
Gotcha. The Activities no-transition still feels less interruptive to me—at
least in that GIF format.

app/src/main/res/values/styles.xml
```diff
@@ -85,6 +86,14 @@
     <item name="android:windowBackground">@color/white</item>
   </style>
 
+  <!-- Activity Transitions-->
+  <style name="SettingsActivityTransitions" parent="android:style/Animation.Activity">
+    <item name="android:activityOpenEnterAnimation">@anim/settings_slide_in_from_bottom</item>
+    <item name="android:activityOpenExitAnimation">@null</item>
+    <item name="android:activityCloseEnterAnimation">@null</item>
```
Rcureton: ✨

PR #483, opened 2019-02-14
Change email design updates

eoji:
# What ❓
- Added new button style `AlternatePrimaryButton` with white background, accent green text and accent green with alpha ripple.
- Moved `ProgressBar` out of scroll view so it's always visible below the toolbar.
- Requesting keyboard focus on the current email field when the screen is opened.
- Initially hiding verification button and warning so they don't flicker when the screen opens.

# Story 📖
[Trello](https://trello.com/c/CbqrFA61/1215-change-email-screen-updates)

#


eoji:
[Change email screen updates](https://trello.com/c/CbqrFA61/1215-change-email-screen-updates)

dannyalright:
🔥 nice! One small change: could we please use the `New email` string instead of `Email address`?

PR #485, opened 2019-02-19
Pull to refresh and paginator bug

eoji:
# What ❓
@dannyalright pointed out that when pull to refresh is performed in Discovery, the project list is cleared and then the new data is displayed. This was intended behavior because that screen uses an `ApiPaginator` and we purposely passed in `true` for `clearWhenStartingOver`. We also use an `ApiPaginator` in the Activity feed that doesn't clear so I thought it would be as simple as changing the boolean. This revealed a few bugs:
1. After refreshing a screen with `clearWhenStartingOver(false)`, when you continue to scroll, no more data is fetched. You can reproduce this by going to the activity feed, refreshing and trying to scroll to see more data.
2. The `RecyclerViewPaginator` suffered from back pressure crashes "due to the volume of scroll events". 

# How I attempted to fix it:
I realized we consumed way too many scroll events in the `RecyclerViewPaginator`, which was causing the back pressure issues. We really should only care about 2 things: if there is new data being fetched and if the user is scrolled to the bottom of the screen. If both of those things are true, we should ask the api for more data. The `RecyclerViewPaginator` now takes in an observable that emits whenever new data is being fetched so it's aware of the `ApiPaginator`'s state. Now every screen that uses a `RecyclerViewPaginator`, passes in the loading observable. These screens are: Activity feed, Comments, Message threads, Profile, Search and Discovery. 

Of course this broke almost all of the tests in `DiscoveryFragmentViewModelTest`. We previously assumed the projects would always be cleared when they were refreshed. I noticed we didn't have a centralized place where we got default search params based on a user, so I created `DiscoveryParams.getDefaultParams` that takes in a nullable user. Most notably, I noticed that we were clearing the activity tout prematurely. It previously was cleared when the search params observable emitted with no discernment about if the params were changing because the user was logging in. 

**Bonus**
Fixed the bug where the Account screen was crashing because we were not using `neverError` in the right place.

# Story 📖
[Trello](https://trello.com/c/X6U8Zsol/1224-stop-content-from-flashing-when-pulled-to-refresh)

# See 👀
device-2019-02-19-115855 2019-02-19 11_59_30

# Note
This means when switching params, the projects don't clear so there's a weird delay before new projects are shown since we don't show any progress indication.


eoji:
[Stop content from flashing when pulled-to-refresh](https://trello.com/c/X6U8Zsol/1224-stop-content-from-flashing-when-pulled-to-refresh)

app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
```diff
@@ -357,6 +374,10 @@ private void saveLastSeenActivityId(final @Nullable Activity activity) {
     @Override public @NonNull Observable<Void> startHeartAnimation() {
       return this.startHeartAnimation;
     }
+    @Override
```
Scollaco: Is there a reason why this Observable breaks a line?

app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
```diff
@@ -200,17 +211,22 @@ public ViewModel(final @NonNull Environment environment) {
...
 
-      this.currentUser.loggedInUser()
+      final Observable<Pair<User, DiscoveryParams>> loggedInUserAndParams = this.currentUser.loggedInUser()
         .distinctUntilChanged((u1, u2) -> !UserUtils.userHasChanged(u1, u2))
-        .compose(combineLatestPair(this.paramsFromActivity))
+        .compose(combineLatestPair(this.paramsFromActivity));
+
+      // Activity should show on the user's default params
+      loggedInUserAndParams
+        .filter(p -> p.second.toString().equals(DiscoveryParams.getDefaultParams(p.first).toString()))
```
Scollaco: Same here. Maybe `pair` instead of `p`?

app/src/main/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModel.java
```diff
@@ -200,17 +211,22 @@ public ViewModel(final @NonNull Environment environment) {
...
         .filter(this::activityHasNotBeenSeen)
         .doOnNext(this::saveLastSeenActivityId)
         .compose(bindToLifecycle())
         .subscribe(this.activity);
 
-      // Clear activity sample when params change
-      this.paramsFromActivity
+      // Clear activity sample when params change from default
+      loggedInUserAndParams
+        .filter(p -> !p.second.toString().equals(DiscoveryParams.getDefaultParams(p.first).toString()))
```
Scollaco: And here

PR #495, opened 2019-03-04
[UI Fix] - Ripple of profile section in logged in drawer layout goes outside of the bounds of the view.

Rcureton:
# What ❓
- changed the container attribute from background to foreground so the ripple stays within the view.

# How to QA? 🤔
- Log into the app `->` open the drawer `->` long press the profile row.

# Story 📖

[Ripple of profile section in logged in drawer layout goes outside of the bounds of the view.](https://trello.com/c/yYvRnJsn/1256-ripple-of-profile-section-in-logged-in-drawer-layout-goes-outside-of-the-bounds-of-the-view)

# See 👀
ripple

### DIFF
--- a/app/src/main/res/layout/discovery_drawer_logged_in_view.xml
+++ b/app/src/main/res/layout/discovery_drawer_logged_in_view.xml
@@ -13,10 +13,6 @@
     android:id="@+id/user_container"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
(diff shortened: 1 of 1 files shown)


eoji:
Looks like there's about 16dps of white space going around the ripple. I think the ripple should take up as much space as possible

Rcureton:
Ok cool I'll make the fix! Thanks

Rcureton:
@eoji  Here's the updated screenshot.

mar-04-2019 10-48-14

PR #497, opened 2019-03-05
[Maintenance] - Kotlin let feature

Rcureton:
# What ❓
- Decided to use a cool Kotlin trick to clean up some small things. So before we were previously checking if a value `!=null` and then executing what we want. With `let` we can clean up our code a little bit but with the same expected function and result. So instead of 

[Kotlin let run also apply with](https://www.journaldev.com/19467/kotlin-let-run-also-apply-with)

```
if (it != null) {
// something we want to do
}
```
We can now use this
```
item.?let {
// Something we want to do
}
```

# How to QA? 🤔
- For `ChangeEmailActivity` and `HelpSettingsActivity` they should both function the same way. So just type and check if the validation text shows and disappears. For the Help screen the chrome tabs should launch as expected.

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ChangeEmailActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ChangeEmailActivity.kt
@@ -95,12 +95,11 @@ class ChangeEmailActivity : BaseActivity<ChangeEmailViewModel.ViewModel>() {
                 .compose(bindToLifecycle())
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe {
-                    if (it != null) {
+                    it?.let {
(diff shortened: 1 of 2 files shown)


Rcureton:
Ok I understand better now, Thanks for the feedback!

selmanon:
It's possible to have the else with ?: operator.

eoji:
`it` in this case is the string resource id we should as the `TextView`'s `text`. We're using its nullability to determine the view's `visibility`, not for an assignment which is what I understand elvis to be used for. https://kotlinlang.org/docs/reference/null-safety.html#elvis-operator

But we could do `email_warning_text_view.text = it ?: ""` if `it` was a `CharSequence`

PR #498, opened 2019-03-11
updated the schema from kickstarter master

Rcureton:
# What ❓
- Updated `GraphQl` schema so that we can support Union Pay.


eoji:
I don't think PR should be labeled as `infrastructure`, it's just an update to a generated file so it should be `maintenance`.

PR #501, opened 2019-03-13
Added Toggle Feature Flag for enabling Horizontal Rewards

Rcureton:
# What ❓
- Added Toggle Feature Flag for enabling Horizontal Rewards
- Created an Extension file level function for a simple way to save the toggling preference for `enabling horizontal rewards`.

# How to QA? 🤔
- Best way to QA this is to add this block of code below in the `SettingsActivity` and when you toggle your preference in `InternalToolsActivity` it will hide and show the `EditProfile` row

```
val isEnabledReward = environment.enableHorizontalRewards().get()
    if (isEnabledReward) {
            edit_profile_row.visibility = View.VISIBLE
        }
```

# Story 📖
[Internal Tools Flag for Horizontal Scrolling](https://trello.com/c/kie1oV01/1278-internal-tools-flag-for-horizontal-scrolling)

# See 👀
Mar-13-2019 14-52-27


app/src/main/java/com/kickstarter/libs/qualifiers/HorizontalRewards.kt
```diff
@@ -0,0 +1,6 @@
+package com.kickstarter.libs.qualifiers
+
+import javax.inject.Qualifier
+
+@Qualifier
+annotation class HorizontalRewards
```
eoji: Let's rename this to `HorizontalRewardsPreference`

app/src/main/java/com/kickstarter/libs/FeatureKey.java
```diff
@@ -3,4 +3,5 @@
 public final class FeatureKey {
   private FeatureKey() {}
 
+  public static final String ENABLE_HORIZONTAL_REWARDS = "enable_horizontal_rewards";
```
eoji: A feature key is usually the name of a preference, some examples of other feature keys we've had are 
```  public static final String ANDROID_MESSAGES = "android_messages";```
```public static final String ANDROID_SURVEYS = "android_surveys";```

So I think this should have a more descriptive name like `internal_horizontal_rewards`.

File: app/src/main/java/com/kickstarter/libs/qualifiers/HorizontalRewards.kt (hunk 1 above)
Comment (Rcureton) [reply app/src/main/java/com/kickstarter/libs/qualifiers/HorizontalRewards.kt]: Fixed!

File: app/src/main/java/com/kickstarter/libs/FeatureKey.java (hunk 2 above)
Comment (Rcureton) [reply app/src/main/java/com/kickstarter/libs/FeatureKey.java]: Fixed!

File: app/src/main/java/com/kickstarter/ApplicationModule.java [hunk 3]
```diff
@@ -378,6 +382,14 @@ static BooleanPreferenceType provideAppRatingPreference(final @NonNull SharedPre
     return new BooleanPreference(sharedPreferences, SharedPreferenceKey.HAS_SEEN_APP_RATING);
   }
 
+  @Provides
+  @Singleton
+  @HorizontalRewardsPreference
+  @NonNull
+  static BooleanPreferenceType provideHorizontalRewards(final @NonNull SharedPreferences sharedPreferences) {
```
eoji: I think you missed one! Let's rename to `provideHorizontalRewardsPreference`

eoji:
[Internal Tools Flag for Horizontal Scrolling](https://trello.com/c/kie1oV01/1278-internal-tools-flag-for-horizontal-scrolling)

PR #513, opened 2019-04-15
Horizontal Rewards Design Sweep

Rcureton:
# What ❓
- Completed by @Scollaco and myself
- We added the `RewardItems` back into the rewards so you can now see them. 
- Also we fixed added the `RewardItemDecorator` class for the divider line between reward items. Previously we were inserting views in the `rewards_item_view.xml`.
- Updated strings from master.

# How to QA? 🤔
- When your feature flag is enabled and you go to view `Rewards` you will see them as expected compared to the final designs. 

# Story 📖
[Horizontal Rewards Design Sweep](https://trello.com/c/D54LycLf/1324-horizontal-rewards-design-sweep)

# See 👀


app/src/main/java/com/kickstarter/libs/utils/RewardItemDecorator.kt
```diff
@@ -0,0 +1,39 @@
...
+                val params = child.layoutParams as RecyclerView.LayoutParams
+
+                val dividerTop = child.bottom + params.bottomMargin
+                val dividerBottom = dividerTop + divider.intrinsicHeight
+                divider.setBounds(dividerLeft, dividerTop, dividerRight, dividerBottom)
+                divider.draw(canvas)
+            }
+        }
+    }
+}
```
eoji: nl

app/src/main/java/com/kickstarter/libs/utils/RewardItemDecorator.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.graphics.drawable.Drawable
+import android.view.View
+import androidx.recyclerview.widget.RecyclerView
+
+
+class RewardItemDecorator(private val divider: Drawable) : RecyclerView.ItemDecoration() {
+
+    override fun getItemOffsets(outRect: Rect, view: View, parent: RecyclerView, state: RecyclerView.State) {
+        val pos = parent.getChildAdapterPosition(view)
+        if (pos != 0 && pos != parent.layoutManager!!.itemCount - 1) {
```
eoji: let's try to never do `!!`
we could do `parent.layoutManager?.let?`

app/src/main/java/com/kickstarter/ui/activities/BackingActivity.java
```diff
@@ -83,6 +84,7 @@ public void onCreate(final @Nullable Bundle savedInstanceState) {
     this.rewardsItemRecyclerView.setAdapter(rewardsItemAdapter);
     final LinearLayoutManager layoutManager = new LinearLayoutManager(this);
     this.rewardsItemRecyclerView.setLayoutManager(layoutManager);
+    this.rewardsItemRecyclerView.addItemDecoration(new RewardItemDecorator(getDrawable(R.drawable.divider_grey_500_horizontal)));
```
eoji: the color of the divider is `ksr_grey_300` according to Abstract. also, instead of passing in this drawable, we could just draw a 1dp  line

app/src/main/java/com/kickstarter/ui/viewholders/HorizontalRewardViewHolder.kt
```diff
@@ -35,6 +40,12 @@ class HorizontalRewardViewHolder(private val view: View, val delegate: Delegate)
 
     init {
 
+        val rewardItemAdapter = RewardsItemAdapter()
+        val itemDivider: Drawable = context().getDrawable(R.drawable.divider_grey_500_horizontal)
```
eoji: there's a warning that this returns an optional

app/src/main/java/com/kickstarter/ui/viewholders/HorizontalRewardViewHolder.kt
```diff
@@ -35,6 +40,12 @@ class HorizontalRewardViewHolder(private val view: View, val delegate: Delegate)
...
+        view.horizontal_rewards_item_recycler_view.adapter = rewardItemAdapter
```
eoji: we can replace the 3 `view.horizontal_rewards_item_recycler_view.` with 
`val itemRecyclerView = view.horizontal_rewards_item_recycler_view.`

app/src/main/java/com/kickstarter/viewmodels/HorizontalRewardViewHolderViewModel.kt
```diff
@@ -165,6 +171,10 @@ interface HorizontalRewardViewHolderViewModel {
 
             this.minimumText = formattedMinimum
 
+            this.rewardsItemList = reward
+                    .map { r -> r.rewardsItems() }
```
eoji: `it`

app/src/main/res/layout/item_reward.xml
```diff
@@ -43,25 +43,46 @@
...
+          android:layout_width="match_parent"
+          android:layout_height="wrap_content"
+          android:layout_marginTop="@dimen/grid_3"
+          android:orientation="vertical">
+
+          <TextView
+            style="@style/RewardSectionTitle"
+            android:text="@string/rewards_info_includes" />
+
+          <androidx.recyclerview.widget.RecyclerView
```
eoji: When there's a really long list, it's almost impossible to scroll to the bottom of the reward. Nesting recyclerViews feels really off to me but here we are. Here's an example of what I mean:
device-2019-04-16-113028 2019-04-16 11_34_47

app/src/test/java/com/kickstarter/viewmodels/HorizontalRewardViewHolderViewModelTest.kt
```diff
@@ -249,6 +252,22 @@ class HorizontalRewardViewHolderViewModelTest: KSRobolectricTestCase() {
...
 
+    @Test
+    fun testRewardsItems() {
+        val project = ProjectFactory.project()
+        setUpEnvironment(environment())
+
+        // Items section should be hidden when there are no items.
+        this.vm.inputs.projectAndReward(project, RewardFactory.reward())
+        this.rewardsItemsAreGone.assertValue(true)
+        this.rewardsItemList.assertValue(emptyList())
```
eoji: why are we emitting an empty list vs not emitting at all because there are no items?

Scollaco: I think this bit of code was taken from the RewardViewModel. There we are using the coalesce function to return an empty list case `rewardsItems()` is null  🤷🏻‍♂️ I changed it so it won't emit at all if the the reward doesn't have items.

app/src/main/java/com/kickstarter/ui/activities/BackingActivity.java
Scollaco: Just for curiosity, what would be advantage of drawing the line instead of passing in the drawable? We are using the same divider in other places, wouldn't be better to have one source of truth? 🤔

eoji: but the truth is in the abstract 👀 the truth will set you free

eoji: also i suggest just drawing a line of that color or else you should make a new divider drawable of that color 😛

Scollaco: Oh, by source of truth I meant in the code. Like if we need to change the color of the dividers in the future, we would need to change it in one place - the drawable, instead of going to individual files and changing the color of the drawn lines. Does it make sense?
And yeah, I've created a `divider_grey_300_horizontal` 🙈

app/src/main/java/com/kickstarter/ui/activities/BackingActivity.java
```diff
@@ -83,6 +86,7 @@ public void onCreate(final @Nullable Bundle savedInstanceState) {
     this.rewardsItemRecyclerView.setAdapter(rewardsItemAdapter);
     final LinearLayoutManager layoutManager = new LinearLayoutManager(this);
     this.rewardsItemRecyclerView.setLayoutManager(layoutManager);
+    this.rewardsItemRecyclerView.addItemDecoration(new RewardItemDecorator(Objects.requireNonNull(getDrawable(R.drawable.divider_grey_300_horizontal))));
```
eoji: why are we adding this in the BackingActivity? 👀

app/src/main/java/com/kickstarter/ui/viewholders/HorizontalRewardViewHolder.kt
```diff
@@ -35,6 +40,14 @@ class HorizontalRewardViewHolder(private val view: View, val delegate: Delegate)
 
     init {
 
+        val rewardItemAdapter = RewardsItemAdapter()
```
eoji: let's move this to a private method?

app/src/main/java/com/kickstarter/viewmodels/HorizontalRewardViewHolderViewModel.kt
```diff
@@ -94,6 +99,7 @@ interface HorizontalRewardViewHolderViewModel {
         private val reward: Observable<Reward>
         private val rewardDescriptionIsGone: Observable<Boolean>
         private val rewardEndDateSectionIsGone: Observable<Boolean>
+        private val rewardsItemList: Observable<List<RewardsItem>>
```
eoji: alpha

app/src/main/java/com/kickstarter/viewmodels/HorizontalRewardViewHolderViewModel.kt
```diff
@@ -165,6 +171,10 @@ interface HorizontalRewardViewHolderViewModel {
 
             this.minimumText = formattedMinimum
 
+            this.rewardsItemList = reward
+                    .filter { it.rewardsItems() != null }
```
eoji: we can do `ObjectUtils.isNotNull`

app/src/main/res/layout/item_no_reward.xml
```diff
@@ -64,6 +64,7 @@
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:layout_gravity="bottom"
+    android:textAllCaps="false"
```
eoji: let's make a style for this!

app/src/main/res/values/dimens.xml
```diff
@@ -167,9 +168,9 @@
 
   <!-- Native Checkout -->
   <dimen name="fab_radius">@dimen/grid_5</dimen>
-  <dimen name="reward_margin">16dp</dimen>
-  <dimen name="card_margin">16dp</dimen>
-  <dimen name="card_padding">16dp</dimen>
+  <dimen name="reward_margin">18dp</dimen>
```
eoji: let's use `grid_3`

app/src/main/res/values/dimens.xml
```diff
@@ -47,6 +47,7 @@
   <dimen name="title_2">22sp</dimen>
   <dimen name="title_1">28sp</dimen>
   <dimen name="title_thank_you">36sp</dimen>
+  <dimen name="reward_title_text_size">24dp</dimen>
```
eoji: text sizes should be in `sp`

app/src/main/res/values/styles.xml
```diff
@@ -668,6 +668,26 @@
     <item name="cardTextErrorColor">@color/ksr_red_400</item>
   </style>
 
+  <!-- Native Checkout -->
```
eoji: let's move text styles to the text section

app/src/main/res/values/styles.xml
```diff
@@ -668,6 +668,26 @@
...
+
+  <style name="RewardTitle">
```
eoji: let's extend these from our existing styles

app/src/test/java/com/kickstarter/viewmodels/HorizontalRewardViewHolderViewModelTest.kt
```diff
@@ -25,6 +26,7 @@ class HorizontalRewardViewHolderViewModelTest: KSRobolectricTestCase() {
     private val reward = TestSubscriber<Reward>()
     private val rewardDescriptionIsGone = TestSubscriber<Boolean>()
     private val rewardEndDateSectionIsGone = TestSubscriber<Boolean>()
+    private val rewardsItemList = TestSubscriber<List<RewardsItem>>()
```
eoji: alpha

app/src/main/java/com/kickstarter/viewmodels/HorizontalRewardViewHolderViewModel.kt
eoji: let's rename to `rewardItems`

app/src/main/java/com/kickstarter/viewmodels/HorizontalRewardViewHolderViewModel.kt
```diff
@@ -94,6 +99,7 @@ interface HorizontalRewardViewHolderViewModel {
         private val reward: Observable<Reward>
         private val rewardDescriptionIsGone: Observable<Boolean>
         private val rewardEndDateSectionIsGone: Observable<Boolean>
+        private val rewardsItemList: Observable<List<RewardsItem>>
         private val rewardsItemsAreGone: Observable<Boolean>
```
eoji: let's rename to `rewardItemsAreGone`

eoji:
[Horizontal Rewards Design Sweep](https://trello.com/c/D54LycLf/1324-horizontal-rewards-design-sweep)

PR #514, opened 2019-04-23
Sign up or log in from Pledge screen

eoji:
# What ❓
Adds continue button to pledge fragment so users can log in or sign up during pledge flow.
Tests.

# How to QA? 🤔
As a logged out user, you should see the `Continue` button at the bottom of the pledge screen.
And when you click it, you will be taken to the Login Tout screen with the sign up and log in buttons.
After signing up or logging in, the `Continue` button should be gone and your payment methods should display.

As a logged in user, you should **not** see the `Continue` button at the bottom of the pledge screen and your payment methods should display.

# Story 📖
[Trello](https://trello.com/c/uWKbjbhp/1333-android-user-can-login-or-register-from-pledge-screen)

# See 👀


eoji:
[Android User Can Login or Register from Pledge Screen](https://trello.com/c/uWKbjbhp/1333-android-user-can-login-or-register-from-pledge-screen)

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -167,12 +200,18 @@ interface PledgeFragmentViewModel {
 
         override fun cards(): Observable<List<StoredCard>> = this.cards
```
Scollaco: Could you alphabetize here?

app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
```diff
@@ -83,13 +89,47 @@ class PledgeFragmentViewModelTest : KSRobolectricTestCase() {
         this.cards.assertValueCount(2)
     }
 
+    @Test
+    fun testContinueButtonIsGone() {
```
Scollaco: I wonder if this test is still relevant, since the `testLoggingInUser` method already tests the outputs in both (logged in / logged out) scenarios.

app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
```diff
@@ -83,13 +89,47 @@ class PledgeFragmentViewModelTest : KSRobolectricTestCase() {
...
+    }
+
     @Test
     fun testEstimatedDelivery() {
         setUpEnvironment(environment())
 
         this.estimatedDelivery.assertValue("March 2019")
     }
 
+    @Test
```
Scollaco: Same here. I think we could either only test the `paymentContainerIsGone` output, as the method name describes, or maybe rename the `testLoggingInUser` test to something more general and delete this test and the `testContinueButtonIsGone`.

app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
eoji: Here's what I am thinking; there are 3 scenarios:
1. The user opens the PledgeFragment logged in
2. The user opens the PledgeFragment logged out
3. The user opens the PledgeFragment logged out and then logs in

Scollaco: Oh, I see what you mean. So could you rename the tests to reflect what you want to test?

eoji: ༼ ༎ຶ ෴ ༎ຶ༽  but they are
1. `testPaymentContainerIsGone`
2. `testContinueButtonIsGone`
3. `testLoggingInUser` - maybe this one could be `testContinueButtonAndPaymentContainerAfterLogin`?

Scollaco: lol I know. What I'm saying is - If the test is called `testPaymentContainerIsGone`, we should preferably only test the `paymentContainerIsGone` output. The same would apply for the `testContinueButtonIsGone`.

Currently we are testing 2 outputs (`paymentContainerIsGone` and `continueButtonIsGone`) on the same test . This is fine, but in this case, we should rename the tests to explicitly say we are testing a flow that involves both outputs and not only one at a time. (But this is just a suggestion.)

PR #518, opened 2019-04-25
[🐞] Fixes bug with rewards container getting resized 

eoji:
# What ❓
I noticed an unsightly bug when the keyboard is dismissed while the `RewardsFragment` is up caused by the evil and deceptive `ConstraintLayout`. For example, this can happen when you quick reply to a notification.

# How to QA? 🤔
Confirm the `RewardsFragment` reward cards are always full height.

# Story 📖
issa bug i found

# See 👀
| Before 🐛 | After 🦋 |
| --- | --- |
| bug | fix |

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
@@ -86,7 +86,7 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>() {
         this.viewModel.outputs.setInitialRewardsContainerY()
                 .compose(bindToLifecycle())
                 .observeOn(AndroidSchedulers.mainThread())
-                .subscribe{ setInitialRewardsContainerY() }
+                .subscribe { setInitialRewardsContainerY() }
 
         this.viewModel.outputs.showRewardsFragment()
                 .compose(bindToLifecycle())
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
```diff
@@ -175,42 +183,47 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>() {
     }
 
     private fun animateRewards(expand: Boolean) {
-        this.showRewardsFragment.removeAllUpdateListeners()
-        this.showRewardsFragment.addUpdateListener { valueAnim ->
-            val initialRadius = resources.getDimensionPixelSize(R.dimen.fab_radius).toFloat()
-            val radius = initialRadius * if (expand) 1 - valueAnim.animatedFraction else valueAnim.animatedFraction
-            this.rewards_container.radius = radius
-        }
+        val showTarget = if (!expand) native_back_this_project_button else pledge_container
```
Scollaco: Here I would reserve names starting with verbs for actions. In this case, `showTarget` can makes us believe that it's a function instead of a target of type T. Maybe `targetToShow` is more appropriated? 🤔

app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
```diff
@@ -175,42 +183,47 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>() {
...
+        val showRewardsFragment: ObjectAnimator = ObjectAnimator.ofFloat(showTarget, View.ALPHA, 0f, 1f)
 
-        val durationTransition = AutoTransition()
-        durationTransition.duration = animDuration
-        TransitionManager.beginDelayedTransition(root, durationTransition)
+        val hideTarget = if (!expand) pledge_container else native_back_this_project_button
```
Scollaco: Same here

PR #528, opened 2019-05-24
KSCurrency refactor

eoji:
# What ❓
~First~ Third pass of fixing currency conversions and updating tests:

- I added a Model for `Country` based off https://github.com/kickstarter/ios-oss/blob/master/KsApi/models/Project.Country.swift.
- The method `formatWithProjectCurrency` was redundant because that's already what `format` does. The method `userCurrencyOptions` was also redundant because we can just use `currencyOptions`.
- Almost all of the tests were either incomplete or just testing the wrong thing.
- Most notably all of the tests for `formatWithUserPreference` weren't actually testing anything because the `fxRate` was never updated. 
- When testing conversions, the config `KSCurrency` uses needs to be explicitly set or it defaults to `server-config.json` (which has a user in the US). (Relevant to the `RewardViewHolderViewModel` tests).
- For some reason (it predates me) we were using `float` to represent currency values. iOS uses `double` so I switched to `double` so we can be wrong together.

# How to QA? 🤔
Change your currency preferences and verify the converted amounts are displayed properly.
Friendly reminder, we round ~UP~ HALF_UP (like normal rounding) for reward and checkout amounts and DOWN for pledged and goal amounts.


app/src/main/java/com/kickstarter/libs/KSCurrency.java
```diff
@@ -110,128 +91,88 @@ public String formatWithUserPreference(final double initialValue, final @NonNull
    * Build {@link CurrencyOptions} based on the project and whether we would prefer to show USD. Even if USD is preferred,
```
ifbarrera: Is this function description still accurate? 🤔

eoji: was it ever 👀 (😭)

eoji: i updated it but i think it could be better still

app/src/main/java/com/kickstarter/libs/KSCurrency.java
```diff
@@ -96,7 +76,8 @@ public String formatWithProjectCurrency(final double initialValue, final @NonNul
    */
   public String formatWithUserPreference(final double initialValue, final @NonNull Project project, final @NonNull RoundingMode roundingMode) {
 
-    final CurrencyOptions currencyOptions = userCurrencyOptions(initialValue, roundingMode, project);
+    final float unconvertedValue = roundingMode == RoundingMode.UP ? (float) Math.ceil(initialValue) : (float) Math.floor(initialValue);
```
Scollaco: To avoid repetition here, could you create a helper function that returns the rounded value? Something like
```
private float roundedValue(final @NonNull double initialValue, final @NonNull RoundingMode roundingMode) {
   return roundingMode == RoundingMode.UP ? (float) Math.ceil(initialValue) : (float) Math.floor(initialValue);
}
```
So you simply use:
```
final float unconvertedValue = this.roundedValue(initialValue, roundingMode);
```
(Forgive my Java skills 😛 )

PR #532, opened 2019-05-30
No reward -> PledgeFragment

eoji:
# What ❓
Selecting a no reward reward takes user to `PledgeFragment`.
Delivery info is hidden.
Tests.

# How to QA? 🤔
Back a project with no reward and verify the Pledge screen is shown and the estimated delivery is gone.

# Story 📖
We discussed but where's the card? I don't know 👀 

# See 👀
device-2019-05-30-113213 2019-05-30 11_32_39

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
+++ b/app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
@@ -20,6 +20,7 @@ import com.kickstarter.libs.BaseFragment
 import com.kickstarter.libs.FreezeLinearLayoutManager
 import com.kickstarter.libs.qualifiers.RequiresFragmentViewModel
 import com.kickstarter.libs.rx.transformers.Transformers.observeForUI
+import com.kickstarter.libs.utils.RewardUtils
 import com.kickstarter.libs.utils.ViewUtils
 import com.kickstarter.ui.ArgumentsKey
 import com.kickstarter.ui.activities.LoginToutActivity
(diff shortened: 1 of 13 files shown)


eoji:
[Delivery Information](https://trello.com/c/St3BnsEK/1384-delivery-information)

app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
```diff
@@ -186,7 +193,12 @@ class PledgeFragment : BaseFragment<PledgeFragmentViewModel.ViewModel>(), Reward
         reward_snapshot.pivotX = 0f
         reward_snapshot.pivotY = 0f
 
-        val rewardViewHolder = HorizontalRewardViewHolder(reward_to_copy, null)
+        val rewardViewHolder = when {
+            RewardUtils.isNoReward(reward) -> HorizontalNoRewardViewHolder(no_reward_layout, null)
+            else -> HorizontalRewardViewHolder(reward_layout, null)
+        }
+
+        rewardViewHolder.itemView.visibility = View.VISIBLE
```
Scollaco: Since we always have the No Reward view holder, wouldn't it be always visible? (Maybe we can set its visibility on the drawable?)

app/src/main/java/com/kickstarter/ui/viewholders/HorizontalNoRewardViewHolder.kt
```diff
@@ -4,23 +4,25 @@ import android.content.Intent
...
-class HorizontalNoRewardViewHolder(val view: View): KSViewHolder(view) {
+class HorizontalNoRewardViewHolder(val view: View, val delegate: HorizontalNoRewardViewHolder.Delegate?): KSViewHolder(view) {
 
-    private var viewModel = HorizontalNoRewardViewHolderViewModel.ViewModel(environment())
+    interface Delegate {
+        fun rewardClicked(screenLocation: ScreenLocation, reward: Reward)
+    }
 
-    private val projectBackButtonString = context().getString(R.string.project_back_button)
+    private var viewModel = HorizontalNoRewardViewHolderViewModel.ViewModel(environment())
```
Scollaco: Can the viewModel be a `val` instead?

app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
eoji: `rewardViewHolder.itemView` is either `no_reward_layout` or `reward_layout` and both of their initial visibility states are `INVISIBLE`

PR #533, opened 2019-05-31
[🔢] Unseen activity count

eoji:
# What ❓
- Displaying logged in user's unseen activity account in drawer with tests.
- Created reusable styles for drawer items with counts.
- Closing the drawer when navigating to activity feed.
- Updated schema to access `clearUserUnseenActivity` mutation. 
- Refreshing user after clearing unseen activity and tests.

# How to QA? 🤔
Well, if you have some unseen activity, the drawer hamburger should show and indicator and the Activity row in the drawer should show the count.

# Story 📖
[Trello Story](https://trello.com/c/DQ8NvTPx/1386-new-activity-indicator-inside-the-app)

# See 👀
I had the foresight to take a screenshot before I cleared it 😎

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
@@ -105,18 +105,18 @@ public interface ActivityFeedViewModel {
 
       this.startUpdateActivity = this.projectUpdateClick;
 
-      final Observable<Void> refreshSurvey = Observable.merge(this.refresh, this.resume).share();
+      final Observable<Void> refreshOrResume = Observable.merge(this.refresh, this.resume).share();
 
       final Observable<User> loggedInUser = this.currentUser.loggedInUser();
 
       loggedInUser
-        .compose(takeWhen(refreshSurvey))
+        .compose(takeWhen(refreshOrResume))
         .switchMap(__ -> this.apiClient.fetchUnansweredSurveys().compose(neverError()).share())
         .compose(bindToLifecycle())
         .subscribe(this.surveys);
 
       loggedInUser
-        .compose(takeWhen(refreshSurvey))
+        .compose(takeWhen(refreshOrResume))
         .map(User::unseenActivityCount)
         .map(IntegerUtils::intValueOrZero)
         .filter(IntegerUtils::isNonZero)
--- a/app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
@@ -5,11 +5,14 @@ import com.kickstarter.libs.ApiPaginator;
 import com.kickstarter.libs.CurrentUserType;
 import com.kickstarter.libs.Environment;
 import com.kickstarter.libs.KoalaContext.Update;
+import com.kickstarter.libs.utils.IntegerUtils;
 import com.kickstarter.libs.utils.ObjectUtils;
 import com.kickstarter.models.Activity;
 import com.kickstarter.models.Project;
 import com.kickstarter.models.SurveyResponse;
+import com.kickstarter.models.User;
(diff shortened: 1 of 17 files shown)


eoji:
[New activity indicator inside the app](https://trello.com/c/DQ8NvTPx/1386-new-activity-indicator-inside-the-app)

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -102,19 +107,32 @@ public ViewModel(final @NonNull Environment environment) {
...
+
+      loggedInUser
         .compose(takeWhen(refreshSurvey))
-        .switchMap(__ -> this.client.fetchUnansweredSurveys().compose(neverError()).share())
+        .switchMap(__ -> this.apiClient.fetchUnansweredSurveys().compose(neverError()).share())
         .compose(bindToLifecycle())
         .subscribe(this.surveys);
 
+      loggedInUser
+        .compose(takeWhen(refreshSurvey))
```
justinswart: Just curious why this is called `refreshSurvey?`

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
```diff
@@ -102,19 +107,32 @@ public ViewModel(final @NonNull Environment environment) {
...
+        .map(User::unseenActivityCount)
+        .map(IntegerUtils::intValueOrZero)
+        .filter(IntegerUtils::isNonZero)
+        .distinctUntilChanged()
+        .switchMap(__ -> this.apolloClient.clearUnseenActivity().compose(neverError()))
+        .switchMap(__ -> this.apiClient.fetchCurrentUser().compose(neverError()))
+        .compose(bindToLifecycle())
+        .subscribe(freshUser -> this.currentUser.refresh(freshUser));
```
justinswart: So just to confirm I am reading this correctly:

1. User refreshes feed.
2. Take the current user and inspect their `unseenActivityCount`.
3. If this is _not_ zero, clear the activity count and refresh the user.

I guess the only edge-case race condition is if, between the user logging in and looking at the feed, new activities come in. Would they be alerted to this by the OS via the push notification in some way? I'm also kind've wondering how to handle this on iOS 🤔

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
eoji: not sure,  I think a better name could be `resumeOrRefresh`

app/src/main/java/com/kickstarter/viewmodels/ActivityFeedViewModel.java
eoji: Yeah that's how it works. And hmm, they would get the push but no, we don't refresh the users on pushes. Should I maybe do a `combineLatest` like
```
refreshSurvey
. combineLatest(currentUser)
.filter(ObjectUtils::isNotNull)
//do the check and refresh
```

eoji: Ooh also I think the situation with log in would be.
They're logged out. 
They log in.
`loggedInUser` emits.
They return to that screen. `onResume` is fired (a lifecycle method).
So we'd have the updated user.

But that only handles log in, not getting pushes or new activity while on that screen.

justinswart: I think if a push comes in while the user is in the app we should refresh the count, but if the user is on the activities view while that happens then I guess we can ignore it, seeing as we'd have to immediately clear it again.

eoji: Spoke to @nnekab! She said for now, Android doesn't need to update the count after a push comes in because you can't actually see the number on that screen. It's only in the drawer in Discovery.

PR #534, opened 2019-06-05
Reward disabled state

Rcureton:
# What ❓
- Added logic to configure the disabled state of the pledge button in the `HorizontalRewardViewHolder`

# How to QA? 🤔
- Go to a project with rewards and when is a reward isn't available you will see the pledge button with the following text `No longer available` with a transparent green shade.

# Story 📖
[Reward disabled state](https://trello.com/c/0NMN6WCr/1393-reward-disabled-state)

# See 👀

disabled


app/src/main/java/com/kickstarter/ui/viewholders/HorizontalRewardViewHolder.kt
```diff
@@ -152,10 +160,15 @@ class HorizontalRewardViewHolder(private val view: View, val delegate: Delegate?
 
     private fun setMinimumText(@NonNull minimum: String) {
```
eoji: This is now a misnomer, so it should be renamed to something like `setButtonText`.

app/src/main/res/values/styles.xml
```diff
@@ -209,6 +209,8 @@
   <style name="PledgeButton">
     <item name="android:layout_gravity">bottom</item>
     <item name="android:textAllCaps">false</item>
+    <item name="android:textColor">@color/white</item>
+    <item name="android:backgroundTint">@drawable/ksr_button_state</item>
```
eoji: This a color. Since it's not used for all of our buttons it should be renamed to `button_reward_pledge`.

app/src/main/java/com/kickstarter/ui/viewholders/HorizontalRewardViewHolder.kt
```diff
@@ -59,7 +60,9 @@ class HorizontalRewardViewHolder(private val view: View, val delegate: Delegate?
         this.viewModel.outputs.isClickable()
                 .compose(bindToLifecycle())
                 .compose(observeForUI())
-                .subscribe { this.view.horizontal_reward_pledge_button.isClickable = it }
+                .subscribe {
```
eoji: single lines should be written as
`.subscribe { configureRewardButton(it) } `

app/src/main/res/drawable/ksr_button_state.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+
+  <item android:color="@color/green_alpha_42"
```
eoji: @dannyalright, is this the correct color for the disabled button states? It looks weird when the reward is long because it's transparent.

app/src/main/java/com/kickstarter/ui/viewholders/HorizontalRewardViewHolder.kt
Rcureton: The `rewardMinimum` text is also being set here that's why I didn't change it but if it doesn't matter I'll rename it.

eoji: There should be one method for setting the minimum text or just a single line in the subscribe block and a separate method for setting the button text.

app/src/main/res/drawable/ksr_button_state.xml
dannyalright: It's meant to be opaque—just a lighter shade of green. That's my bad for using opacity in the design mock ups. Seeing that we don't have a green tint in our library for this, could you please add and use #57bfa4? That's our core green (#009E74) at 66% opacity. Has as much contrast as possible whilst still looking disabled.

dannyalright:
@Rcureton did you get that color change in? Not sure if you saw my comment because I didn't @ you

eoji:
[Reward Disabled State](https://trello.com/c/0NMN6WCr/1393-reward-disabled-state)

eoji:
[Disabling sold out Rewards](https://trello.com/c/hqMCOBCG/1281-disabling-sold-out-rewards)

PR #517, opened 2019-04-25
User can view shipping countries from pledge screen

Rcureton:
# What ❓
- This was a great collaboration between the three of us! @Scollaco @eoji 🎉
- You can now view shipping countries from the pledge screen for rewards with shipping rules.
- This is part of the `native checkout` feature. 

# How to QA? 🤔
- [ ] Go to a project with the `native checkout feature flag` enabled -> select a reward -> on the pledge screen 
- [ ]  If a reward has shipping rules then you will see the EditText which you can start typing the country you want to ship to.
- [ ] If a reward has no shipping rules then the entire row will be gone.

# Story 📖
[User can view shipping countries from pledge screen](https://trello.com/c/ClAhqsC2/1271-user-can-view-shipping-countries-from-pledge-screen-android)

# See 👀
| With Shipping Rules | Without Shipping Rules |
| --- | --- |
|  |  |

Apr-25-2019 16-48-06

PR #535, opened 2019-06-06
[🔢] Currency refactor 2

eoji:
# What ❓
My fourth and hopefully final refactor of `KSCurrency`.

- Added `KSCurrency.currencyNeedsCode` to determine if a country's currency requires the country code and KSCurrency.getCurrencySymbol to return currency symbols for a country. These are helper methods to do the spanned currencies. Currently only shown in the `PledgeFragment`.
- By default, we round `DOWN`, so I created `KSCurrency.format` that only takes in a value and project.
- If we're rounding `DOWN`, we floor the number, or else we just leave it as is.
- There shouldn't be any platform dependent code in unit tests so `KSCurrency` never returns a `SpannableString`.
- Updated `KSCurrencyTest` to show how rounding works and added tests for `formatWithUserPreference` with new precision option.
- Added `StringUtils.trim` to remove trailing white space characters with tests.
- Added `ViewUtils.styleCurrency` to return stylized currency.
- Both examples in `ShippingRuleFactory` returned the same amount so there was no way to actually test that the number updated.
- Updated `*RewardViewModelTests` to show how rounding works.

# How to QA? 🤔
- Currency conversions should match the web.
- Only currencies with a country code should be spanned in `PledgeFragment`.

# Story 📖
This is a refactor that was necessary after #517 was merged without any tests for the currency code.

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/StringUtils.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/StringUtils.java
@@ -39,6 +39,13 @@ public final class StringUtils {
   }
 
   /**
+   * Returns a string with no leading or trailing whitespace.
+   */
+  public static String trim(final @NonNull String str) {
+    return str.replace('\u00A0', ' ').trim();
+  }
+
+  /**
    * Returns a string wrapped in parentheses.
    */
   public static @NonNull String wrapInParentheses(final @NonNull String str) {
--- a/app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
@@ -198,7 +198,7 @@ interface PledgeFragmentViewModel {
 
             rewardAmount
                     .compose<Pair<Double, Project>>(combineLatestPair(project))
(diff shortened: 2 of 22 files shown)


app/src/main/java/com/kickstarter/libs/utils/StringUtils.java
```diff
@@ -38,6 +38,13 @@ public static boolean isValidPassword(final @Nullable String str) {
       : str.substring(0, 1).toUpperCase(Locale.getDefault()) + str.substring(1).toLowerCase(Locale.getDefault());
   }
 
+  /**
+   * Returns a string with no leading or trailing whitespace.
+   */
+  public static String trim(final @NonNull String str) {
+    return str.replace('\u00A0', ' ').trim();
```
Rcureton: 👍🏾

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -69,10 +69,10 @@ interface PledgeFragmentViewModel {
         /**  Emits a boolean determining if the continue button should be hidden. */
         fun continueButtonIsGone(): Observable<Boolean>
 
-        /** Set the USD conversion.  */
+        /** Emits a string representing the total pledge amount in the user's preferred currency.  */
```
Rcureton: 👍🏾

PR #537, opened 2019-06-07
No reward shipping rules hidden

Scollaco:
# What ❓
- Hides Delivery Info with rewards has no shipping rules.
- Hides Shipping Rules section when user selects `No Reward`.
- Fixes a bug that didn't hide the `+` icon of the shipping amount when shipping rules were being loaded.

# How to QA? 🤔
- [x] Go to a project and select `No Reward`: the delivery info and shipping rules section should be hidden.
- [x] Go to a project and select `Reward with no shipping rules`: the delivery info and shipping rules section should be hidden.
- [x] Go to a project and select `Reward with shipping rules`: the delivery info and shipping rules section should be shown.

# Story 📖

[Hide Delivery Info if no shipping rules](https://trello.com/c/d96vOW9E/1398-hide-delivery-info-when-user-selects-reward-with-no-shipping-rules)

# See 👀
| No Reward | No Shipping Rules |
| --- | --- |
|  | 
| Shipping Rules |
|  |

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
@@ -319,56 +319,29 @@ interface PledgeFragmentViewModel {
                     .filter { ObjectUtils.isNull(it.first) || it.first.isEmpty() }
                     .compose<Pair<Pair<List<ShippingRule>, Reward>, Double>>(combineLatestPair(rewardAmountPlusAdditional))
 
-            val initialTotalAmountWithReward = rulesAndRewardAndAdditional
+            val initialAmountWithReward = rulesAndRewardAndAdditional
                     .map { it.second }
-                    .compose<Pair<Double, Project>>(combineLatestPair(project))
-                    .map<SpannableString> { this.ksCurrency.formatWithProjectCurrency(it.first, it.second, RoundingMode.UP, 2) }
-                    .compose(bindToLifecycle())
 
-            val initialTotalAmountNoReward = rewardAmountPlusAdditional
+            val initialAmountNoReward = rewardAmountPlusAdditional
                     .compose<Pair<Double, Reward>>(combineLatestPair(reward))
                     .filter { RewardUtils.isNoReward(it.second) }
                     .map { it.first }
-                    .compose<Pair<Double, Project>>(combineLatestPair(project))
-                    .map<SpannableString> { this.ksCurrency.formatWithProjectCurrency(it.first, it.second, RoundingMode.UP, 2) }
-                    .compose(bindToLifecycle())
 
-            val initialTotalAmount = Observable.merge(initialTotalAmountNoReward, initialTotalAmountWithReward)
+            val initialAmount = Observable.merge(initialAmountNoReward, initialAmountWithReward)
 
-            val totalWithShippingRule = rewardAmountPlusAdditional
+            val totalAmount = rewardAmountPlusAdditional
                     .compose<Pair<Double, Double>>(combineLatestPair(shippingAmount))
                     .map { it.first + it.second }
+
+            Observable.merge(initialAmount, totalAmount)
                     .compose<Pair<Double, Project>>(combineLatestPair(project))
                     .map<SpannableString> { this.ksCurrency.formatWithProjectCurrency(it.first, it.second, RoundingMode.UP, 2) }
-                    .compose(bindToLifecycle())
-
-            Observable.merge(initialTotalAmount, totalWithShippingRule)
                     .compose(bindToLifecycle())
                     .subscribe(this.totalAmount)
 
-            val initialTotalConversionAmountNoReward = rewardAmountPlusAdditional
-                    .compose<Pair<Double, Reward>>(combineLatestPair(reward))
-                    .filter { RewardUtils.isNoReward(it.second) }
-                    .map { it.first }
+            Observable.merge(initialAmount, totalAmount)
                     .compose<Pair<Double, Project>>(combineLatestPair(project))
                     .map { this.ksCurrency.formatWithUserPreference(it.first, it.second, RoundingMode.UP, 2) }
-                    .compose(bindToLifecycle())
-
-            val initialTotalConversionAmountWithReward = rulesAndRewardAndAdditional
-                    .map { it.second }
-                    .compose<Pair<Double, Project>>(combineLatestPair(project))
-                    .map { this.ksCurrency.formatWithUserPreference(it.first, it.second, RoundingMode.UP, 2) }
-                    .compose(bindToLifecycle())
-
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -215,20 +216,25 @@ interface PledgeFragmentViewModel {
                     .compose(bindToLifecycle())
                     .subscribe(this.shippingRulesAndProject)
 
-            rulesAndProject
-                    .compose(bindToLifecycle())
+            val hasNoShippingRules = rulesAndProject
                     .map { ObjectUtils.isNull(it.first) || it.first.isEmpty() }
+
+            val isNoReward = reward
+                    .filter { RewardUtils.isNoReward(it) }
```
eoji: You can just do 
`.map { RewardUtils.isNoReward(it) }`

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -313,12 +319,22 @@ interface PledgeFragmentViewModel {
...
                     .compose<Pair<Pair<List<ShippingRule>, Reward>, Double>>(combineLatestPair(rewardAmountPlusAdditional))
 
-            val initialTotalAmount = rulesAndRewardAndAdditional
+            val initialTotalAmountWithReward = rulesAndRewardAndAdditional
                     .map { it.second }
                     .compose<Pair<Double, Project>>(combineLatestPair(project))
                     .map<SpannableString> { this.ksCurrency.formatWithProjectCurrency(it.first, it.second, RoundingMode.UP, 2) }
                     .compose(bindToLifecycle())
 
+            val initialTotalAmountNoReward = rewardAmountPlusAdditional
```
eoji: Since this is coming up again; instead of calling ```.map<SpannableString> { this.ksCurrency.formatWithProjectCurrency(it.first, it.second, RoundingMode.UP, 2) }``` 3 times, you can do this:
```
            val initialTotalAmountWithReward = rulesAndRewardAndAdditional
                    .map { it.second }
                    .compose<Pair<Double, Project>>(combineLatestPair(project))

            val initialTotalAmountNoReward = rewardAmountPlusAdditional
                    .compose<Pair<Double, Reward>>(combineLatestPair(reward))
                    .filter { RewardUtils.isNoReward(it.second) }
                    .map { it.first }
                    .compose<Pair<Double, Project>>(combineLatestPair(project))

            val initialTotalAmount = Observable.merge(initialTotalAmountNoReward, initialTotalAmountWithReward)

            val totalWithShippingRule = rewardAmountPlusAdditional
                    .compose<Pair<Double, Double>>(combineLatestPair(shippingAmount))
                    .map { it.first + it.second }
                    .compose<Pair<Double, Project>>(combineLatestPair(project))

            Observable.merge(initialTotalAmount, totalWithShippingRule)
                    .map<SpannableString> { this.ksCurrency.formatWithProjectCurrency(it.first, it.second, RoundingMode.UP, 2) }
                    .compose(bindToLifecycle())
                    .subscribe(this.totalAmount)
```
`totalAmount` is what cares about the `SpannableString` so that's when we can map to get it. Also we only need to call `.compose(bindToLifecycle())` before subscribing.

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -330,12 +346,22 @@ interface PledgeFragmentViewModel {
                     .compose(bindToLifecycle())
                     .subscribe(this.totalAmount)
 
-            val initialTotalConversionAmount = rulesAndRewardAndAdditional
+            val initialTotalConversionAmountNoReward = rewardAmountPlusAdditional
```
eoji: Same thing here:

```
            val initialTotalConversionAmountWithReward = rulesAndRewardAndAdditional
                    .map { it.second }
                    .compose<Pair<Double, Project>>(combineLatestPair(project))

            val initialTotalConversionAmount = Observable.merge(initialTotalConversionAmountNoReward, initialTotalConversionAmountWithReward)

            val totalConversionAmount = rewardAmountPlusAdditional
                    .compose<Pair<Double, Double>>(combineLatestPair(shippingAmount))
                    .map { it.first + it.second }
                    .compose<Pair<Double, Project>>(combineLatestPair(project))

            Observable.merge(initialTotalConversionAmount, totalConversionAmount)
                    .map { this.ksCurrency.formatWithUserPreference(it.first, it.second, RoundingMode.UP, 2) }
                    .compose(bindToLifecycle())
                    .subscribe { this.conversionText.onNext(it) }
```

app/src/main/res/layout/fragment_pledge.xml
```diff
@@ -256,6 +256,13 @@
             android:layout_height="wrap_content"
             android:layout_weight=".1">
 
+            <LinearLayout
```
eoji: I'm confused. Why do we need this container? Are we not hiding the whole section?

Scollaco: The `+` icon is not being hidden while the shipping rules are loading, as a result, we can see icon close to the shipping_amount_loading_view`. So to solve this, I put the ImageView and TextView  in a `LinerLayout` to hide/show the icon and amount all at once.

eoji: ok i think i get it!

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
Scollaco: I added the `.filter { RewardUtils.isNoReward(it) }` here because we only want this to emit if it's a  `NoReward`. Without the filter, the `this.shippingRulesSectionIsGone` and `this.estimatedDeliveryInfoIsGone` will emit twice if the rewards `isReward`.
Here I'm trying to avoid a situation where `this.shippingRulesSectionIsGone` would emit `false` (because `RewardUtils.isNoReward(it)` would be false) and then emit `true` if the reward doesn't have shipping  rules.

eoji: but won't it emit the same values? then you can do `distinctUntilChange`

Scollaco: Not on the situation I mentioned above. It would emit `false` and then `true`, so the `distinctUntilChange` would not work.

Scollaco: I edited my comment to try to make it more clear. So in the situation I mentioned above, the `disctinctUntilChange` will not work, since the values are different.

eoji:
[Hide Delivery Info if no shipping rules](https://trello.com/c/d96vOW9E/1398-hide-delivery-info-if-no-shipping-rules)

PR #539, opened 2019-06-09
Fix unintended navigation on ProjectActivity upon config changes

mkano9:
# What ❓

- Fixing a navigation bug caused by BehaviorSubject re-emitting when the Activity re-subscribes after config changes. ViewModel survives config changes, so the BehaviorSubject still holds the old event. These navigation events are supposed to be single event and not a state.

# How to QA? 🤔

1. Go to Project page. 
2. Click on any buttons that trigger navigation event such as "Comments". 
3. Go back to Project page.
4. Rotate the device to trigger a config change.

The page automatically navigates to the Comments page instead of Project page. 

# See 👀


mkano9:
@eoji Awesome, thanks for the reviews!

PR #541, opened 2019-06-10
Literally native checkout 🙃

eoji:
# What ❓
Pledging with a `StoredCard`:
- Added mutation for `nativeCheckout`. 
  - It's kind of funky because `projectId` and `rewardId` are relay IDs while `locationId` and `paymentSourceId` are just strings.
  - It returns `Checkout.State` and I consider the call to have been successful if `state == VERIFYING`.
- We needed a 5th loading state `RewardLoadingCardViewHolder` for the stored cards in the `PledgeFragment.` Instead of `showPledgeCard` returning `Pair<Int, Boolean>`, I created an enum `CardState` and that has 3 values: `SELECT`, `PLEDGE`, `LOADING`. 
  - So the normal state is `SELECT`. If a user clicks `"Select"`, its new state will be `PLEDGE` and when they click `"Pledge"`, it'll be `LOADING`. If the pledge call is successful, the state doesn't really matter because we're going to destroy the `ProjectActivity`. But if it's unsuccessful, the state will go back to `PLEDGE` and `showPledgeError` will emit. So `showPledgeCard` now emits `Pair<Int, CardState>`. If successful, `startThanksActivity` emits.
  - (What should happen if a user hits back?)
- The shipping rules API call was missing `neverError()` so it crashed the app every time the call failed.
- UI: 
  - `shipping_amount_container` is `invisible` instead of `gone` so the screen doesn't awkwardly resize when it's visible.
  - Shifted the shipping amount to align with the total.
- Enforcing `RewardUtils.isShippable`  as the arbiter of determining whether we fetch shipping rules as iOS does. It also stops us from making the call unnecessarily.
- Simplified how we get the `total` output. The total is either with shipping or without. If a reward is `shippable`, the total won't emit unless it also has that value.
- Per Danny, we shouldn't load the stored cards until the total has been calculated.
- Tests for no reward, digital and physical rewards.
- Danny pointed out that digital rewards can have delivery dates.

# How to QA? 🤔
Back a project with a stored card 😛 

# Story 📖
[Trello Story](https://trello.com/c/bRloDmly/1363-user-can-back-a-project-using-a-stored-card)

# See 👀
device-2019-06-10-115121 2019-06-10 11_52_50 device-2019-06-10-115457 2019-06-10 11_55_29

For future consideration: 
- When the call is really fast, the user doesn't see my very nice animation of the progress bar.
- What should the error copy be? What's returned from the server is not user friendly.
- I will find out why the Thanks page is so slow and destroy it.

### DIFF
--- a/app/src/main/java/com/kickstarter/services/KSApolloClient.kt
+++ b/app/src/main/java/com/kickstarter/services/KSApolloClient.kt
@@ -1,5 +1,6 @@
 package com.kickstarter.services
 
+import CheckoutMutation
 import ClearUserUnseenActivityMutation
 import CreatePasswordMutation
 import DeletePaymentSourceMutation
@@ -17,18 +18,45 @@ import com.apollographql.apollo.api.Response
 import com.apollographql.apollo.exception.ApolloException
 import com.google.android.gms.common.util.Base64Utils
 import com.kickstarter.libs.utils.ObjectUtils
-import com.kickstarter.models.Project
-import com.kickstarter.models.Relay
-import com.kickstarter.models.StoredCard
-import com.kickstarter.models.User
+import com.kickstarter.models.*
 import rx.Observable
 import rx.subjects.PublishSubject
+import type.CheckoutState
 import type.CurrencyCode
 import type.PaymentTypes
 import java.nio.charset.Charset
 import kotlin.math.absoluteValue
 
 class KSApolloClient(val service: ApolloClient) : ApolloClientType {
+
+    override fun checkout(project: Project, amount: String, paymentSourceId: String, locationId: String?, reward: Reward?): Observable<Boolean> {
+        return Observable.defer {
+            val ps = PublishSubject.create<Boolean>()
+            service.mutate(CheckoutMutation.builder()
+                    .projectId(encodeRelayId(project))
+                    .amount(amount)
+                    .paymentSourceId(paymentSourceId)
+                    .locationId(locationId?.let { it })
+                    .rewardId(reward?.let { encodeRelayId(it) })
+                    .build())
+                    .enqueue(object : ApolloCall.Callback<CheckoutMutation.Data>() {
+                        override fun onFailure(exception: ApolloException) {
+                            ps.onError(exception)
+                        }
+
+                        override fun onResponse(response: Response<CheckoutMutation.Data>) {
+                            if (response.hasErrors()) {
+                                ps.onError(java.lang.Exception(response.errors().first().message()))
+                            }
+                            val state = response.data()?.nativeCheckout()?.checkout()?.state()
+                            val success = state == CheckoutState.VERIFYING
+                            ps.onNext(success)
+                            ps.onCompleted()
+                        }
+                    })
+            return@defer ps
+        }
+    }
 
     override fun clearUnseenActivity(): Observable<Long> {
         return Observable.defer {
--- a/app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
+++ b/app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
@@ -171,7 +171,7 @@ class PledgeFragment : BaseFragment<PledgeFragmentViewModel.ViewModel>(), Reward
                 .compose(observeForUI())
                 .subscribe {
                     shipping_amount.text = it
-                    ViewUtils.setGone(shipping_amount_container, false)
+                    ViewUtils.setInvisible(shipping_amount_container, false)
                     ViewUtils.setGone(shipping_amount_loading_view, true)
                 }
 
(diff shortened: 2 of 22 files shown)


eoji:
[User Can Back a Project using a Stored Card](https://trello.com/c/bRloDmly/1363-user-can-back-a-project-using-a-stored-card)

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
```diff
@@ -17,19 +18,46 @@ import com.apollographql.apollo.api.Response
...
+                        override fun onFailure(exception: ApolloException) {
+                            ps.onError(exception)
+                        }
+
+                        override fun onResponse(response: Response<CheckoutMutation.Data>) {
+                            if (response.hasErrors()) {
+                                ps.onError(java.lang.Exception(response.errors().first().message()))
+                            }
+                            val state = response.data()?.nativeCheckout()?.checkout()?.state()
+                            val success = state == CheckoutState.VERIFYING
```
Rcureton: For curiosity, why do you consider `VERIFYING` as successful instead of `SUCCESSFUL`?

app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
```diff
@@ -219,12 +237,14 @@ class PledgeFragment : BaseFragment<PledgeFragmentViewModel.ViewModel>(), Reward
         this.viewModel.inputs.newCardButtonClicked()
     }
 
+
```
Rcureton: NL

app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
```diff
@@ -443,17 +463,24 @@ class PledgeFragment : BaseFragment<PledgeFragmentViewModel.ViewModel>(), Reward
         pledge_details.y = pledge_root.height.toFloat()
     }
 
-    private fun updatePledgeCardSelection(positionAndSelected: Pair<Int, Boolean>) {
+    private fun updatePledgeCardSelection(positionAndSelected: Pair<Int, CardState>) {
```
Rcureton: Can you rename this to `positionAndCardState` according to the implementation `CardState` is more appropriate.

app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
```diff
@@ -163,8 +212,26 @@ class PledgeFragmentViewModelTest : KSRobolectricTestCase() {
...
+                .currentUser(MockCurrentUser(UserFactory.user()))
+                .build()
+        setUpEnvironment(environment)
+
+        this.cards.assertValueCount(1)
+        this.continueButtonIsGone.assertValue(true)
+        this.paymentContainerIsGone.assertValue(false)
+    }
+
+
```
Rcureton: NL

app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
```diff
@@ -374,18 +463,116 @@ class PledgeFragmentViewModelTest : KSRobolectricTestCase() {
...
-        this.totalAmount.assertValue("$20")
+        this.totalAmount.assertNoValues()
     }
 
     @Test
     fun testTotalAmountWithoutShippingRules_WhenStepperChangesValue() {
         val environment = environmentForShippingRules(ShippingRulesEnvelopeFactory.emptyShippingRules())
         setUpEnvironment(environment)
-        this.totalAmount.assertValue("$20")
+        this.totalAmount.assertNoValues()
```
Rcureton: Isn't the `totalAmount` supposed to emit even when we don't have shipping rules?

app/src/test/java/com/kickstarter/viewmodels/RewardLoadingCardViewHolderViewModelTest.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.kickstarter.mock.factories.StoredCardFactory
+import org.junit.Test
+import rx.observers.TestSubscriber
+import java.util.*
+
+class RewardLoadingCardViewHolderViewModelTest : KSRobolectricTestCase() {
+
+    private lateinit var vm: RewardLoadingCardViewHolderViewModel.ViewModel
+
+    private val estimatedDelivery = TestSubscriber.create<String>()
```
Rcureton: I think this should be `expirationDate` instead of `estimatedDelivery`

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
eoji: According to @frewsxcv,
> if a checkout gets returned from the mutation, it means the backend successfully recorded the checkout, and it'll be in a `verifying` state.

app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
eoji: Not if the shipping rules are empty on a project that is shippable.

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
frewsxcv: yep. `VERIFYING` means you have successfully sent the request to kickstarter, and we've recorded the checkout. once stripe gets back to us after authorizing the payment source, we'll change the state to `FAILED` or `SUCEESSFUL`

PR #542, opened 2019-06-10
Reward minimum span

eoji:
# What ❓
Rewards should only show the minimum amount with the mini currency code if it requires the country code.
- Added `ViewUtils.styleCurrency` that takes in a boolean to determine the style of shrunken currency code. 
  -  If we're just not centering it, we shrink the currency code to 70% (currency code is `17sp`  and the amount is `24sp`).
- Added `minimumAmountTitle` output for styled reward minimum.
- Added separate output for showing the unavailable button `limitReachedButtonTextIsVisible`.
- `HorizontalRewardViewHolderViewModel` is written in Kotlin so it can make use of the implicit parameter `it`.
- Tests.

# How to QA? 🤔
| Currency with country code | Currency without|
| --- | --- |
|  |  |

# Story 📖
[Trello](https://trello.com/c/Y8IWJpgZ/1403-reward-cell-amount-formatting-android)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/ViewUtils.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/ViewUtils.java
@@ -69,25 +69,7 @@ public final class ViewUtils {
    *
    */
   public static @NotNull SpannableString styleCurrency(final double value, final Project project, final @NonNull KSCurrency ksCurrency) {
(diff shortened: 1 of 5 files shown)


eoji:
[Reward cell amount formatting Android](https://trello.com/c/Y8IWJpgZ/1403-reward-cell-amount-formatting-android)

PR #543, opened 2019-06-12
⬆️Facebook SDK

eoji:
# What ❓
- Updated Facebook SDK to the dynamic version, per their recommendation https://developers.facebook.com/docs/android/getting-started/
- `LoginToutViewModelTest` only had tests for clicking log in and sign up so I added 2 more for Facebook success and errors but the VM needs some refactoring and more tests.
- The `fade_out` animation used to take .5 seconds, that's a long time. It's now .2 seconds.
- Using fade in/out transition in `LoginToutActivity`.

# How to QA? 🤔
- Log in or sign up with Facebook.

# Story 📖
[Trello Story](https://trello.com/c/uGFzObNM/1406-update-facebook-sdk)

# See 👀
This is what users who have already logged in with Facebook and have the Facebook app installed see:

device-2019-06-12-155132 2019-06-12 15_53_28

I don't love it but it's basically what we already have now.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -197,7 +197,6 @@ final butterknife_version = "7.0.1"
 final crashlytics_version = "2.9.6@aar"
 final dagger_version = "2.11"
 final exoplayer_version = "2.9.1"
-final facebook_version = "4.33.0"
 final firebase_version = "16.0.6"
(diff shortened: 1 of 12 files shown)


eoji:
[Update Facebook SDK](https://trello.com/c/uGFzObNM/1406-update-facebook-sdk)

PR #527, opened 2019-05-24
🍔✨ Hamburger animation

eoji:
# What ❓
Showing an indicator on the menu icon when a user has unread messages.

# How to QA? 🤔
1. Verify the unadorned menu icon is shown when logged out.
2. Verify the unadorned menu icon is shown when logged in and there are no unread messages.
3. Verify the adorned menu icon is shown and animates when logged in and there are unread messages.

# Story 📖
[Trello Story](https://trello.com/c/XxDhitMZ/1378-hamburger-animaton-when-new-activity-messages)

# See 👀

PR #546, opened 2019-06-18
[🐜] Fixes crash on API 23 for AnimatedVectorDrawables

eoji:
# What ❓
Showing the animated hamburger crashes on API 23 (not on every phone 🤕).

https://stackoverflow.com/questions/46261325/cannot-cast-to-animatedvectordrawablecompat-in-nougat
> Long answer:
> 
> I was having the same problem when I was trying to loop an animated vector drawable. Until I found out the support library returns different classes (AnimatedVectorDrawable and AnimatedVectorDrawableCompat) on different SDK levels.
> 
> It was not documented anywhere except this wonderful blog post of Nick Butcher:
> 
> https://medium.com/androiddevelopers/re-animation-7869722af206

Casting it to an `Animatable` fixes this issue.

# How to QA? 🤔
Log in and verify the hamburger animates if you have unread activity.
#527

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/DiscoveryActivity.java
+++ b/app/src/main/java/com/kickstarter/ui/activities/DiscoveryActivity.java
@@ -1,7 +1,7 @@
 package com.kickstarter.ui.activities;
 
 import android.content.Intent;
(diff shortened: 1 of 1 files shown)


eoji:
It's already released so will close the PR!

PR #563, opened 2019-07-04
Proper stepping!

eoji:
# 📲 What
Stepping by project country minimums and setting max based on country.

# 🤔 Why
So we step by sensible amounts in countries with a lower exchange rate.

# 🛠 How
- By pulling the `country` from the project's currency.

# 👀 See
| After 🦋 |
| --- |
| device-2019-07-03-205616 2019-07-03 20_56_57 |

# 📋 QA
Try to back any project, the step amount should be the project's country's `minPledge` amount.
https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/libs/models/Country.kt

# Story 📖
[Trello Story](https://trello.com/c/p316VtXB/1444-stepping-by-min-pledge-amount)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/RewardUtils.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/RewardUtils.java
@@ -20,8 +20,6 @@ import static com.kickstarter.libs.utils.BooleanUtils.isTrue;
 public final class RewardUtils {
   private RewardUtils() {}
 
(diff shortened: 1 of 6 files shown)


eoji:
[Make stepper increments match minimum pledge amount](https://trello.com/c/FF7Wdgmn/792-make-stepper-increments-match-minimum-pledge-amount)

PR #565, opened 2019-07-12
Cancel pledge

eoji:
# 📲 What
Users can attempt to cancel a live pledge and provide a note why.

# 🤔 Why
In case they change their mind, users should be able to cancel their pledge and optionally provide a note.
This matches the existing capabilities of checkout in the webview.

# 🛠 How
Changed `ProjectActivity`'s `windowSoftInputMode ` to `adjustPan` so the input in the cancel screen is visible.
Renamed `slide_in_bottom` animation to `slide_up`.
Added `slide_down` animation.
Updated graphQL schema.
Added `backing.locationId` so I don't need to fetch the backing in the `PledgeFragment`.
Added `cancelBacking` mutation and call to `ApolloClientType`.
`ProjectActivity` shows success snackbar when pledge is canceled.
Showing `CancelFragment` when user clicks cancel button in `PledgeFragment` with animation.
Added actual `PledgeFragment` name to transaction that adds it to the `fragment_container`.
Added dimens for `reward_card_radius` and `card_container_radius`.

# 👀 See
| Before 🐛 | After 🦋 |
| --- | --- |
| screenshot-2019-07-12_153956 | screenshot-2019-07-12_153909 |
| device-2019-07-18-140644 2019-07-18 14_15_06 |

# 📋 QA
0. Enable the `Native Checkout` feature flag in the Internal Tools.
1. Navigate to a live project you've backed. (The simplest way is to open your profile or just back a new project.
2. Click the `Manage` button.
3. Click the `Manage your pledge` button of your currently pledged reward.
4. Click the `Cancel pledge` button.
5. Click the `Yes, cancel my pledge`

**Note:** Right now the server cannot successfully cancel a pledge that was created on staging so the error message pops up.

# Story 📖
[Trello Story](https://trello.com/c/qY7q7T9B/1435-user-can-cancel-pledge)


eoji:
[User Can Cancel Pledge](https://trello.com/c/qY7q7T9B/1435-user-can-cancel-pledge)

app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.kt
```diff
@@ -158,10 +172,12 @@ interface ProjectViewModel {
         private val rewardsButtonColor = BehaviorSubject.create<Int>()
         private val rewardsButtonText = BehaviorSubject.create<Int>()
         private val setInitialRewardPosition = BehaviorSubject.create<Void>()
+        private val scrimIsVisible = BehaviorSubject.create<Boolean>()
```
Scollaco: Alphabetization is off here.

app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.kt
```diff
@@ -158,10 +172,12 @@ interface ProjectViewModel {
...
+        private val showCancelPledgeSuccess = PublishSubject.create<Void>()
         private val showRewardsFragment = BehaviorSubject.create<Boolean>()
-        private val startLoginToutActivity = PublishSubject.create<Void>()
         private val showShareSheet = PublishSubject.create<Project>()
         private val showSavedPrompt = PublishSubject.create<Void>()
+        private val startLoginToutActivity = PublishSubject.create<Void>()
```
Scollaco: And here

app/src/main/java/com/kickstarter/models/Backing.java
```diff
@@ -24,6 +24,7 @@
   public abstract @Nullable DateTime completedAt();
   public abstract long id();
   public abstract @Nullable Location location();
+  public abstract @Nullable Long locationId();
```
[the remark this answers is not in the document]
eoji: That's what I originally had but `backing.location.id` is null so I had to make a second call to fetch the backing. Using the `locationId` means I don't have to make the extra call woohoo!

PR #566, opened 2019-07-12
Dependency sweep

eoji:
# 📲 What
Updating out of date dependencies that don't require refactoring.

# 🤔 Why
It is ideal for dependencies to be up to date 🙂 These don't require any code changes. I have a list of dependencies that have breaking changes.

# 🛠 How
Updated out of date dependencies:
- Jetpack libraries
- Exoplayer
- Firebase
- Firebase messaging
- Jsoup
- Phoenix
- Play Services
- Material components
- Test libraries

# 👀 See
Nothing to visually see. App should behave the same, IF NOT BETTER.

# 📋 QA
Regression testing.

# Story 📖
👀


eoji:
I'M BACK BAYBEE

PR #571, opened 2019-07-25
[🐜🕷] Mark as received bug, View pledge bug

eoji:
# 📲 What
- This PR fixes the bug where the `Mark as received` switch was not visible when backers clicked the `View pledge` button (vs clicking the reward).
- Removing the `View Pledge` button in the `View pledge` screen when the user came from the creator bio.
- Fixes unsightly(!) overscroll in the `View pledge` screen.
- Fixes the crash when backers who try to start a message thread with the creator crashed when they clicked `View Pledge`.

# 🤔 Why
BUGS

# 🛠 How
- Previously~ I foolishly assumed that we'd only supply the `backer` to the `intent` to start the `BackingActivity` when a creator was viewing a backer's pledge not a backer viewing their own pledge... so now I'm just checking if the current user's `id` is the same as the backer's.
- Friendly reminder: If you're going to nest scrolling views, they need to be in a parent that supports nested scrolling or else they make scrolling janky (here I used `NestedScrollView`).
- Related to that, on the reward items list in the backing screen, I set `nestedScrollingEnabled` to `false` and `overscrollMode` to `never`.
- The `MessageViewModel` assumed the thread would never be empty so even though `View Pledge` was visible, the app crashed because we tried to access `messageThread.project` and the `messageThread` was null.
- Updated tests.

# 👀 See
Backer viewing their pledge by clicking `View Pledge` in the project footer

| Before 🐛 | After 🦋 |
| --- | --- |
| screenshot-2019-07-25_135859 | screenshot-2019-07-25_135906 |
| Ugly overscroll 🐛 | After 🦋|
| device-2019-07-25-140226 2019-07-25 14_03_08 | device-2019-07-25-140200 2019-07-25 14_04_30 |

# 📋 QA
1. There are 3 ways for a backer to get to the `View Pledge` screen where they should **always** see the mark as received switch:
  a. Clicking the `View Pledge` button at the footer of the project page
  b. Clicking their backed reward on the project page
  c. Clicking the `View Pledge` button from the message screen
2. There is 1 way for a creator to get to the `View Pledge` screen where they should **never** see the mark as received switch:
  a. Clicking the `View Pledge` button from the message screen (Dashboard -> Messages -> Select a message from a backer)

# Story 📖
[Trello Story](https://trello.com/c/OEkPWaY1/1485-bug-android-bug-report-no-longer-able-to-mark-pledges-as-received)

### DIFF
--- a/app/src/main/java/com/kickstarter/services/apiresponses/MessageThreadEnvelope.java
+++ b/app/src/main/java/com/kickstarter/services/apiresponses/MessageThreadEnvelope.java
@@ -16,7 +16,7 @@ import auto.parcel.AutoParcel;
 @AutoParcel
 public abstract class MessageThreadEnvelope implements Parcelable {
   public abstract @Nullable List<Message> messages();
-  public abstract MessageThread messageThread();
+  public abstract @Nullable MessageThread messageThread();
   public abstract List<User> participants();
 
   @AutoParcel.Builder
--- a/app/src/main/java/com/kickstarter/viewmodels/BackingViewModel.java
+++ b/app/src/main/java/com/kickstarter/viewmodels/BackingViewModel.java
@@ -137,8 +137,10 @@ public interface BackingViewModel {
       this.currentUser = environment.currentUser();
       this.ksCurrency = environment.ksCurrency();
 
+      final Observable<User> loggedInUser = this.currentUser.loggedInUser();
+
       final Observable<User> initialBacker = intent()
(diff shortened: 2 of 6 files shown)


eoji:
[[Bug] Android Bug Report - no longer able to mark pledges as received](https://trello.com/c/OEkPWaY1/1485-bug-android-bug-report-no-longer-able-to-mark-pledges-as-received)

PR #575, opened 2019-07-30
[🆕] Webview error state

eoji:
# 📲 What
Displaying a retry page for webviews that encounter an error (like no service).

# 🤔 Why
The current implementation crashes the app when the user tries to load a webview with no service and I want to be better.

# 🛠 How
- Updated strings 
- Previously, every `Activity` with a `Webview` implemented `KSWebViewClient.Delegate` when they might not have needed to.
- `KSWebView` is now a `FrameLayout` that contains a `Webview`, `ProgressBar` and retry layout. The `KSWebViewClient`'s `Delegate` is now this parent view. `KSWebView` also has its own `Delegate` if any screen is so inclined.

# 👀 See
| Before 🐛 | After 🦋 |
| --- | --- |
| screenshot-2019-07-30_112227 | screenshot-2019-07-30_110933 |
|  | device-2019-07-30-110209 2019-07-30 11_29_07 |

# 📋 QA
_Engineers only:_ You need to update the code in `KSWebViewClient` to run this branch.

1️⃣ Open a project page. Turn on Airplane Mode. Navigate to any of our lovely screens with webviews: Checkout, Project Updates, Campaign, Creator Bio, Help pages (if you don't have Chrome).

# Story 📖
[Trello Story](https://trello.com/c/KxMd34IB/1438-not-crashing-the-app-when-a-webview-network-call-fails)


eoji:
[Not crashing the app when a webview network call fails](https://trello.com/c/KxMd34IB/1438-not-crashing-the-app-when-a-webview-network-call-fails)

PR #577, opened 2019-08-01
Change method name 'param' to 'getID'

pdhung3012:
This class is used to manage User.  This method named 'param' is to get user ID. Thus, the method name 'getID' is more intuitive than 'param'.


eoji:
Hi @pdhung3012, thanks for opening the PR but I think that method is pretty explicit. If we just wanted the `id`, we'd use `id()` but `param` is specifying that it's returning the value to be used as a URL param.

pdhung3012:
thank you for your notification. Do you think that the method name 'getUrlParam' is the more appropriate name for this function?

eoji:
Hmm, I don't think so because I don't think we use `param` to mean something other than a URL param.

PR #578, opened 2019-08-02
[💳] One time card pledge

eoji:
# 📲 What
Adding non-reusable cards in native checkout.

# 🤔 Why
So users can add a card that's not stored.

# 🛠 How
- Updated schema.
- `NewCardActivity` now opens with the keyboard up.
- No longer checking for `errorMessage` and  `isSuccessful` in `SavePaymentMethod` mutation since the server now properly handles errors.
- `ApolloClientType.savePaymentMethod` now returns a `StoredCard`.
- `StoredCard.id` is now required so we don't need to do weird stuff when handling server responses.
- `NewCardFragment` can now be presented modally by passing in `ArgumentsKey. NEW_CARD_MODAL`. When it's modal, it:
  - uses the `modal_fragment_new_card.xml` layout 
  - errors are piped to `modalError` instead of `error`
  - the `AppBarLayout` does not have elevation
  - there is no divider in the form
  - the reusable switch container is visible
-  `NewCardFragment.OnCardSavedListener` now returns the successfully saved card.
- `ProjectActivity` now implements `NewCardFragment.OnCardSavedListener` and passes the result to the `PledgeFragment`.
- When the `Add new card` button is clicked in the `PledgeFragment`, `NewCardFragment` is presented modally.
- Created a style for modally presented fragments, `CardViewFragment`.
- Updated strings.
- Tests.

# 👀 See
| Before 🐛 | After 🦋 |
| --- | --- |
| device-2019-08-02-135929 2019-08-02 14_00_55 | device-2019-08-02-140003 2019-08-02 14_01_06 |
| device-2019-08-02-140738 2019-08-02 14_08_00 | device-2019-08-02-140547 2019-08-02 14_06_59 |

# 📋 QA
:one: Clicking the `Add New Card` button at the end of the stored cards list in the Pledge screen, shows the `Add New Card` screen modally with a switch to toggle storing the card.

:two: When a card is successfully saved, it appears in the Pledge screen with a green `Pledge` Button.

:three: User can pledge successfully with a payment source that is not stored.

:four: Regression test: Adding a new card from Settings -> Account -> Payment Methods

# Story 📖
[Trello Story](https://trello.com/c/cwojH2iC/1487-pledging-with-a-one-time-only-card)


eoji:
[Pledging with a one time only card](https://trello.com/c/cwojH2iC/1487-pledging-with-a-one-time-only-card)

app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
```diff
@@ -192,6 +191,14 @@ class PledgeFragment : BaseFragment<PledgeFragmentViewModel.ViewModel>(), Reward
                 .compose(observeForUI())
                 .subscribe { (cards_recycler.adapter as RewardCardAdapter).takeCards(it) }
 
+        this.viewModel.outputs.card()
```
Scollaco: Since the `PledgeFragment` is a fairly complex fragment, I would choose a more descriptive name for the output to make it easier to understand what is this output. (ie. `selectedCard`, `addedCard`...)

app/src/main/java/com/kickstarter/viewmodels/NewCardFragmentViewModel.kt
```diff
@@ -40,23 +44,35 @@ interface NewCardFragmentViewModel {
...
+
+        /** Emits a drawable to be shown based on when the card widget has focus. */
         fun cardWidgetFocusDrawable(): Observable<Int>
 
-        /** Emits when the password update was unsuccessful. */
-        fun error(): Observable<String>
+        /** Emits a boolean determining if the form divider should be visible. */
+        fun dividerIsVisible(): Observable<Boolean>
+
+        /** Emits when the password update was unsuccessful and the fragment is not modal. */
```
Scollaco: Copy pasta? 🤔

app/src/main/java/com/kickstarter/viewmodels/NewCardFragmentViewModel.kt
```diff
@@ -40,23 +44,35 @@ interface NewCardFragmentViewModel {
...
+        fun error(): Observable<Void>
+
+        /** Emits when the password update was unsuccessful and the fragment is modal. */
+        fun modalError(): Observable<Void>
```
Scollaco: And here.

app/src/main/java/com/kickstarter/viewmodels/NewCardFragmentViewModel.kt
```diff
@@ -111,34 +156,43 @@ interface NewCardFragmentViewModel {
...
 
             saveCardNotification
                     .compose(values())
-                    .subscribe {
-                        this.success.onNext(null)
-                        this.koala.trackSavedPaymentMethod()
-                    }
+                    .subscribe()
 
             saveCardNotification
```
Scollaco: Could you move this down close to the `this.koala.trackViewedAddNewCard()` line? This way we can keep the koala related code grouped.

app/src/main/java/com/kickstarter/viewmodels/NewCardFragmentViewModel.kt
```diff
@@ -161,53 +215,49 @@ interface NewCardFragmentViewModel {
...
 
-        override fun allowedCardWarningIsVisible(): Observable<Boolean> {
-            return this.allowedCardWarningIsVisible
-        }
+        override fun appBarLayoutHasElevation(): Observable<Boolean> = this.appBarLayoutHasElevation
 
-        override fun cardWidgetFocusDrawable(): Observable<Int> {
-            return this.cardWidgetFocusDrawable
-        }
+        override fun allowedCardWarningIsVisible(): Observable<Boolean> = this.allowedCardWarningIsVisible
```
Scollaco: Alpha 😬

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -174,6 +181,8 @@ interface PledgeFragmentViewModel {
 
     class ViewModel(@NonNull val environment: Environment) : FragmentViewModel<PledgeFragment>(environment), Inputs, Outputs {
 
+        private val cardSaved = PublishSubject.create<StoredCard>()
```
Scollaco: Alpha is off here too.

eoji:
Alright, I addressed your renaming and alpha feedback. (And then I broke my local build, it's irreparable, I ended up cloning the app again 😐)

To address your question about the switch is more long winded. This is an issue I keep running into with iOS folks so I'm going to be very explicit so I can just link back to it 😛

***Necessary background information:***
On Android, the keyboard does not know or care about what else is on the screen. It just knows that it's shown or not shown and you can tell it what type of input it should be configured for.

***Damn, that sucks. So what can we do?***
We can tell our screens how to behave when a keyboard is up. You specify this with the `windowSoftInputMode` in the `AndroidManifest.` To learn more about what the heck that is: https://developer.android.com/guide/topics/manifest/activity-element#wsoft.

And now, since we have that useful information.

***On this screen:***
We have our `windowSoftInputMode` set to `adjustResize` because we are educated!  Hence why the entire form is scrollable so you can always access everything in it. And why it looks like this on my Pixel 2, where I use a short keyboard:


When the keyboard is not up, it magically resizes!


**_TLDR; no_**

Scollaco:
Thanks for the educational moment 😄 Now I understand better why this happens and had forgotten that the keyboard on Android has its own life.

PR #582, opened 2019-08-06
[🍗] Select

eoji:
# 📲 What
Native checkout reward pledge button default CTA is `Select`

# 🤔 Why
> Button copy should not be dynamic

# 🛠 How
Native checkout reward pledge buttons now display `Select` as the live, unbacked CTA.
Renamed `RewardViewUtils.pledgeButtonAlternateText` to `pledgeButtonText` since `Select` is now the default value.
Removed `minimum` output from `NativeCheckoutRewardViewHolder` since it's now unused.
Updated tests.

# 👀 See

| Select 🍗 | 
| --- |
| screenshot-2019-08-06_152707 | 

# 📋 QA
1️⃣ Live, not backed -> `Select`
2️⃣ Live, backed -> `Manage your pledge`
3️⃣ Ended, backed -> `View your pledge`
4️⃣ Live, sold out/expired (not backed) -> `No longer available`
5️⃣ Live, not backed -> `Select this instead`
6️⃣ Ended, not backed -> Button is gone

# Story 📖
https://dripsprint.atlassian.net/browse/NT-43

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/NativeCheckoutRewardViewHolderViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/NativeCheckoutRewardViewHolderViewModel.kt
@@ -27,8 +27,8 @@ interface NativeCheckoutRewardViewHolderViewModel {
     }
 
     interface Outputs {
-        /**  Emits the string resource ID to set the pledge button when not displaying the minimum. */
-        fun alternatePledgeButtonText(): Observable<Int>
+        /**  Emits the string resource ID to set on the pledge button. */
+        fun buttonCTA(): Observable<Int>
 
         /** Emits `true` if pledge button can be clicked, `false` otherwise.  */
         fun buttonIsEnabled(): Observable<Boolean>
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/viewmodels/NativeCheckoutRewardViewHolderViewModel.kt
```diff
@@ -27,8 +27,8 @@ interface NativeCheckoutRewardViewHolderViewModel {
     }
 
     interface Outputs {
-        /**  Emits the string resource ID to set the pledge button when not displaying the minimum. */
-        fun alternatePledgeButtonText(): Observable<Int>
+        /**  Emits the string resource ID to set on the pledge button. */
+        fun buttonCTA(): Observable<Int>
```
Scollaco: Could this be renamed to buttonCTAText instead? So we know what the output is emitting.

eoji: but it's not text 👀

PR #585, opened 2019-08-08
[🖲] Clickable reward

eoji:
# 📲 What
Adding some a11y for rewards. 

# 🤔 Why
So the cards can be navigated reasonably when TalkBack™ is on.

# 🛠 How
Added `focusable` `ViewGroup`s for each section of the reward card.

# 👀 See
Nothing 2 see here. This one's for the visually impaired!

# 📋 QA
0️⃣ Turn on TalkBack in the settings
1️⃣ The reward minimum, conversion, and title should be read as one section. 
2️⃣ The `Includes` and list are their own focusable item. Lists handle their own voice over for individual row items.
3️⃣ The description and reward limits are read as one section.

# Story 📖
https://dripsprint.atlassian.net/browse/NT-46


app/src/main/java/com/kickstarter/ui/viewholders/NativeCheckoutRewardViewHolder.kt
```diff
@@ -169,12 +169,14 @@ class NativeCheckoutRewardViewHolder(private val view: View, val delegate: Deleg
                 .compose(observeForUI())
                 .subscribe { this.view.reward_check.setBackgroundResource(it) }
 
-        view.reward_pledge_button.setOnClickListener {
+        this.view.reward_pledge_button.setOnClickListener {
```
roblum: Cool! TIL in Kotlin you can set these properties in the constructor via `private val view: View` or `val view: View`

app/src/main/res/layout/item_reward.xml
```diff
@@ -153,6 +170,7 @@
           android:layout_width="match_parent"
           android:layout_height="wrap_content"
           android:layout_marginTop="@dimen/grid_3"
+          android:focusable="false"
```
roblum: Curious why we're setting this focusable to false!

app/src/main/java/com/kickstarter/ui/viewholders/NativeCheckoutRewardViewHolder.kt
```diff
@@ -169,12 +169,14 @@ class NativeCheckoutRewardViewHolder(private val view: View, val delegate: Deleg
...
             this.viewModel.inputs.rewardClicked()
         }
+
     }
 
     override fun bindData(data: Any?) {
+        @Suppress("UNCHECKED_CAST")
```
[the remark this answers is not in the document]
eoji: yep!

app/src/main/java/com/kickstarter/ui/viewholders/NativeCheckoutRewardViewHolder.kt
eoji: Yeah! I'm actually wondering if I should use `RxView.clicks` here to get the click events as an `Observable` 🤔

app/src/main/res/layout/item_reward.xml
eoji: This is the placeholder invisible button that I use to give the bottom of the reward card the right padding!

app/src/main/java/com/kickstarter/ui/viewholders/NativeCheckoutRewardViewHolder.kt
roblum: What would be the main benefit to switching to `RxView.clicks` rather than the current callback?

eoji: Nothing in this case because I'm not doing anything special with the click event but it looks cool 😎 
It'd look like this 
```
RxView.clicks(this.view.reward_pledge_button)
      .compose(bindToLifecycle())
      .subscribe { this.viewModel.inputs.rewardClicked() }
```

roblum: Sounds very cool, I say go for it!

PR #587, opened 2019-08-12
[🔜] Prelaunch project MVP

eoji:
# 📲 What
Prelaunch projects open in the browser 🌍 

# 🤔 Why
So users are taken to the prelaunch webpage when it's #activated.

# 🛠 How
- One does not simply _open a URL in the browser_, especially if that link is identical to a deep link it supports because then it just creates a very sad feedback loop. So, I had to use the same logic in the `DeepLinkActivity`; by creating an `Intent` chooser to present the user with options to open the link, not including the app they're already in!
- When the API returns a project where `prelaunchActivated` is `true`, the project opens in the browser.
- When the API returns a project where `prelaunchActivated` is `false`, the project opens natively.
- Tests ❤️ 

# 👀 See
| Pretend this project is prelaunchActivated~ | 
| --- |
| device-2019-08-12-145520 2019-08-12 14_56_15 | 

# 📋 QA
🔙 Regression test:  Projects that are not prelaunch activated should deep link normally.
🔜 Prelaunch projects should present the user with options of how to open the link. If you only have one browser installed, it'll just open it.

# Story 📖
https://dripsprint.atlassian.net/browse/NT-97

cc @kickstarter/procomm

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
@@ -377,7 +377,28 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
     }
 
     private fun openProjectAndFinish(url: String) {
-        startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(url)))
+        val uri = Uri.parse(url)
+
+        val fakeUri = Uri.parse("http://www.kickstarter.com")
+        val browserIntent = Intent(Intent.ACTION_VIEW, fakeUri)
+        val queryIntentActivities = packageManager.queryIntentActivities(browserIntent, 0)
+        val targetIntents = queryIntentActivities
+                .filter { !it.activityInfo.packageName.contains("com.kickstarter") }
+                .map { resolveInfo ->
+                    val intent = Intent(Intent.ACTION_VIEW, uri)
+                    intent.setPackage(resolveInfo.activityInfo.packageName)
+                    intent
+                }
+                .toMutableList()
+
+        if (targetIntents.isNotEmpty()) {
+            /* We need to remove the first intent so it's not duplicated
+            when we add the EXTRA_INITIAL_INTENTS intents. */
+            val chooserIntent = Intent.createChooser(targetIntents.removeAt(0), "")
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
```diff
@@ -370,6 +376,32 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
         }
     }
 
+    private fun openProjectAndFinish(url: String) {
```
justinswart: Hey just wanna understand this function - is this a test? It looks like it isn't so I'm just wondering how we use the hard-coded `fakeUri` here?

eoji: > One does not simply open a URL in the browser, especially if that link is identical to a deep link it supports because then it just creates a very sad feedback loop. So, I had to use the same logic in the DeepLinkActivity; by creating an Intent chooser to present the user with options to open the link, not including the app they're already in!

It's not a test. In Android, you can't say "open this URL" if it's a deep link URL. It just tries opening the app again. 

This method uses a `fakeUri` to see which apps can open it. For example, for `http://www.kickstarter.com/` can be opened by Chrome, Kickstarter, and Firefox (I don't have that installed but pretend.) Then I filter out Kickstarter because we're already here. And present the options. 

Let me know if that makes sense. Happy to rename!

eoji: And for emails: I see now that they're broken for deep linking 🙄

justinswart: Is hard-coding the URI a problem?

eoji: It's just an arbitrary URI, I could use `http://google.com`. The goal is to see what apps can handle a URI of that structure.

justinswart: ah gotcha

eoji:
I spoke to @darellkoh today! And she said one part of the API work is done, another is in progress, and she'll let me know when and how we can test it when they're ready!

jamielynnroth:
@justinswart @eoji the backend work to support this has been merged so you can test this fully and get it merged today

PR #594, opened 2019-08-21
[🔁] Cancel Pledge redesign

eoji:
# 📲 What
Implementation of the cancel pledge redesign.

# 🤔 Why
Redesign™

# 🛠 How
- Changed background color to `ksr_grey_300`.
- Removed toolbar.
- The pledge amount and project name are now **bold**.
- There were some `BehaviorSubject`s that should've been `PublishSubject`s, you know how it goes.
- It now animates horizontally, instead of vertically.

# 👀 See
| After 🦋 | Before 🐛| 
| --- | --- |
| screenshot-2019-08-21_114404 | screenshot-2019-07-12_153909 |
| device-2019-08-21-114534 2019-08-21 11_46_02 | cancelpledge |

# 📋 QA
Notes 2 u, the QA individual: 
1️⃣ Tertiary (gray) buttons are `ksr_grey_400` and not `ksr_grey_500` so it does not match the Abstract.
2️⃣ [Outlined text fields](https://material.io/design/components/text-fields.html#outlined-text-field) do not have a fill. We can switch to the [filled text fields](https://material.io/design/components/text-fields.html#filled-text-field) if desired.
3️⃣ Regression test: you should still be able to cancel your pledge or dismiss the screen.

# Story 📖
https://dripsprint.atlassian.net/browse/NT-200

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/fragments/CancelPledgeFragment.kt
+++ b/app/src/main/java/com/kickstarter/ui/fragments/CancelPledgeFragment.kt
@@ -1,14 +1,18 @@
 package com.kickstarter.ui.fragments
 
+import android.graphics.Typeface
 import android.os.Bundle
+import android.text.Spannable
+import android.text.SpannableString
+import android.text.style.StyleSpan
 import android.util.Pair
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.FragmentManager
-import com.google.android.material.snackbar.Snackbar
 import com.kickstarter.KSApplication
 import com.kickstarter.R
+import com.kickstarter.extensions.snackbar
 import com.kickstarter.extensions.text
 import com.kickstarter.libs.BaseFragment
 import com.kickstarter.libs.qualifiers.RequiresFragmentViewModel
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/ui/fragments/CancelPledgeFragment.kt
```diff
@@ -42,12 +46,12 @@ class CancelPledgeFragment : BaseFragment<CancelPledgeViewModel.ViewModel>() {
         this.viewModel.outputs.showServerError()
                 .compose(bindToLifecycle())
                 .compose(Transformers.observeForUI())
-                .subscribe { Snackbar.make(snackbar_anchor, R.string.Something_went_wrong_please_try_again, Snackbar.LENGTH_SHORT).show() }
+                .subscribe { snackbar(cancel_pledge_root, getString(R.string.Something_went_wrong_please_try_again)).show() }
```
oliverswitzer: 🍿

<3 functions as first class citizens in Kotlin! 🎆

app/src/main/java/com/kickstarter/ui/fragments/CancelPledgeFragment.kt
```diff
@@ -90,8 +95,15 @@ class CancelPledgeFragment : BaseFragment<CancelPledgeViewModel.ViewModel>() {
         val ksString = (activity?.applicationContext as KSApplication).component().environment().ksString()
         val amount = amountAndProjectName.first
         val name = amountAndProjectName.second
-        cancel_prompt.text = ksString.format(getString(R.string.Are_you_sure_you_wish_to_cancel_your_amount_pledge_to_project_name),
+        val formattedString = ksString.format(getString(R.string.Are_you_sure_you_wish_to_cancel_your_amount_pledge_to_project_name),
                 "amount", amount, "project_name", name)
+
+        val spannableString = SpannableString(formattedString)
```
oliverswitzer: Ah ok so at this point formatted string already has the data interpolated, and using `SpannableString` is really just to add bold-ness to the string for the amount and name of the project, right?

I'm getting it!

oliverswitzer: ...I think

eoji: yes!

eoji:
Hey @oliverswitzer, np!! 

So, a `PublishSubject` is a special type of `Observable` that when it's subscribed to, it will only emit new emissions.

Compared to a `BehaviorSubject`. A `BehaviorSubject` is a special type of `Observable` that when it's subscribed to, it will emit its last emission **and** any new emissions.

So you want to use a `BehaviorSubject` when you want the VM to maintain that state (say like when you're rotating your phone!) and a `PublishSubject` when that state is ephemeral. Using the `CancelPledgeFragmentViewModel` as an example, if we were showing a `ProgressBar`, we still want to show it after a rotation. For an error, we just want to show it once, and not repeat it if the user rotated their phone.

PR #595, opened 2019-08-21
[🍡] View/manage pledge menu

eoji:
# 📲 What
 (¯\`'•.¸(¯\`'•.¸(¯`'•Another one•'´¯)¸.•'´¯)¸.•'´¯) 
Special menu for view your pledge (when project is not live).

# 🤔 Why
You can't manage a pledge when the project has ended*. 
*(not true if your payment method errors but for the purposes of this PR, pretend ✨).

# 🛠 How
- Added `manage_pledge_ended` menu and renamed `manage_pledge` to `manage_pledge_live`.
- `managePledgeMenuIsVisible` is now `managePledgeMenu` which emits an optional `Int` that is either a menu to display, or if it's `null`, the menu is cleared.
- Tests, my fave.

# 👀 See
| Live backed project 🌞 | Ended backed project 🌚|
| - | - |
| screenshot-2019-08-21_143342 | screenshot-2019-08-21_143542 |
| Managing ☁️ | Unbacked project 🤔  |
| screenshot-2019-08-21_143917 | screenshot-2019-08-21_143553 |

# 📋 QA
🌞 Live backed projects display a menu with: `Update pledge`, `Change payment method`, `View rewards`, `Cancel pledge` and `Contact creator`.
🌚 Ended backed projects display a menu with: `View rewards` and `Contact creator`. 
☁️  When you leave the interstitial screen, the menu hides.
🤔If you haven't backed the project, there's no interstitial and therefore, no menu.

# Story 📖
https://dripsprint.atlassian.net/browse/NT-151


app/src/test/java/com/kickstarter/viewmodels/ProjectViewModelTest.kt
```diff
@@ -510,13 +510,27 @@ class ProjectViewModelTest : KSRobolectricTestCase() {
     }
 
     @Test
-    fun testManagePledgeMenuIsVisible_whenProjectBacked() {
+    fun testManagePledgeMenuIsVisible_whenProjectBackedAndLive() {
```
eoji: plz rename these tests

app/src/test/java/com/kickstarter/viewmodels/ProjectViewModelTest.kt
```diff
@@ -510,40 +510,54 @@ class ProjectViewModelTest : KSRobolectricTestCase() {
...
+        this.managePledgeMenu.assertValue(R.menu.manage_pledge_live)
 
         this.vm.inputs.cancelPledgeClicked()
         this.vm.inputs.fragmentStackCount(1)
-        this.managePledgeMenuIsVisible.assertValues(true, false)
+        this.managePledgeMenu.assertValues(R.menu.manage_pledge_live, null)
 
         this.vm.inputs.fragmentStackCount(0)
-        this.managePledgeMenuIsVisible.assertValues(true, false, true)
+        this.managePledgeMenu.assertValues(R.menu.manage_pledge_live, null, R.menu.manage_pledge_live)
```
Setsun: These tests overall look good! Just for my understanding, is this saying that these function was called with all three of these inputs at once, or in this order? Seeing the type is:

```
fun managePledgeMenu(): Observable<Int?>
```

Out of curiosity what testing library is used for the assertions?

app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
```diff
@@ -623,7 +624,11 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
         project_action_button.layoutParams = buttonParams
     }
 
-    private fun toggleManagePledgeVisibility(visible: Boolean) {
-        if (visible) rewards_toolbar.inflateMenu(R.menu.manage_pledge) else rewards_toolbar.menu.clear()
+    private fun updateManagePledgeMenu(@MenuRes menu: Int?) {
```
Setsun: Could we also check the presence of the menu at the callsite of `updateManagePledgeMenu`, and pass down a `Boolean`? 

```
fun updateManagePledgeMenu(menuVisible: Boolean): void

this.updateManagePledgeMenu(menu?)
```

Either way works though!

app/src/test/java/com/kickstarter/viewmodels/ProjectViewModelTest.kt
eoji: In this order! That's the stream of emissions.

We use http://robolectric.org/ to test.

app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
eoji: The menu always exists! https://developer.android.com/reference/android/widget/Toolbar.html#getMenu()

PR #607, opened 2019-09-06
[ℹ️] Sending the cookie ref in the createBacking mutation

eoji:
# 📲 What
What it says on the tin.

# 🤔 Why
So we credit the referrer for the pledge.

# 🛠 How
I use `RefTagUtils.storedCookieRefTagForProject` to get that sweet initially set cookie ref tag 🍪 . This value can be `null` and it would be `null` if you deep linked into that project without a `ref` query string.

# 👀 See
Nothing 2 c.

# 📋 QA
@cj-dimaggio and @Ru-T are figuring out how to test this but I verified I am sending it on my end 🙏 

# Story 📖
https://dripsprint.atlassian.net/browse/NT-157


app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
```diff
@@ -1514,7 +1517,9 @@ class PledgeFragmentViewModelTest : KSRobolectricTestCase() {
         val project = ProjectFactory.project()
         val environment = environment().toBuilder()
                 .apolloClient(object : MockApolloClient() {
-                    override fun createBacking(project: Project, amount: String, paymentSourceId: String, locationId: String?, reward: Reward?): Observable<Boolean> {
+                    override fun createBacking(project: Project, amount: String,
```
Setsun: Would there be value in adding a test for checking that the expected refTag is passed in?

app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
```diff
@@ -987,6 +993,36 @@ class PledgeFragmentViewModelTest : KSRobolectricTestCase() {
...
+        val environment = environment()
+                .toBuilder()
+                .cookieManager(cookieManager)
+                .sharedPreferences(sharedPreferences)
+                .apolloClient(object : MockApolloClient() {
+                    override fun createBacking(project: Project, amount: String,
+                                               paymentSourceId: String, locationId: String?,
+                                               reward: Reward?, refTag: RefTag?): Observable<Boolean> {
+                        //Assert that stored cookie is passed in
+                        TestCase.assertEquals(refTag, RefTag.discovery())
```
Setsun: 👍

PR #628, opened 2019-10-03
[🆕] NT-399 Adding new Koala properties

eoji:
# 📲 What
Adding some new Koala properties for ✨ tracking ✨ 

# 🤔 Why
FOR TRACKING

# 🛠 How
- Added `is_vo_on` to default event properties. The value is defined as if TalkBack™ is on *not* Select to Speak™ 
- Added `payment_method` to default project properties. The value is defined as values from the `CreditCardPaymentType` enum (`ANDROID_PAY, APPLE_PAY, BANK_ACCOUNT, CREDIT_CARD`) which is what's returned in `backing.paymentSource.paymentType`.
- Added `pledge_total`  to default project properties. The value is `backing.amount`.
- *something important to note*: In order to accurately track `payment_method` and `pledge_total`, we don't track the project page view events until the project is fully loaded.
- Alphabetized properties in `assertProjectProperties`.
- Updated tests.

# 👀 See
Nothing 2 c here

# 📋 QA
check ur local LogCat
On debug builds, we log the Koala events and their properties.

All events should be sending the `is_vo_on` property as a `boolean`.
Backed projects should be sending `payment_method` (friendly reminder, if you back a project that wasn't created in the same environment as your payment method, the payment method won't be returned.)
Backed projects should be sending `pledge_total`.

# Story 📖
[NT-339]

[NT-339]: https://dripsprint.atlassian.net/browse/NT-339


app/src/test/java/com/kickstarter/libs/KoalaTest.kt
```diff
@@ -237,6 +245,7 @@ class KoalaTest : KSRobolectricTestCase() {
         assertEquals("portrait", expectedProperties["device_orientation"])
         assertEquals("uuid", expectedProperties["distinct_id"])
         assertEquals("unavailable", expectedProperties["google_play_services"])
+        assertEquals(false, expectedProperties["is_vo_on"])
```
Setsun: would it be possible to replicate an environment where the `AccessibilityManager` is in a state where `is_vo_on` is true? 

this is mostly to make sure we're testing under as close to real conditions.

eoji: so since these are unit tests, not instrumented tests, i'd say no (learn more about instrumented tests here! https://source.android.com/compatibility/tests/development/instrumentation). the only think I could do would be to override `MockKoalaClient` and return `true`. but i'll be pairing with Infrastructure next week to actually run the tests on a device in Circle CI 😱

PR #631, opened 2019-10-03
[🛤] NT-396 Added checkout events for project page

eoji:
# 📲 What
Added checkout tracking events project page.

# 🤔 Why
TRACKING

# 🛠 How
- Added `BACK_THIS_PROJECT_BUTTON_CLICKED`  to `KoalaEvent`.
- Added `MANAGE_PLEDGE_BUTTON_CLICKED`  to `KoalaEvent`.
- Added `MANAGE_PLEDGE_OPTION_CLICKED`  to `KoalaEvent`.
- Added `VIEW_REWARDS_BUTTON_CLICKED`  to `KoalaEvent`.
- Added `VIEW_YOUR_PLEDGE_BUTTON_CLICKED`  to `KoalaEvent`.
- Removed `VIEWED_PROJECT_PAGE` from `KoalaEvent`.
- Tracking `Back this Project Button Clicked` when a user clicks the `Back this project` button on the project page.
- Tracking `Manage Pledge Button Clicked` when a user clicks the `Manage` button on the project page.
- Tracking `View Rewards Button Clicked` when a user clicks the `Back this project` button on the project page.
- Tracking `View Your Pledge Button Clicked` when a user clicks the `Back this project` button on the project page.
- Tracking `Manage Pledge Option Clicked` when a user clicks an option in the manage/view pledge menu.
  - This event sends an additional property `cta`, with the values 
    - "Update pledge" - `update_pledge`
    - "Change payment method" - `change_payment_method`
    - "Choose another reward" - `choose_another_reward`
    - "Cancel pledge" - `cancel_pledge`
    - "Contact creator" - `contact_creator`
    - "View rewards" -` view_rewards`
- Added `ProjectFactory.backedSuccessfulProject` to make testing easier.
- Tests.

# 👀 See
nothing 2 c

# 📋 QA
check ur local LogCat
On debug builds, we log the Koala events and their properties.

Clicking the `Back this project` button on the project page should fire the `Back this Project Button Clicked` event with default project properties.

Clicking the `Manage` button on the project page should fire the `Manage Pledge Button Clicked` event with default project properties.

Clicking the `View rewards` button on the project page should fire the `View Rewards Button Clicked` event with default project properties.

Clicking the `View your pledge` button on the project page should fire the `View Your Pledge Button Clicked` event with default project properties.

Selecting an option from the view/manage pledge overflow menu should fire a `Manage Pledge Option Clicked` event with the additional corresponding `cta` property defined above.

# Story 📖
[NT-396]

[NT-396]: https://dripsprint.atlassian.net/browse/NT-396


app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.kt
```diff
@@ -589,6 +589,46 @@ interface ProjectViewModel {
                     .map { p -> if (p.isStarred) R.drawable.icon__heart else R.drawable.icon__heart_outline }
                     .subscribe(this.heartDrawableId)
 
+            //Tracking
+            this.rewardsButtonText
```
oliverswitzer: It seems this file has gotten pretty large. I realize this may be stylistic, but in an effort to have this class take on a bit less responsibility, would it make sense (or even be possible?? I don't know the android things) to extract a class that's responsible for doing tracking in this view? I.e. its purpose is to subscribe to all relevant  emissions (from `nativeCheckoutProject` and whatever else we'd want to listen to) and track the correct things? Might just make it one less responsibility to test/worry about in the `ProjectViewModel`

eoji: Hmm...I like the idea. So it'd be basically an extension of the ProjectViewModel, right? But it needs to have access to the inputs and outputs and whatever helper vars I create 🤔 

But yeah this screen and the pledge screen basically hold all the majority of the business logic of the app.

oliverswitzer: wow just tried to pseudo code what I think the class would look like but was failing, hard lol

I didn't realize that you'd have to also pass in the PublishSubjects `viewRewardsClicked`, `cancelPledgeClicked`, `contactCreatorClicked`, `updatePledgeClicked` and `nativeProjectActionButtonClicked` in addition to all the observables so maybe this pattern doesn't make a ton of sense :X Sorry!

eoji: No need to be sorry! I would like move all the pledge sheet logic to their own ViewHolder (and corresponding ViewModel) but haven't figured that out yet 🤔 🔜

PR #635, opened 2019-10-11
[🗣] NT-252 Improved a11y on project page

eoji:
# 📲 What
Improving a11y on project page when native checkout is disabled or enabled.

# 🤔 Why
Support the people.

# 🛠 How
- Project screen is now fully traversable when native checkout is enabled or disabled.
  - `Manage` button's `contentDescription` is `Manage your pledge`.
  - When the `Fragment`s back stack changes, we set all but the top `Fragment`'s view as accessible.
  - When the pledge sheet is collapsed:
    - Project toolbar requests focus.
    - The project details are accessible.
  - When the pledge sheet is expanded:
    - Pledge toolbar requests focus.
    - The project details are no longer accessible (because they're obscured).
- Added a 1 second delay to loading the project so the pledge sheet doesn't steal focus when native checkout is enabled.
- Renamed `rewards_toolbar` to `pledge_toolbar` for accuracy.
  - Added proper collapse icon as `ic_arrow_down.xml`
- Renamed `rewards_container` to `pledge_container_root` for accuracy.
- Removed explicitly setting the project stats `contentDescription` in the `ProjectViewHolder` because it's called before the project is fully loaded and incomplete.
- Fixed crash when attempting to update `BackingFragment` and `PledgeFragment` when the `ProjectActivity` is finishing.

# 👀 See
| After 🦋 | Before 🐛 |
| --- | --- |
| screenshot-2019-10-11_101547 | screenshot-2019-10-11_101609 |

# 📋 QA
Turn on TalkBack™ and navigate to the project screen. The focus should start at the toolbar and work down the rest of the screen. 

# Story 📖
[NT-252]

[NT-252]: https://dripsprint.atlassian.net/browse/NT-252


app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
```diff
@@ -130,6 +132,15 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
 
                 this.supportFragmentManager.addOnBackStackChangedListener {
                     this.viewModel.inputs.fragmentStackCount(this.supportFragmentManager.backStackEntryCount)
+                    val fragments = this.supportFragmentManager.fragments
+                    val lastFragmentWithView = fragments.last { it.view != null }
+                    for (fragment in fragments) {
+                        if (fragment == lastFragmentWithView) {
+                            fragment.view?.importantForAccessibility = View.IMPORTANT_FOR_ACCESSIBILITY_YES
```
Setsun: [TIL](https://developer.android.com/reference/android/view/View#IMPORTANT_FOR_ACCESSIBILITY_YES)

app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
```diff
@@ -345,6 +359,14 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
         }
     }
 
+    private fun setProjectActionCTA(stringRes: Int) {
+        project_action_button.setText(stringRes)
+        project_action_button.contentDescription = when (stringRes) {
+            R.string.Manage -> getString(R.string.Manage_your_pledge)
```
Setsun: Ah interesting, is this string mapping because the `contentDescription` is what is used by screen readers?

app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.kt
```diff
@@ -26,6 +26,7 @@ import rx.subjects.BehaviorSubject
 import rx.subjects.PublishSubject
 import java.math.RoundingMode
 import java.net.CookieManager
+import java.util.concurrent.TimeUnit
```
Setsun: might be a leftover import

app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
```diff
@@ -553,14 +585,14 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
...
-        rewards_container.y = (rewards_container.height - guideline).toFloat()
+        pledge_container_root.y = (pledge_container_root.height - guideline).toFloat()
         this.projectRecyclerView.setPadding(0, 0, 0, guideline)
     }
 
     private fun showBackingFragment(project: Project) {
         val (rewardsFragment, backingFragment) = supportFragmentManager.findFragmentById(R.id.fragment_rewards) as RewardsFragment to
                 supportFragmentManager.findFragmentById(R.id.fragment_backing) as BackingFragment
-        if(!rewardsFragment.isHidden && supportFragmentManager.backStackEntryCount == 0) {
+        if(!rewardsFragment.isHidden && supportFragmentManager.backStackEntryCount == 0 && !isFinishing) {
```
Setsun: What does `isFinishing` mean in this context? I might have missed where it was defined in this file.

app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
eoji: yep! and normally for `Button`s it just reads the text

app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
eoji: Oh it's a property of an `Activity` https://developer.android.com/reference/android/app/Activity 
Screen Shot 2019-10-11 at 10 57 02 AM

So like when the user hits the back button, the `Activity` will call `finish`

Setsun: Ahh okay got it, thanks for the clarification!

PR #637, opened 2019-10-11
[🔁] NT-379 Synchronous createBacking

eoji:
# 📲 What
Handling the synchronous createBacking mutation and taking users back to the project page after Thanks.

# 🤔 Why
The server now executes this work synchronously woohoo. Thanks @cj-dimaggio 

# 🛠 How
- `CreateBacking` mutation now determines a checkout with a `SUCCESSFUL` state as a successful response since it's synchronous~
- I played myself by adding delays to the Project page so I removed them. To properly fix a11y and not have any random delays, I just gave the pledge container root some initial margin so it's not at the same level of the toolbar and RecyclerView with help from this [answer](https://stackoverflow.com/a/48157015)
- Instead of checking if the project `isFinishing` and committing a `FragmentTransaction` (the Activity might still call `saveInstanceState` even if it's not finishing),I'm  now just setting the `BackingFragment` and `RewardsFragment` as `VISIBLE` or `GONE`.
- The pledge sheet can now be collapsed without animating which fixes having to delay showing the cancel pledge `Snackbar`:
  - The `expandPledgeSheet` output in `ProjectViewModel` is now a `Pair<Boolean, Boolean>` where the first boolean is whether we're expanding and the second is whether we're animating.
- Renamed `ProjectActivity.animateRewards` to `expandPledgeSheet` for consistency.
- Added `pledgeSuccessfullyCreated` to the `PledgeDelegate`.
  - `ProjectViewModel` now handles showing the `ThanksActivity` with the output `startThanksActivity`.
- `PledgeFragmentViewModel` output `startThanksActivity `is now `showPledgeSuccess` that notifies the delegate.
- Hitting the close button of the `ThanksActivity` when native checkout is enabled goes back to the `ProjectActivity`.
- Tests.

# 👀 See
| After 🦋 | Before 🐛|
| --- | --- |
| device-2019-10-11-164841 2019-10-11 16_49_29 | device-2019-10-11-164803 2019-10-11 16_49_36 |

# 📋 QA
Create a backing!

# Story 📖
[NT-379]

[NT-379]: https://dripsprint.atlassian.net/browse/NT-379


app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
```diff
@@ -132,7 +132,7 @@ class PledgeFragmentViewModelTest : KSRobolectricTestCase() {
         this.vm.outputs.startLoginToutActivity().subscribe(this.startLoginToutActivity)
         this.vm.outputs.startRewardExpandAnimation().subscribe(this.startRewardExpandAnimation)
         this.vm.outputs.startRewardShrinkAnimation().subscribe(this.startRewardShrinkAnimation)
-        this.vm.outputs.startThanksActivity().subscribe(this.startThanksActivity)
+        this.vm.outputs.showPledgeSuccess().subscribe(this.showPledgeSuccess)
```
eoji: this should be alphabetized

app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.kt
```diff
@@ -256,11 +263,11 @@ interface ProjectViewModel {
         private val rewardsToolbarTitle = BehaviorSubject.create<Int>()
         private val scrimIsVisible = BehaviorSubject.create<Boolean>()
         private val setInitialRewardPosition = BehaviorSubject.create<Void>()
-        private val showBackingFragment = BehaviorSubject.create<Project>()
+        private val showBackingFragment = PublishSubject.create<Project>()
```
eoji: let's use a `BehaviorSubject` and see if it affects rotation

eoji: So this will affect rotation, here's the scenario:

ok yes i remember why I changed them to `BehaviorSubject`!
so let's say you view an unbacked project
`showRewardsFragment` will emit
then you back it
`showBackingFragment` will emit
then you cancel it
`showRewardsFragment` will emit again

now you rotate and the project is refreshed
if they're `BehaviorSubject`s both will re-emit but that's not what we want
we only want one to emit

PR #641, opened 2019-10-16
[😭] NT-412 Pledge validation

eoji:
# 📲 What
Disabling pledge buttons when user has entered an invalid pledge amount. Showing a warning if they go above the maximum pledge.

# 🤔 Why
So users are clear about what the pledge expectations are.

# 🛠 How
HOW?!
## XML
- Added `item_reward_pledge_card_disabled.xml` which is basically identical to `item_reward_pledge_card.xml` but the `Pledge` button is not enabled.
  - Calling `RecyclerView.Adapter.notifyItemChanged` does not guarantee that `onBindViewHolder` will be called again. But `getItemViewType` will so by returning a different layout, we can guarantee that the proper UI will be shown.
- Fixed the bug with clicking the close card button in `item_reward_pledge_card_disabled.xml` and `item_reward_pledge_card.xml`
- Moved `pledge_minimum` label under the `Your pledge amount` label and added `pledge_maximum` label that is initially `INVISIBLE`.
- The maximum number of characters a user can enter in the pledge amount is 13. The maximum pledge (numerically) is `1,200,000.00` which is 12 characters including the separators **but** if the user didn't enter a decimal they could type `1,200,000,000`.

## Business logic
- Added output `pledgeMaximum` to `PledgeFragmentViewModel` that emits the maximum amount a user can pledge (including shipping).
- Added output `pledgeMaximumIsGone` to `PledgeFragmentViewModel` that emits a `Boolean` determining if the maximum should be visible to the user. It returns true when the user has entered an amount that would put them over the project's country's currency maximum.
- Added output `pledgeButtonIsEnabled` to `PledgeFragmentViewModel` that emits a `Boolean` determining if the `Pledge` button should be enabled. It returns true when the user has entered a valid pledge amount. It only emits for logged in users pledging or updating their payment method.
- Added output `continueButtonIsEnabled` to `PledgeFragmentViewModel` that emits a `Boolean` determining if the `Pledge` button should be enabled. It returns true when the user is logged out and has entered a valid pledge amount. It only emits for logged out users.
- Removed output `showMinimumWarning` from `PledgeFragmentViewModel` because we no longer need to warn them if they attempt to pledge an invalid amount; the button will be disabled.

## Tests
Well, it damn near broke every test in `PledgeFragmentViewModelTest`.

# 👀 See
| After 🦋 | Before 🐛 | 
| --- | --- |
| device-2019-10-16-134923 2019-10-16 13_49_58 | screenshot-2019-10-16_141345 |
| screenshot-2019-10-16_142804 | screenshot-2019-10-16_142941 |
| screenshot-2019-10-16_142633 | screenshot-2019-10-16_142533 |

# 📋 QA
Attempt to pledge with some invalid amounts!

# Story 📖
[NT-412]

[NT-412]: https://dripsprint.atlassian.net/browse/NT-412


app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -590,8 +568,57 @@ interface PledgeFragmentViewModel {
...
+                    .map { it.first in it.second.first..it.second.second }
+                    .distinctUntilChanged()
+
+            totalIsValid
+                    .map { if (it) R.color.ksr_green_500 else R.color.ksr_red_400 }
+                    .distinctUntilChanged()
+                    .compose(bindToLifecycle())
+                    .subscribe(this.pledgeTextColor)
+
+            total
```
eoji: we can swap so we don't have to negate

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -590,8 +568,57 @@ interface PledgeFragmentViewModel {
...
+            maximumPledge
+                    .compose<Pair<Double, Double>>(combineLatestPair(shippingAmount))
+                    .map { it.first - it.second }
+                    .compose<Pair<Double, Project>>(combineLatestPair(project))
+                    .map { this.ksCurrency.format(it.first, it.second, RoundingMode.HALF_UP) }
+                    .distinctUntilChanged()
+                    .compose(bindToLifecycle())
+                    .subscribe(this.pledgeMaximum)
+
+            pledgeInput
```
eoji: i think we need the total here, not the input

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -590,8 +568,57 @@ interface PledgeFragmentViewModel {
...
+
+            total
+                    .compose<Pair<Double, Double>>(combineLatestPair(maximumPledge))
+                    .map { it.first > it.second }
+                    .map { BooleanUtils.negate(it) }
+                    .distinctUntilChanged()
+                    .compose(bindToLifecycle())
+                    .subscribe(this.pledgeMaximumIsGone)
+
+            maximumPledge
```
eoji: rename...

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -759,30 +793,31 @@ interface PledgeFragmentViewModel {
                     .compose(bindToLifecycle())
                     .subscribe(this.showNewCardFragment)
 
-            pledgeLessThanMinimum
+            totalIsValid
```
eoji: we don't need the check anymore

app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -759,30 +793,31 @@ interface PledgeFragmentViewModel {
...
+                    .filter { it == PledgeReason.PLEDGE || it == PledgeReason.UPDATE_PAYMENT }
+                    .compose<Pair<PledgeReason, Boolean>>(combineLatestPair(totalIsValid))
                     .map { it.second }
-                    .map { this.ksCurrency.format(it.second.minimum(), it.first, RoundingMode.HALF_UP) }
                     .compose(bindToLifecycle())
-                    .subscribe(this.showMinimumWarning)
+                    .subscribe(this.pledgeButtonIsEnabled)
 
-            val validPledgeClick = pledgeLessThanMinimum
+            val validPledgeClick = totalIsValid
```
eoji: i don't need to check if valid because button would be disabled

PR #642, opened 2019-10-16
[💰] NT-441 Showing backer if they used Google Pay or Apple Pay

eoji:
# 📲 What
Displaying backing payment source when it's Google Pay™ or Apple Pay™. 

# 🤔 Why
So the user knows what they used to pledge.

# 🛠 How
- Added `apple_pay_mark.xml` 

- Added `google_pay_mark.xml`

- Renamed `cardIsGone` output in `BackingFragmentViewModel` to `paymentMethodIsGone` and we now hide the whole row when it's `true`.
- The payment method section is hidden if the `backing.paymentSource` is not Google Pay, Apple Pay or a credit card.
- Created a helper method `cardLogo` to determine what logo we should show for the payment method.
- Updated tests.

# 👀 See
| After 🦋 |
| --- |
|  |

# 📋 QA
You would need a backing created on the same environment you're testing with Apple Pay or Google Pay for this to work. (Almost all my prod backings are GPay)

# Story 📖
[NT-441]

[NT-441]: https://dripsprint.atlassian.net/browse/NT-441


app/src/main/java/com/kickstarter/ui/fragments/BackingFragment.kt
```diff
@@ -54,10 +54,10 @@ class BackingFragment: BaseFragment<BackingFragmentViewModel.ViewModel>()  {
                 .compose(Transformers.observeForUI())
                 .subscribe { setCardLastFourText(it) }
 
-        this.viewModel.outputs.cardIsGone()
+        this.viewModel.outputs.paymentMethodIsGone()
```
ifbarrera: What does "Gone" mean in this context? 😅

eoji: It means that the row is not visible to the user. `GONE` is a visibility flag of a `View`. This line in the description:
> The payment method section is hidden if the backing.paymentSource is not Google Pay, Apple Pay or a credit card.

So basically it'd hide for bank accounts.

PR #643, opened 2019-10-17
[🔥] Internal tools redesign and config feature flags display

eoji:
# 📲 What
An absolutely fantastic redesign of the `Internal tools` screen, if I say so myself.
Displaying the config's feature flags in the `Feature Flags` screen.

# 🤔 Why
So we know what the current config's feature flags are and for fun.

# 🛠 How
- `build.gradle` had a method that allegedly was the `commitTime` but it just got the current date and cached it? So I made it actually return the time the build was created. It doesn't get cached because we want the most recent build date.
- Renamed `FeatureFlagActivity` to `FeatureFlagsActivity`.
- `FeatureFlagsActivity` now displays all the config's Android flags with a cute lil decoration and any overrides.
- `InternalToolsActivity` got a redesign.
  - It now properly displays the `Build Date`.
  - It now displays the `API Endpoint`.
- I wrote tests but we don't run tests on the internal builds in Circle 🤕 

# 👀 See

| Internal tools after 🦋 | Internal tools before 🐛 |
| --- | --- |
| screenshot-2019-10-17_170637 | screenshot-2019-10-17_170645 |
| Feature flags after 🦋 | Feature flags before 🐛 |
| screenshot-2019-10-17_170821 | screenshot-2019-10-17_170816 |

# 📋 QA
## Plz note
The `android_native_checkout` flag is not currently being read because the feature isn't done 🙂 

Check out the Internal Tools from the drawer of an Internal build!

# Story 📖
[NT-345]

[NT-345]: https://dripsprint.atlassian.net/browse/NT-345


eoji:
@yatriks should creating an internal release run the internal tests? right now it runs the external tests only

PR #646, opened 2019-10-21
[💳] NT-369 createBacking with SCA support

eoji:
# 📲 What
SCA support when pledging to project for the first time.

# 🤔 Why
So SCA affected users can successfully back projects.

# 🛠 How
- Added `Checkout` and `Checkout.Backing` models and corresponding `CheckoutBackingFactory` for testing.
- Moved `CreateBacking` data class to its own class and renamed to `CreateBackingData`.
- Updated `CreateBacking` mutation to return `checkout.backing.requiresAction` and `checkout.backing.clientSecret`. It now takes in a `CreateBackingData` object and returns a `Checkout.Backing` object.
  - When `requiresAction` is `true`, we start the SCA flow by calling [`Stripe. authenticateSetup` ](https://stripe.dev/stripe-android/com/stripe/android/Stripe.html#authenticateSetup-fragment-clientSecret-)  with the `clientSecret`. In `onActivityResult` we handle the result from Stripe.
    - If it's successful, we show the Thanks page.
    - If it's unsuccessful, we show the error `Snackbar` and let the user attempt to pledge again.
  - When `requiresAction` is `false`, we assume the backing was created successfully and show the Thanks page.

## Bug fix
- Reducing the initial margin of `pledge_container_root` to `1px` fixes the bug with the screen not resizing when `Add new card` is visible and maintains TalkBack behavior.

# 👀 See
Well, you actually can't see for security reasons. The screen is black.
device-2019-10-21-152115 2019-10-21 15_59_59

# 📋 QA
Back a project using a test card from [here](https://stripe.com/docs/mobile/android/authentication#testing)

# Story 📖
[NT-369]

[NT-369]: https://dripsprint.atlassian.net/browse/NT-369


app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -880,7 +876,7 @@ interface PledgeFragmentViewModel {
 
             val createBackingSuccess = Observable.merge(this.stripeSetupResultSuccessful,
                     createBackingValues
-                            .map { it.backing().requiresAction() }
+                            .map { it.requiresAction() }
```
Scollaco: Now looking again I don't think it's necessary to create a variable here either.

PR #648, opened 2019-10-22
[👷🏾‍♀️] NT-452 Stripe 11.1.3 -> 12.0.1

eoji:
# 📲 What
Updated Stripe to 12.0.1 and fixed every thing that broke.

# 🤔 Why
Alwayz b upgrading

# 🛠 How
- `Stripe` is now final and can't be extended so `MockStripe` is deleted.
- `StripeEditText` is now `internal` so I can't access it in `ProjectActivity`
- Using same callback pattern as in #646 in the `NewCardFragment` for creating Stripe `Token`.
- Updated tests.

# 👀 See
Looks the same.

# 📋 QA
Save a [card](https://stripe.com/docs/testing)

# Story 📖
[NT-452]

[NT-452]: https://dripsprint.atlassian.net/browse/NT-452


app/src/main/java/com/kickstarter/viewmodels/NewCardFragmentViewModel.kt
```diff
@@ -184,46 +195,51 @@ interface NewCardFragmentViewModel {
...
                     .compose(bindToLifecycle())
+                    .subscribe {
+                        this.createStripeToken.onNext(it)
+                        this.progressBarIsVisible.onNext(true)
+                    }
+
+            val saveCardNotification = this.stripeTokenResultSuccessful
+                    .map { token -> token.card?.id?.let { Pair(token.id, it) } }
+                    .compose<Pair<Pair<String, String>, Boolean>>(combineLatestPair(reusable))
+                    .switchMap { this.apolloClient.savePaymentMethod(PaymentTypes.CREDIT_CARD, it.first.first, it.first.second, it.second).materialize() }
```
eoji: make data object for this input

app/src/main/java/com/kickstarter/viewmodels/NewCardFragmentViewModel.kt
```diff
@@ -184,46 +195,51 @@ interface NewCardFragmentViewModel {
...
                     .share()
 
             saveCardNotification
                     .compose(values())
+                    .compose(bindToLifecycle())
                     .subscribe {
-                        this.success.onNext(null)
-                        this.koala.trackSavedPaymentMethod()
+                        this@ViewModel.success.onNext(it)
```
eoji: we don't need @ViewModel

PR #651, opened 2019-10-23
[🙈] NT-466 Pledge summary visibility in the view/manage pledge screen

eoji:
# 📲 What
Hiding the pledge summary in the view/manage pledge screen when a backing doesn't have a location.

# 🤔 Why
The pledge amount is the same as the total.

# 🛠 How
- Added `pledgeSummaryIsGone` output to the `BackingFragmentViewModel`. It  uses the same logic as the `shippingSummaryIsGone` output. They're both hidden if the backing does not have a `locationId`.

# 👀 See
| After 🦋 | Before 🐛 |
| --- | --- |
| screenshot-2019-10-23_140012 | screenshot-2019-10-23_135949 |

# 📋 QA
View a pledge with no shipping!

# Story 📖
[NT-466]

[NT-466]: https://dripsprint.atlassian.net/browse/NT-466

### DIFF
--- a/app/src/test/java/com/kickstarter/viewmodels/BackingFragmentViewModelTest.kt
+++ b/app/src/test/java/com/kickstarter/viewmodels/BackingFragmentViewModelTest.kt
@@ -255,7 +255,7 @@ class BackingFragmentViewModelTest :  KSRobolectricTestCase() {
     }
 
     @Test
-    fun testPledgeSummaryIsGone_whenDigitalReward() {
+    fun testPledgeSummaryIsGone_whenLocationId_isNull() {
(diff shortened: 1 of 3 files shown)


app/src/test/java/com/kickstarter/viewmodels/BackingFragmentViewModelTest.kt
```diff
@@ -252,6 +254,40 @@ class BackingFragmentViewModelTest :  KSRobolectricTestCase() {
         this.pledgeDate.assertValue("August 28, 2019")
     }
 
+    @Test
```
eoji: let's add a test for no reward

PR #652, opened 2019-10-23
[⚠️] NT-448 Displaying deadline warning on pledge screen

eoji:
# 📲 What
Displaying deadline warning on pledge screen.

# 🤔 Why
So backers are CRYSTAL CLEAR 👓 on what they're pledging.

# 🛠 How
- Updated config and strings.
- Moved `addBoldSpan` to `ViewUtils` since it is used in multiple places. (`CancelPledgeFragment` and `PledgeFragment`)
- Added `totalAndDeadline` output to `PledgeFragmentViewModel`. It emits the current total in the current project's currency and the current project's deadline.
- Added deadline warning `TextView` to `fragment_pledge.xml`.
- Updated spacing in total section `fragment_pledge_section_total.xml`.
- Tests 😭 

# 👀 See
| Logged Out -  Unbacked - Same currency | Logged Out -  Unbacked - Same currency |
| --- | --- |
| screenshot-2019-10-23_173243 | screenshot-2019-10-23_175158 |
| Logged Out -  Unbacked - Different currency | Logged Out -  Unbacked - Different currency |
| screenshot-2019-10-23_174953 | screenshot-2019-10-23_175103 |
| Updating pledge | Updating payment |
| screenshot-2019-10-23_175609 | screenshot-2019-10-23_175617 |
| Choosing a new reward | 
| screenshot-2019-10-23_175731 |

# 📋 QA
View the pledge screen in the above situations ⬆️ 

# Story 📖
[NT-448]

[NT-448]: https://dripsprint.atlassian.net/browse/NT-448


app/src/main/java/com/kickstarter/libs/utils/ViewUtils.java
```diff
@@ -32,6 +36,12 @@
 public final class ViewUtils {
   private ViewUtils() {}
 
+  public static void addBoldSpan(final @NonNull SpannableString spannableString, final @NonNull String substring) {
+    final int index = spannableString.toString().indexOf(substring);
```
ifbarrera: What happens if the `substring` can't be found in the `spannableString`? 🤔

app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
```diff
@@ -486,6 +491,23 @@ class PledgeFragment : BaseFragment<PledgeFragmentViewModel.ViewModel>(), Reward
...
+                "total", total,
+                "project_deadline", deadline)
+
+        val spannableWarning = SpannableString(warning)
+
+        ViewUtils.addBoldSpan(spannableWarning, total)
+        ViewUtils.addBoldSpan(spannableWarning, deadline)
+
+        deadline_warning.text = spannableWarning
+        deadline_warning.visibility = View.VISIBLE
```
ifbarrera: I wondering if setting the visibility of this view should be it's own view model output? Or alternatively we could consider re-naming the output from `totalAndDeadline` to something that also communicates that the label will also be set to `Visible` 🤔

app/src/main/java/com/kickstarter/libs/utils/ViewUtils.java
eoji: i let it blow up, (which happened yesterday because I used the wrong string 😛) I can abandon ship if the string can't be found!

app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
eoji: it should always be visible, i just set it to `visible` so it fades in because it depends on the total loading (which depends on the shipping rules call)

PR #658, opened 2019-11-01
[🐛] NT-467 Fixes bug with pledge text color and max warning

eoji:
# 📲 What
Fixes bug with pledge text color and max warning

# 🤔 Why
👞🐜

# 🛠 How
- The `pledgeTextColor` output now depends on the rewardMinimum and the pledge maximum (country max - shipping). It's green when it's valid and red when it's invalid.
- The `pledgeMaximumIsGone` output now only emits true when the maximum pledge is greater than OR equal to the pledge input. 
- Added additional test for when the pledge input is equal to the maximum pledge.

# 👀 See
| Before 🐛 | After 🦋 |
| --- | --- |
| screenshot-2019-11-01_102253 | screenshot-2019-11-01_102238 |

# 📋 QA
Invalid amounts should turn the pledge input red.
The max pledge warning should show when the pledge amount is greater.

# Story 📖
[NT-467]

[NT-467]: https://dripsprint.atlassian.net/browse/NT-467


eoji:
I was wondering the same thing because the tests test:
1. Initial amount
2. Valid additional amount
3. Less than min amount amount
4. more than max amount

Should I add 5. equal to max amount?

ifbarrera:
> I was wondering the same thing because the tests test:
> 
> 1. Initial amount
> 2. Valid additional amount
> 3. Less than min amount amount
> 4. more than max amount
> 
> Should I add 5. equal to max amount?

Yea I think adding `5. equal to max amount` would make sense! That way if this ever regressed a test would catch it 👍

PR #284, opened 2018-07-31
[Enhancement] - Added Crashlytics for better analytics tracking

Rcureton:
## What
- Integrated Crashlytics into the Android app.
- Renamed toolbar dimen because of naming conflict with stripe sdk.
## Why
- To better track information about our app and our users.

PR #659, opened 2019-11-01
[💥] NT-443 Fixing Crashlytics setup

eoji:
# 📲 What
Fixes Crashlytics setup and cleans up `app/build.gradle`.

# 🤔 Why
So we can track those sweet (BUT EXTREMELY RARE) crashes.

# 🛠 How
- In #284, Fabric was added without the API key meta-data in the manifest which didn't cause issues _until_ the migration to Firebase happened and none of the apps were showing crashes.
   - Since we don't want to commit the key because #OSS, we can use a [`manifestPlaceHolder`](https://developer.android.com/studio/build/manifest-build-variables) and read in the value when we build the app (https://stackoverflow.com/questions/37101589/how-to-read-a-properties-files-and-use-the-values-in-project-gradle-script)
  - Added a button to manually crash the app to the Internal Tools so I could easily verify the crashes were being sent to Fabric/Firebase.
  - Updated Crashlytics library from 2.9.6 to 2.10.1
- I have found it pretty frustrating to figure out what version numbers are for which library in `build.gradle`. 
  - Alphabetized all of our dependencies.
  - Only version numbers that are used in multiple places use a variable.

# 👀 See
| New crash button 💥 |
| --- |
|  |

Screen Shot 2019-11-01 at 12 51 34 PM

# 📋 QA
SMASH THAT CRASH BUTTON 🗣 

# Story 📖
[NT-443]

[NT-443]: https://dripsprint.atlassian.net/browse/NT-443


app/build.gradle
```diff
@@ -166,113 +170,75 @@ apollo {
...
     implementation fileTree(include: ['*.jar'], dir: 'libs')
     implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'androidx.appcompat:appcompat:1.1.0-rc01'
+    implementation 'androidx.browser:browser:1.0.0'
     implementation 'androidx.cardview:cardview:1.0.0'
     implementation 'androidx.recyclerview:recyclerview:1.1.0-beta01'
-    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
     implementation 'androidx.multidex:multidex:2.0.1'
-    implementation 'androidx.browser:browser:1.0.0'
+    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
```
Setsun: Inlining the versions makes sense 👍

app/src/internal/java/com/kickstarter/ui/activities/InternalToolsActivity.java
```diff
@@ -99,6 +99,11 @@ public void changeEndpointProductionButton() {
     setEndpointAndRelaunch(ApiEndpoint.PRODUCTION);
   }
 
+  @OnClick(R.id.crash_button)
+  public void crashButtonClicked() {
+    throw new RuntimeException("Forced a crash!");
+  }
```
Setsun: Noice

PR #660, opened 2019-11-01
[⛔️] NT-509 Hiding reward CTA for creators on their own project

eoji:
# 📲 What
Hiding the `Select` button when a creator views the rewards of their own project.

# 🤔 Why
Because they can't back their own project.

# 🛠 How
- The `buttonIsGone` output of `NativeCheckoutRewardViewHolderViewModel` now checks if the current user is the current project's creator and returns `true` if they are.
- Split all the `buttonUIOutput` test cases into their own tests.
- Added tests for making sure the button is gone when the current user is the current project's creator.

# 👀 See
| Logged  in 🦋 | Logged out 🐛 |
| --- | --- |
| device-2019-11-01-151731 2019-11-01 15_20_09 | device-2019-11-01-151640 2019-11-01 15_20_20 |

# 📋 QA
View the rewards of a project you created.

# Story 📖
[NT-509]

[NT-509]: https://dripsprint.atlassian.net/browse/NT-509

### DIFF
--- a/app/src/test/java/com/kickstarter/viewmodels/NativeCheckoutRewardViewHolderViewModelTest.kt
+++ b/app/src/test/java/com/kickstarter/viewmodels/NativeCheckoutRewardViewHolderViewModelTest.kt
@@ -5,10 +5,8 @@ import androidx.annotation.NonNull
 import com.kickstarter.KSRobolectricTestCase
 import com.kickstarter.R
 import com.kickstarter.libs.Environment
-import com.kickstarter.mock.factories.BackingFactory
-import com.kickstarter.mock.factories.LocationFactory
-import com.kickstarter.mock.factories.ProjectFactory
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/kickstarter/viewmodels/NativeCheckoutRewardViewHolderViewModelTest.kt
```diff
@@ -115,91 +113,183 @@ class NativeCheckoutRewardViewHolderViewModelTest : KSRobolectricTestCase() {
     }
 
     @Test
-    fun testButtonUIOutputs() {
+    fun testButtonUIOutputs_whenProjectIsLiveAndUnbacked_availableReward() {
```
Setsun: Splitting up the tests into multiple smaller ones makes sense 👍

PR #558, opened 2019-06-28
Native checkout landscape and More ™

eoji:
# 📲 What
Landscape native checkout.

# 🤔 Why
Native checkout landscape did not match the Abstract and non native checkout should not show the footer button.

# 🛠 How
- Any reference of "horizontal rewards" has been stripped of "horizontal" or renamed to "native checkout" for accuracy (the rewards might scroll vertically and then we would live in house of lies).
- Moved `rewardsButtonText` to `ProjectViewUtils` and added tests.
- There is some UI set up that needs to happen based on whether native checkout is enabled.
  - Created a completely separate layout activity_project for the ProjectActivity  when native checkout is enabled (the toolbar collapses amen).
    - When it's not enabled, we need to handle the visibility of the footer project action buttons and add their click listeners.
    - When it is enabled, we need to set the correct initial peek height of the rewards sheet.
- Matching Abstract designs for landscape native checkout.
  - The rewards sheet should hide all the way at the bottom of the screen.
  - Displaying View rewards button so we can access the rewards when native checkout is enabled.
- Outputs in ProjectViewModel specific to native checkout only emit when it's enabled.
- Fixed creator byline getting cut off in landscape.
- Finally fixed the back button bug when you minimize the sheet while the pledge screen is up and hit back.

# 👀 See
| Before 🐛 | After 🦋 |
| --- | --- |
|  screenshot-2019-06-28_160234 | screenshot-2019-06-28_160155 |
| --- | --- |
| screenshot-2019-06-28_160246 | screenshot-2019-06-28_160319 |

device-2019-06-28-160336 2019-06-28 16_07_56

# 📋 QA
- View the project page in landscape with or without native checkout enabled.

# Story 📖
[Landscape](https://trello.com/c/ayPAjOWn/1440-landscape-native-checkout)
[View rewards](https://trello.com/c/g9UlBQag/1431-project-states-view-rewards-non-backer-non-live-project)

PR #661, opened 2019-11-01
[🔙] NT-479 Pledge toolbar navigation

eoji:
# 📲 What
Reverting #558 logic for how the navigation icon in the pledge toolbar works and updating icon drawable based on fragment back stack count.

# 🤔 Why
Everyone clicks the down icon thinking it'll take them back and it doesn't. GIVE THE PEOPLE WHAT THEY WANT

# 🛠 How
- Added `ic_arrow_back.xml`. We now use this rounded style everywhere. Deleted `ic_action_icon__arrow_left.xml`. 
Screen Shot 2019-11-01 at 3 50 05 PM

- Updated `ic_arrow_down.xml` color to `ksr_soft_black`.
- Renamed `ProjectViewModel` input `collapsePledgeSheet ` to `pledgeToolbarNavigationClicked`.
- Added `pledgeToolbarNavigationIcon` output to `ProjectViewModel`. It emits the down icon drawable when the back stack is empty and the back icon when it's not.
- Added `goBack ` output to `ProjectViewModel`. It emits when the pledge toolbar navigation icon is clicked and there are fragments on the back stack.
  - `expandPledgeSheet` now only emits when the back stack is empty and we need to collapse the sheet.
- Renamed `rewardsToolbarTitle` to `pledgeToolbarTitle`.
- Tests

# 👀 See
| After 🦋 | Before 🐛 |
| --- | --- |
| device-2019-11-01-181539 2019-11-01 18_17_15 | device-2019-11-01-181626 2019-11-01 18_17_35 |

# 📋 QA
Navigate through all the pledge screens.

# Story 📖
[NT-479]

[NT-479]: https://dripsprint.atlassian.net/browse/NT-479

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
@@ -110,6 +110,11 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe { expandPledgeSheet(it) }
 
+        this.viewModel.outputs.goBack()
+                .compose(bindToLifecycle())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe { back() }
+
         this.viewModel.outputs.heartDrawableId()
                 .compose(bindToLifecycle())
                 .compose(Transformers.observeForUI())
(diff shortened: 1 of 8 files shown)


app/src/test/java/com/kickstarter/viewmodels/ProjectViewModelTest.kt
```diff
@@ -676,43 +680,92 @@ class ProjectViewModelTest : KSRobolectricTestCase() {
...
+
+        this.pledgeToolbarNavigationIcon.assertValue(R.drawable.ic_arrow_down)
+
+        this.vm.inputs.fragmentStackCount(1)
+
+        this.pledgeToolbarNavigationIcon.assertValues(R.drawable.ic_arrow_down, R.drawable.ic_arrow_back)
+
+        this.vm.inputs.fragmentStackCount(0)
+
+        this.pledgeToolbarNavigationIcon.assertValues(R.drawable.ic_arrow_down, R.drawable.ic_arrow_back, R.drawable.ic_arrow_down)
```
Setsun: Splitting up these tests makes the business logic much easier to review, thanks for doing that! 🙌

app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
```diff
@@ -489,6 +483,22 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
         finish()
     }
 
+    private fun pledgeFragment() = supportFragmentManager
+            .findFragmentByTag(PledgeFragment::class.java.simpleName) as PledgeFragment?
+
+    private fun pledgeFragmentShouldAnimateOut(): Boolean {
+        val backStackEntryCount = supportFragmentManager.backStackEntryCount
+        val backStackIsNotEmpty = backStackEntryCount > 0
+        val topOfStackIndex = backStackEntryCount.minus(1)
```
Setsun: still find it so interesting that the `-` operator and the function call are seemingly interchangable in Kotlin

PR #685, opened 2019-11-26
[✨] NT-587 Go rewardless carousel treatment

eoji:
# 📲 What
Go rewardless carousel treatment

# 🤔 Why
So users are ENCOURAGED to go rewardless.

# 🛠 How
- Added `background` output to `NativeCheckoutRewardViewHolderViewModel` that emits a `Int` representing the background of the reward contents. [`0`](https://developer.android.com/reference/android/view/View#setBackgroundResource(int)) is how you clear the background.
  - Added `bg_go_rewardless` as the special go rewardless background `Drawable`.
- Added special go rewardless copy when a user is viewing `No reward` and they haven't backed `No reward`.
- Decreased the size of the gradient behind the `Select` button for readability.
- Tests

# 👀 See
| Enabled ✔️  | Disabled ✖️  |
| --- | --- |
| screenshot-2019-11-25_171021 | screenshot-2019-11-25_171010 |

# 📋 QA
View no reward when go rewardless is enabled.

# Story 📖
[NT-587]

[NT-587]: https://dripsprint.atlassian.net/browse/NT-587

### DIFF
--- a/app/src/test/java/com/kickstarter/viewmodels/NativeCheckoutRewardViewHolderViewModelTest.kt
+++ b/app/src/test/java/com/kickstarter/viewmodels/NativeCheckoutRewardViewHolderViewModelTest.kt
@@ -5,7 +5,10 @@ import androidx.annotation.NonNull
 import com.kickstarter.KSRobolectricTestCase
 import com.kickstarter.R
 import com.kickstarter.libs.Environment
+import com.kickstarter.libs.FeatureKey
 import com.kickstarter.libs.MockCurrentUser
+import com.kickstarter.libs.preferences.MockBooleanPreference
+import com.kickstarter.mock.MockCurrentConfig
 import com.kickstarter.mock.factories.*
 import com.kickstarter.models.Project
 import com.kickstarter.models.Reward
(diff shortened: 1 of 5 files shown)


app/src/test/java/com/kickstarter/viewmodels/NativeCheckoutRewardViewHolderViewModelTest.kt
```diff
@@ -112,6 +117,33 @@ class NativeCheckoutRewardViewHolderViewModelTest : KSRobolectricTestCase() {
         this.backersCountIsGone.assertValue(true)
     }
 
+    @Test
+    fun testBackground_whenReward() {
+        setUpEnvironment(environmentWithGoRewardlessDisabled())
```
ifbarrera: Did you mean for this to be enabled, so you can test that a regular reward with the `Go Rewardless` feature _enabled_ doesn't apply a background? Or maybe that needs a different test?

eoji: hello and thank you for looking! yeah, i think i'll need another test! should i also add one for `testDescriptionOutputs_whenReward_goRewardlessEnabled`. There a lot of tests 😓

eoji: Added `testBackground_whenReward_goRewardlessEnabled` and renamed this one to `testBackground_whenReward_goRewardlessDisabled`

PR #688, opened 2019-11-26
[💥] NT-625 No Crash Zone™ 

eoji:
# 📲 What
Adding some bumpers in the Project page.

# 🤔 Why
So we don't crash!

# 🛠 How
- Added tracking breadcrumbs to Crashlytics in `KoalaBackgroundService`
- Since `FragmentManager.findFragmentById` returns nullable `Fragment`s, safely unwrapping `BackingFragment` and `RewardsFragment`.

# 👀 See
Check out these sweet breadcrumbs https://fabric.io/kickstarter2/android/apps/com.kickstarter.kickstarter.internal.debug/issues/60f1455fd23c9568a2e1edd61e889530/sessions/5DDDB6990210000116C5F000C2B933B0_DNE_0_v2?

# 📋 QA
I can't replicate these crashes IRL so smash the crash button in Internal Tools and check out the tracking breadcrumbs.

# Story 📖
[NT-625]

[NT-625]: https://dripsprint.atlassian.net/browse/NT-625

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
@@ -774,7 +774,9 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
 
     private fun updateFragments(project: Project) {
         try {
-            safeLet(rewardsFragment(), backingFragment()) { rewardsFragment, backingFragment ->
+            val rewardsFragment = rewardsFragment()
+            val backingFragment = backingFragment()
+            if (rewardsFragment != null && backingFragment != null) {
                 when {
                     supportFragmentManager.backStackEntryCount == 0 -> when {
                         project.isBacking -> if (!rewardsFragment.isHidden) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kickstarter/ui/activities/ProjectActivity.kt
```diff
@@ -768,27 +773,38 @@ class ProjectActivity : BaseActivity<ProjectViewModel.ViewModel>(), CancelPledge
...
-                            .commitNow()
-                }
-                else -> if (!backingFragment.isHidden) {
-                    supportFragmentManager.beginTransaction()
-                            .show(rewardsFragment)
-                            .hide(backingFragment)
-                            .commitNow()
+        try {
+            safeLet(rewardsFragment(), backingFragment()) { rewardsFragment, backingFragment ->
+                when {
```
justinswart: So the way I'm understanding `safeLet` is that this closure just won't execute if either of these values are `null`. Just wondering if it's necessary to define the new generic function for this purpose or just have that same `if (rewardsFragment() != null && backingFragment() != null)` just right here? Unless you feel it improves readability split out into a function.

Also - are there any threading concerns here? Do we need to be sure `block(p1, p2)` is executing on the main thread?

eoji: yes that's what it does! but no it does not improve readability, i was just being fancy :3 

wrt threading; it executes in a `run` block so I think I'd be okay

PR #689, opened 2019-11-27
[🚀] 2.1.0 RC

eoji:
# 📲 What
Updated strings, configs, version codes and names post 2.1.0 RC creation.

# 🤔 Why
We're going to the Play Store baybeeeeee


joshhenning:
Makes sense to move this forward whenever you’re ready - thanks, Izzy!

PR #692, opened 2019-12-04
[💧] NT-653 Set up additional tracking client to hit the Data Lake™ 

eoji:
# 📲 What
Added additional tracking client to send events to the Data Lake™ 

# 🤔 Why
We're cleaning up our event tracking.

# 🛠 How
- Abstracted reusable parts `KoalaTrackingClient` to `TrackingClient` so now `KoalaTrackingClient` is an extension of a `TrackingClient`.
- Added `LakeTrackingClient` which is an extension of a `TrackingClient` to the `Environment`.
  - Added `LakeBackgroundService` to handle background task of sending tracking event requests.
  - Added `LakeService` as API interface. It currently only sends a `PUT` request to the `record` endpoint with a `"Content-Type: application/json"` header.
  - Added `@LakeEndpoint`, `@LakeRetrofit`, `@LakeTracker`, and `@KoalaTracker` qualifiers so our dependency injection can differentiate between clients.
- Added fake values for `Secrets.LakeEndpoint.Staging` and `Secrets.LakeEndpoint.Production`. **Note:** I'm currently only sending events to staging until the events are fully finalized.

## what's getting tracked?
- I'm currently only tracking `App Open`, `App Close` and `Discover List View` events with the properties `time` and `user_logged_in` because those are whitelisted.
- These new events are being tested 🤓 

# 👀 See
I'm going surfing
Screen Shot 2019-12-03 at 7 45 11 PM

# 📋 QA
ktk the staging data lake endpoint

# Story 📖
[NT-653]

[NT-653]: https://dripsprint.atlassian.net/browse/NT-653

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/Environment.java
+++ b/app/src/main/java/com/kickstarter/libs/Environment.java
@@ -35,6 +35,7 @@ public abstract class Environment implements Parcelable {
   public abstract Koala koala();
   public abstract KSCurrency ksCurrency();
   public abstract KSString ksString();
+  public abstract Koala lake();
   public abstract Logout logout();
   public abstract BooleanPreferenceType nativeCheckoutPreference();
   public abstract PlayServicesCapability playServicesCapability();
@@ -62,6 +63,7 @@ public abstract class Environment implements Parcelable {
     public abstract Builder koala(Koala __);
     public abstract Builder ksCurrency(KSCurrency __);
     public abstract Builder ksString(KSString __);
+    public abstract Builder lake(Koala __);
     public abstract Builder logout(Logout __);
     public abstract Builder nativeCheckoutPreference(BooleanPreferenceType __);
     public abstract Builder playServicesCapability(PlayServicesCapability __);
--- a/app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
@@ -1,0 +1,151 @@
+package com.kickstarter.libs
+
+import android.content.Context
+import android.content.res.Configuration
+import android.os.Bundle
+import android.util.Log
+import android.view.accessibility.AccessibilityManager
+import com.google.android.gms.common.ConnectionResult
+import com.google.android.gms.common.GoogleApiAvailability
+import com.google.firebase.iid.FirebaseInstanceId
+import com.kickstarter.BuildConfig
+import com.kickstarter.R
+import com.kickstarter.libs.qualifiers.ApplicationContext
+import com.kickstarter.libs.utils.ConfigUtils
+import com.kickstarter.libs.utils.MapUtils
+import com.kickstarter.models.User
+import com.kickstarter.services.LakeBackgroundService
+import com.kickstarter.services.firebase.dispatchJob
+import com.kickstarter.ui.IntentKey
+import io.fabric.sdk.android.Fabric
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+import timber.log.Timber
+import java.util.*
+import javax.inject.Inject
+
+class LakeTrackingClient(
+        @param:ApplicationContext private val context: Context,
+        @set:Inject var currentUser: CurrentUserType,
+        @set:Inject var build: Build,
+        @set:Inject var currentConfig: CurrentConfigType) : TrackingClientType() {
+    private var loggedInUser: User? = null
+    private var config: Config? = null
+
+    init {
+
+        // Cache the most recent logged in user for default Lake properties.
+        this.currentUser.observable().subscribe { u -> this.loggedInUser = u }
+
+        // Cache the most recent config for default Lake properties.
+        this.currentConfig.observable().subscribe { c -> this.config = c }
+    }
+
+    override fun track(eventName: String, additionalProperties: Map<String, Any>) {
+        try {
+            val trackingData = getTrackingData(eventName, combinedProperties(additionalProperties))
+            val bundle = Bundle()
+            bundle.putString(IntentKey.LAKE_EVENT_NAME, eventName)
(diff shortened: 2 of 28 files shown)


app/src/main/java/com/kickstarter/libs/Environment.java
```diff
@@ -35,6 +35,7 @@
   public abstract Koala koala();
   public abstract KSCurrency ksCurrency();
   public abstract KSString ksString();
+  public abstract Koala lake();
```
Scollaco: Why is the `lake` property of type `Koala`? Is it to do exactly the same but with "clean properties"? 🤔

app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
```diff
@@ -0,0 +1,80 @@
...
+    init {
+
+        // Cache the most recent logged in user for default Lake properties.
+        this.currentUser.observable().subscribe { u -> this.loggedInUser = u }
+
+        // Cache the most recent config for default Lake properties.
+        this.currentConfig.observable().subscribe { c -> this.config = c }
+    }
+
+    override fun track(eventName: String, additionalProperties: Map<String, Any>) {
```
Scollaco: This function is very similar to the one existing on the `KoalaTrackingClient`. I wonder if we could move it to the parent class and call it from the children passing the parameters we need (`uniqueJobName`, `eventName`, `data`, etc)

app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
```diff
@@ -0,0 +1,80 @@
...
+            if (this.build.isDebug) {
+                Timber.e("Failed to encode event: $eventName")
+            }
+            Fabric.getLogger().e(TAG, "Failed to encode event: $eventName")
+        }
+
+    }
+
+    @Throws(JSONException::class)
+    private fun getTrackingData(eventName: String, newProperties: Map<String, Any>): String {
```
Scollaco: This one is also very similar (with the exception of the random string). What if we have a function on the parent that does at least part of the job? So we duplicate the less amount of code possible.

app/src/main/java/com/kickstarter/services/LakeBackgroundService.kt
```diff
@@ -0,0 +1,71 @@
...
+import rx.schedulers.Schedulers
+import javax.inject.Inject
+
+class LakeBackgroundService : JobService() {
+    @Inject
+    lateinit var lakeService: LakeService
+    @Inject
+    lateinit var build: Build
+
+
```
Scollaco: Extra line here

app/src/main/java/com/kickstarter/libs/Environment.java
eoji: `Koala` is just a wrapper around a `TrackingClient`. And yeah, `lake` can track properties but with the different JSON structure and "clean properties"

app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
eoji: They depend on different services `LakeBackgroundService` vs `KoalaBackgroundService` and the 2 services depend on different `IntentKey`s. So would I also be passing in the service class's name,`TAG`, and `IntentKey`s?

app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
eoji: oh no these are very different! Koala's data structure is:
```
{
    "event": "event_name",
    "properties": {}
}
```

and Lake's is:
```
{
  "partition-key": "d78cc21f-1b09-4ed8-a530-d897dcaf631a",
  "data": {
    "event": "event_name",
    "properties": {
      
    }
  }
}
```

app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
Scollaco: Yes... whatever differs from each other we would pass as argument. What do you think?

app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
Scollaco: Hmm got it!

app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
eoji: I can try it and open it as a PR against this branch since it'll touch 3 files!

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
```diff
@@ -119,6 +123,7 @@ public void onLowMemory() {
   public void onTrimMemory(final int i) {
     if(i == ComponentCallbacks2.TRIM_MEMORY_UI_HIDDEN) {
       this.koala.trackAppClose();
+      this.lake.trackAppClose();
```
justinswart: 🤔 If the structure on iOS is:

`Callsite` -> `Koala` -> `TrackingClient` -> `Network`

Is it correct to say the structure on Android is:

`Callsite` -> `TrackingClient` -> `BackgroundService` -> `Network`?

What we did on iOS was to reference two `TrackingClientType`s within `Koala` which allowed us to not change any of the callsites. When we whitelist the events we can then do that in `Koala` and test that it did not send certain events to the `DataLake` `TrackingClientType`.

But in that case I think it would mean referring to two `BackgroundService`s?

eoji: I think it's this on Android
`Callsite` -> `Koala` -> `TrackingClient` -> `BackgroundService` (handles Network)

PR #694, opened 2019-12-09
[📱] NT-690 Sending custom User Agent header with graphQL requests

eoji:
# 📲 What
Sending custom User Agent header with graphQL requests.

# 🤔 Why
Android pledges currently have a `null` backing `Channel`. The UA is used to determine the pledge came from a `phone`.

# 🛠 How
- Created `WebUtils.userAgent` and tests.
- Adds `User-Agent` header to `GraphQLInterceptor`,

# 👀 See
Screen Shot 2019-12-09 at 1 30 21 PM

# 📋 QA
View your logs in a debug build or verify that backings created from the app have the channel `phone`.

# Story 📖
[NT-690]

[NT-690]: https://dripsprint.atlassian.net/browse/NT-690

### DIFF
--- a/app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
+++ b/app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
@@ -1,17 +1,22 @@
 package com.kickstarter.services.interceptors
 
+import com.kickstarter.libs.Build
 import com.kickstarter.libs.CurrentUserType
 import okhttp3.Interceptor
 import okhttp3.Response
 
-class GraphQLInterceptor(private val clientId: String, private val currentUser: CurrentUserType) : Interceptor {
+class GraphQLInterceptor(private val clientId: String,
+                         private val currentUser: CurrentUserType,
+                         private val build: Build) : Interceptor {
     override fun intercept(chain: Interceptor.Chain?): Response {
         val original = chain!!.request()
         val builder = original.newBuilder().method(original.method(), original.body())
-        if (this.currentUser.exists()) {
-            builder.addHeader("Authorization", "token " + this.currentUser.accessToken)
+        val accessToken = this.currentUser.accessToken
+        if (accessToken != null) {
+            builder.addHeader("Authorization", "token $accessToken")
         }
-        builder.addHeader("X-KICKSTARTER-CLIENT", this.clientId)
+        builder.addHeader("User-Agent", WebRequestInterceptor.userAgent(this.build))
+                .addHeader("X-KICKSTARTER-CLIENT", this.clientId)
         return chain.proceed(builder.build())
     }
 }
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
```diff
@@ -1,17 +1,22 @@
...
         val original = chain!!.request()
         val builder = original.newBuilder().method(original.method(), original.body())
-        if (this.currentUser.exists()) {
-            builder.addHeader("Authorization", "token " + this.currentUser.accessToken)
+        val accessToken = this.currentUser.accessToken
+        if (accessToken != null) {
+            builder.addHeader("Authorization", "token $accessToken")
         }
-        builder.addHeader("X-KICKSTARTER-CLIENT", this.clientId)
+        builder.addHeader("User-Agent", WebRequestInterceptor.userAgent(this.build))
```
Setsun: Is the output value of `this.build` usually `Kickstarter Android Mobile Variant/internalDebug ...`?

If there is time, we could add a quick unit test to enforce that this is always passed in.

app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
```diff
@@ -1,17 +1,22 @@
...
-class GraphQLInterceptor(private val clientId: String, private val currentUser: CurrentUserType) : Interceptor {
+class GraphQLInterceptor(private val clientId: String,
+                         private val currentUser: CurrentUserType,
+                         private val build: Build) : Interceptor {
     override fun intercept(chain: Interceptor.Chain?): Response {
         val original = chain!!.request()
         val builder = original.newBuilder().method(original.method(), original.body())
-        if (this.currentUser.exists()) {
-            builder.addHeader("Authorization", "token " + this.currentUser.accessToken)
+        val accessToken = this.currentUser.accessToken
```
Setsun: We were checking for existence of `this.currentUser` previously, so could we run into a null-pointer exception by keying into `.accessToken`?

app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
eoji: ooh yeah I don't think there's any tests test for this! i can add one!

app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
eoji: i don't think so because `currentUser` is just an interface. and calling `.accessToken` is fetching a stored property 🤔 but i don't want to introduce anything so i'm going to revert it!

app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
eoji: so i wrote a test to test the value but to test the actual interceptor i'd need to add a new dependency https://github.com/square/okhttp/tree/master/mockwebserver which i think we should but not in this PR, i'm going to ask Jamie to file a separate JIRA ticket for it

Setsun: Ah gotcha. The test looks good, and mocking out the web server does seem like a separate task so I agree, LGTM 👍

PR #700, opened 2020-01-03
[❔] NT-601 Qualtrics survey support

eoji:
# 📲 What
Adding support for Qualtrics survey.

# 🤔 Why
So we can survey users!

# 🛠 How
- Added Qualtrics library.
- Added `QualtricsIntercept` for survey intercept values.
  - `QualtricsIntercept.id` returns the `testId` unless the app package is the prod package.
- Added `QualtricsResult` as a wrapper around `TargetingResult` so it's testable.
- Added `currentUser`, and `qualtricsPromptIsGone `, and `showQualtricsSurvey` outputs to `DiscoveryViewModel`.
- Added `qualtricsConfirmsClicked `, and `qualtricsDismissClicked `, and `qualtricsResult ` inputs to `DiscoveryViewModel`.
- The no service `Snackbar` won't appear over the Qualtrics prompt.

# 👀 See
| Prompt for logged in user | Opting in to survey | Opting out of survey |
| --- | --- | --- |
| device-2020-01-02-162611 2020-01-03 16_24_02 | device-2020-01-03-183639 2020-01-03 18_39_01 | device-2020-01-03-183707 2020-01-03 18_37_53 |

# 📋 QA
Log in on a device whose language is set to English.

# Story 📖
[NT-601]

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/DiscoveryActivity.java
+++ b/app/src/main/java/com/kickstarter/ui/activities/DiscoveryActivity.java
@@ -2,7 +2,6 @@ package com.kickstarter.ui.activities;
 
 import android.content.Intent;
 import android.graphics.drawable.Animatable;
-import android.net.Uri;
 import android.os.Bundle;
 import android.view.View;
 import android.widget.ImageButton;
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/kickstarter/ui/activities/DiscoveryActivity.java
```diff
@@ -213,6 +251,24 @@ public void onTabReselected(final TabLayout.Tab tab) {
     });
   }
 
+  private void setUpQualtrics(final @Nullable User user) {
```
justinswart: We chatted on Slack about just adding some tests for these values 👍

PR #702, opened 2020-01-06
[🛁] Adding new tracking session properties

eoji:
# 📲 What
Adding "Session Properties" tracking group.

# 🤔 Why
The first step of many for new tracking events.

# 🛠 How
- Added `abExperiments` property to `Config` model
- Added `ConfigUtils.currentVariants` to return a `JSONArray` of current experiments
- Added new session properties to `TrackingClientType`
- Implemented new session property values in `TrackingClient`
- Testzzzzz

# 👀 See
Nothing 2 c here

# 📋 QA
All events sent to the Lake 💧 should have these properties. This is a WIP so I'm on hitting staging.

# Story 📖
Part of [NT-654]

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/TrackingClientType.java
+++ b/app/src/main/java/com/kickstarter/libs/TrackingClientType.java
@@ -32,12 +32,11 @@ public abstract class TrackingClientType {
   private @NonNull Map<String, Object> sessionProperties(final boolean userIsLoggedIn) {
     final Map<String, Object> properties = new HashMap<String, Object>() {
       {
-        put("android_uuid", androidUUID());
         put("app_build_number", buildNumber());
         put("app_release_version", versionName());
         put("client_type", "native");
         put("current_variants", currentVariants());
-        put("device_distinct_id", androidUUID());
+        put("device_distinct_id", deviceDistinctId());
         put("device_format", deviceFormat());
         put("device_manufacturer", manufacturer());
         put("device_model", model());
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/kickstarter/libs/TrackingClientType.java
```diff
@@ -22,12 +24,38 @@ public final void track(final String eventName) {
...
-    hashMap.put("user_logged_in", userIsLoggedIn);
+    hashMap.putAll(sessionProperties(userIsLoggedIn));
 
     return hashMap;
   }
 
+  private @NonNull Map<String, Object> sessionProperties(final boolean userIsLoggedIn) {
+    final Map<String, Object> properties = new HashMap<String, Object>() {
+      {
+        put("android_uuid", androidUUID());
```
ifbarrera: On iOS we used to have `ios_uuid` but removed it since it's just the same value as `device_distinct_id`. You might be able to remove this prop as well.

app/src/test/java/com/kickstarter/libs/LakeTest.kt
```diff
@@ -44,12 +44,32 @@ class LakeTest : KSRobolectricTestCase() {
 
     private fun assertDefaultProperties(user: User?) {
```
ifbarrera: Should we consider renaming these tests to `assertSessionProperties` so that it's consistent with the `sessionProperties` function name?

eoji: so these will include all properties including the session, context, and user groups so i don't think `default` is right but not sure `session` is true either. maybe just `assertProperties`?

ifbarrera: Ahh gotcha. I think maybe default makes sense then.

eoji: so i renamed this method but left the test methods as `assertDefaultProperties`!

PR #704, opened 2020-01-08
[🛁] Adding new tracking discover properties

eoji:
# 📲 What
Adding "Discover Properties" tracking group.

# 🤔 Why
The next next step of many for new tracking events 😛

# 🛠 How
- Added new discover properties to TrackingClientType.
- Implemented new discover property values in TrackingClient.
- Testz

# 👀 See
nothing 2 c

# 📋 QA
All discovery events sent to the Lake 💧 should have these properties. This is a WIP so I'm only hitting staging.

# Story 📖
Part of [NT-654]

### DIFF
--- a/app/src/test/java/com/kickstarter/libs/KoalaTest.kt
+++ b/app/src/test/java/com/kickstarter/libs/KoalaTest.kt
@@ -56,21 +56,28 @@ class KoalaTest : KSRobolectricTestCase() {
         client.eventProperties.subscribe(this.propertiesTest)
         val koala = Koala(client)
 
-        val params = DiscoveryParams.builder().staffPicks(true).category(CategoryFactory.artCategory()).build()
+        val params = DiscoveryParams.builder()
+                .staffPicks(true)
+                .category(CategoryFactory.artCategory())
+                .build()
 
         koala.trackDiscovery(params, false)
 
         assertDefaultProperties(user)
         val expectedProperties = propertiesTest.value
         assertEquals(1L, expectedProperties["discover_category_id"])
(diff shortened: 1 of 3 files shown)


app/src/test/java/com/kickstarter/libs/KoalaTest.kt
```diff
@@ -56,21 +56,28 @@ class KoalaTest : KSRobolectricTestCase() {
...
+        assertEquals(null, expectedProperties["discover_search_term"])
         assertEquals(false, expectedProperties["discover_social"])
-        assertEquals(true, expectedProperties["discover_staff_picks"])
-        assertEquals(false, expectedProperties["discover_starred"])
-        assertEquals(null, expectedProperties["discover_term"])
-        assertEquals(false, expectedProperties["discover_everything"])
-        assertEquals(1, expectedProperties["page"])
-        assertEquals(15, expectedProperties["per_page"])
+        assertEquals(null, expectedProperties["discover_sort"])
+        assertNull(expectedProperties["discover_subcategory_id"])
```
ifbarrera: Should we also add a test for a subcategory to cover these props? 🤔

eoji: So I do test that in `LakeTest` here https://github.com/kickstarter/android-oss/pull/704/files#diff-47aba103d20cb8a1a0a18e91c468769bR140. I've only been fixing tests in `KoalaTest` that break, not adding new ones since I will delete the whole class when we're done.

ifbarrera: okay cool sounds good!

PR #715, opened 2020-01-16
[🆕] NT-790 Additional Qualtrics properties

eoji:
# 📲 What
Additional Qualtrics properties. 

# 🤔 Why
😢

# 🛠 How
- Added `first_session` `SharedPreference` that's passed through to Qualtrics as `first_app_session`.
- Added `{interceptID}_impression_count` property to Qualtrics.
- Added `logged_in` and `user_uid` as embedded data in the survey url.
- Added tests for survey url, impression count and first session. 

# 👀 See
Nothing 2 c

# 📋 QA
@jamielynnroth this can't be verified until we update the intercept logic 🙏 

# Story 📖
[NT-790]

[NT-790]: https://kickstarter.atlassian.net/browse/NT-790

### DIFF
--- a/app/src/main/java/com/kickstarter/models/QualtricsIntercept.kt
+++ b/app/src/main/java/com/kickstarter/models/QualtricsIntercept.kt
@@ -9,4 +9,8 @@ enum class QualtricsIntercept(private val prodId: String, private val testId: St
             else -> testId
         }
     }
+
+    fun impressionCountKey(packageName: String): String {
+        return id(packageName) + "_impression_count"
+    }
 }
--- a/app/src/main/java/com/kickstarter/ui/SharedPreferenceKey.java
+++ b/app/src/main/java/com/kickstarter/ui/SharedPreferenceKey.java
@@ -5,6 +5,7 @@ public final class SharedPreferenceKey {
 
   public static final String ACCESS_TOKEN = "access_token";
   public static final String CONFIG = "config";
+  public static final String FIRST_SESSION = "first_session";
   public static final String GO_REWARDLESS = "go_rewardless";
   public static final String HAS_SEEN_APP_RATING = "has_seen_app_rating";
   public static final String HAS_SEEN_GAMES_NEWSLETTER = "has_seen_games_newsletter";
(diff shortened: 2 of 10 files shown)


app/src/main/java/com/kickstarter/models/QualtricsIntercept.kt
```diff
@@ -9,4 +9,8 @@ enum class QualtricsIntercept(private val prodId: String, private val testId: St
             else -> testId
         }
     }
+
+    fun impressionCountKey(packageName: String): String {
```
justinswart: I was under the _impression_ (lol) that we didn't need the impression count anymore if we're confident that Qualtrics won't prompt us once we've taken the survey? Anyway, suppose no real harm in keeping it unless you wanted to simplify your code without it.

app/src/main/java/com/kickstarter/ui/SharedPreferenceKey.java
```diff
@@ -5,6 +5,7 @@ private SharedPreferenceKey() {}
 
   public static final String ACCESS_TOKEN = "access_token";
   public static final String CONFIG = "config";
+  public static final String FIRST_SESSION = "first_session";
```
justinswart: Do you need to persist this outside of Qualtrics? I figured since you're able to get the value stored in Qualtrics' properties you could check if there is already something there otherwise set it (for the first time user thing).

app/src/main/java/com/kickstarter/models/QualtricsIntercept.kt
eoji: This count is so we don't show users the prompt more than 3 times. Otherwise, we'd be hounding them until they took the survey.

app/src/main/java/com/kickstarter/ui/SharedPreferenceKey.java
eoji: I learned the hard way that Qualtrics calls the initialization block more than once 😒 so that's why I'm moved it here

eoji:
Tested and it seems to work woo woooo!!

PR #724, opened 2020-01-21
[ℹ️] NT-813 Updating "partition-key" value in Data Lake events

eoji:
# 📲 What
Updating "partition-key" value in Data Lake events

# 🤔 Why
To help with session tracking.

# 🛠 How
- When user is logged **in** the value of `partition-key` is `user.id`.
- When user is logged **out** the value of `partition-key` is `deviceDistinctId` that returns a non-null `String`.

# 👀 See
No visuals~

# 📋 QA
tail your device's events in `ktk`

# Story 📖
[NT-813]

[NT-813]: https://kickstarter.atlassian.net/browse/NT-813

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
@@ -9,7 +9,6 @@ import com.kickstarter.services.LakeBackgroundService
 import com.kickstarter.ui.IntentKey
 import org.json.JSONException
 import org.json.JSONObject
-import java.util.*
 
 class LakeTrackingClient(
         @param:ApplicationContext private val context: Context,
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/kickstarter/libs/LakeTrackingClient.kt
```diff
@@ -51,7 +50,7 @@ class LakeTrackingClient(
         data.put("properties", propertiesJSON)
 
         val record = JSONObject()
-        record.put("partition-key", UUID.randomUUID().toString())
+        record.put("partition-key", this.loggedInUser?.id() ?: deviceDistinctId())
```
eoji: @amyfain does it matter that this will be a number for logged in users and a string for logged out users?

amyfain:
Looks good to me! 🎉

PR #730, opened 2020-02-03
[✅] Fixing add new cards tests

eoji:
# 📲 What
Fixing `NewCardFragmentViewModelTest` validation tests.

# 🤔 Why
So the tests pass.

# 🛠 How
Added `futureYear` that always returns a year in the future so the expiration date is valid.

# 👀 See
Testz only

# 📋 QA
CircleCI 🙏 

# Story 📖
N/A


eoji:
We do, we set it to the current time though 🤔

PR #734, opened 2020-02-04
[💧] NT-815 Hitting prod data Lake in prod

eoji:
# 📲 What
Hitting prod data Lake when hitting prod API.

# 🤔 Why
We're prepping for 2.2.0 release and need to start hitting prod Data Lake so events get sent to the right Amplitude project.

# 🛠 How
- Removed comment and now return `Secrets.LakeEndpoint.PRODUCTION` if the API endpoint is production in `ApplicationModule.provideLakeEndpoint`

# 👀 See
Nothing 2 c

# 📋 QA
Tail the prod data lake!

# Story 📖
[NT-815]

[NT-815]: https://kickstarter.atlassian.net/browse/NT-815


eoji:
😓@ifbarrera, we don't _currently_ test our network calls but we do have a ticket for it in the backlog because it's a bit of work to set up all of the mock servers https://github.com/kickstarter/android-oss/pull/694#discussion_r355704905

PR #736, opened 2020-02-06
[🔙] NT-843 Project Page Pledge Button Clicked event

eoji:
# 📲 What
Adding `Project Page Pledge Button Clicked` event.

# 🤔 Why
We have new Back a Project events.

# 🛠 How
- Added `LakeEvent.PROJECT_PAGE_PLEDGE_BUTTON_CLICKED` with value `"Project Page Pledge Button Clicked"`
- Added `Koala.trackProjectPagePledgeButtonClicked`
- Tracking when a user clicks the `"Back this project"` button in `ProjectViewModel`
- Added tests in `ProjectViewModelTest`

# 👀 See
No visual changes.

# 📋 QA
So many ways 2 QA:
- `ktk` the staging lake
- check the `Logcat` in Android Studio
- Look at the `dev` project in Amplitude

# Story 📖
[NT-843]

[NT-843]: https://kickstarter.atlassian.net/browse/NT-843


app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.kt
```diff
@@ -708,6 +708,44 @@ interface ProjectViewModel {
                     .subscribe(this.heartDrawableId)
 
             //Tracking
+            val refTagsAndProject = Observable.combineLatest<RefTag, RefTag, Project, RefTagsAndProject>(refTag, cookieRefTag, currentProject)
+            { refTagFromIntent, refTagFromCookie, project -> RefTagsAndProject(refTagFromIntent, refTagFromCookie, project) }
+                    .filter { it.project.hasRewards() }
+                    .take(1)
```
justinswart: Do we need the `take(1)` with the `takeWhen` further down?

eoji: I think so because `currentProject` emits again when the `Project` gets refreshed and we don't want to track this screen view again for those refreshes.

justinswart: Oh right - I guess in this case it wouldn't matter because the event doesn't track based on the screen refreshing but on the button press. Either way 🤷‍♂

PR #752, opened 2020-02-18
[⏳] NT-890 Syncing Optimizely every 10 minutes

eoji:
# 📲 What
Adding datafile and event dispatch intervals to Optimizely.

# 🤔 Why
So we get the latest data file and events dispatched every 10 minutes.

# 🛠 How
Added datafile and event dispatch intervals to Optimizely client with a value of 10 minutes per example: https://docs.developers.optimizely.com/full-stack/docs/initialize-sdk-android#section-example
 
# 👀 See
Nothing 2 c

# 📋 QA
Change your variant and wait...10 minutes.

# Story 📖
[NT-890]

[NT-890]: https://kickstarter.atlassian.net/browse/NT-890

### DIFF
--- a/app/src/main/java/com/kickstarter/ApplicationModule.java
+++ b/app/src/main/java/com/kickstarter/ApplicationModule.java
@@ -9,6 +9,9 @@ import android.content.res.AssetManager;
 import android.content.res.Resources;
 import android.util.Log;
 
+import androidx.annotation.NonNull;
+import androidx.preference.PreferenceManager;
+
 import com.apollographql.apollo.ApolloClient;
 import com.crashlytics.android.Crashlytics;
 import com.google.gson.FieldNamingPolicy;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/kickstarter/ApplicationModule.java
```diff
@@ -649,6 +650,8 @@ ExperimentsClientType provideOptimizely(final @ApplicationContext @NonNull Conte
 
     final OptimizelyManager optimizelyManager = OptimizelyManager.builder()
       .withSDKKey(optimizelyKey)
+      .withEventDispatchInterval(60L * 10L)
```
ifbarrera: What's the default event dispatch interval? 10 minutes seems high, I'd imagine we want to dispatch events fairly frequently?

eoji: it's `-1` lol
```        // -1 will cause the background download to not be initiated.
        private long datafileDownloadInterval = -1L;
        // -1 will cause the background download to not be initiated.
        private long eventDispatchInterval = -1L;```

PR #754, opened 2020-02-20
[🧪] NT-888 Campaign details experiment on Campaign details page

eoji:
# 📲 What
Displaying native_project_page_campaign_details variants on live and not backed Campaign details screen.

# 🤔 Why
🧬

# 🛠 How
- Added `ActivityRequestCodes.SHOW_REWARDS` when starting `CampaignDetailsActivity`
- Added `IntentKey.PROJECT_DATA` for passing `ProjectData` to started Activity
- `CampaignDetails`
  - `CampaignDetailsActivity` is now started with `ProjectData`
  - Added faux pledge container `campaign_details_pledge_container`
  - `CampaignDetailsViewModel`
    - Added `pledgeActionButtonClicked` input that is called when the user clicks the pledge action button.
    - Added `goBackToProject` output that emits when we should dismiss this screen and show the rewards.
    - Added `pledgeContainerIsVisible` output that emits a `Boolean` determining if the faux pledge container should be visible.
    - Tests
- `Project`
  - `ProjectActivity.startCampaignWebViewActivity` now takes in a `[ProjectData]` and starts `CampaignDetailsActivity` for a result of `SHOW_REWARDS`
  - `ProjectViewModel`
    - `startCampaignWebViewActivity` output now emits `ProjectData`
    - `expandPledgeSheet` emits when an `OK` `SHOW_REWARDS` result is received.
    - Tests

# 👀 See
| Variant 2 🐛 | Variant 1 🐛 | Control 🐛 |
| --- | --- | --- |
| screenshot-2020-02-20_152913 | screenshot-2020-02-20_152903 | screenshot-2020-02-20_152903 |
| device-2020-02-20-134307 2020-02-20 13_46_55 |

# 📋 QA
Get yourself into variant-2 of `native_project_page_campaign_details` and check out the campaign details!
The faux pledge container should only be visible on live and unbacked projects.

# Story 📖
[NT-888]

[NT-888]: https://kickstarter.atlassian.net/browse/NT-888

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/ActivityRequestCodes.java
+++ b/app/src/main/java/com/kickstarter/libs/ActivityRequestCodes.java
@@ -8,4 +8,5 @@ public final class ActivityRequestCodes {
   public final static int CHECKOUT_ACTIVITY_WALLET_CHANGE_REQUEST = 592_11;
   public final static int CHECKOUT_ACTIVITY_WALLET_OBTAINED_FULL = 592_12;
   public final static int SAVE_NEW_PAYMENT_METHOD = 592_13;
+  public final static int SHOW_REWARDS = 592_14;
 }
--- a/app/src/main/java/com/kickstarter/viewmodels/CampaignDetailsViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/CampaignDetailsViewModel.kt
@@ -1,42 +1,79 @@
 package com.kickstarter.viewmodels
 
-import androidx.annotation.NonNull
+import android.util.Pair
 import com.kickstarter.libs.ActivityViewModel
 import com.kickstarter.libs.Environment
-import com.kickstarter.models.Project
+import com.kickstarter.libs.models.OptimizelyExperiment
+import com.kickstarter.libs.rx.transformers.Transformers.combineLatestPair
+import com.kickstarter.models.User
 import com.kickstarter.ui.IntentKey
 import com.kickstarter.ui.activities.CampaignDetailsActivity
+import com.kickstarter.ui.data.ProjectData
 import rx.Observable
 import rx.subjects.BehaviorSubject
+import rx.subjects.PublishSubject
 
 interface CampaignDetailsViewModel {
 
-    interface Inputs
+    interface Inputs {
+        /** Call when the user clicks the pledge action button. */
+        fun pledgeActionButtonClicked()
+    }
 
     interface Outputs {
+        /** Emits when we should return to the [com.kickstarter.ui.activities.ProjectActivity] with the rewards visible. */
+        fun goBackToProject(): Observable<Void>
+
+        /** Emits a boolean determining if the faux pledge container is visible. */
+        fun pledgeContainerIsVisible(): Observable<Boolean>
+
         /** Emits the URL of the campaign to load in the web view. */
         fun url(): Observable<String>
     }
 
     class ViewModel(environment: Environment) : ActivityViewModel<CampaignDetailsActivity>(environment), Inputs, Outputs {
+        private val currentUser = environment.currentUser()
+        private val optimizely = environment.optimizely()
 
+        private val pledgeButtonClicked = PublishSubject.create<Void>()
+
+        private val goBackToProject = PublishSubject.create<Void>()
+        private val pledgeContainerIsVisible = BehaviorSubject.create<Boolean>()
         private val url = BehaviorSubject.create<String>()
 
         val inputs: Inputs = this
         val outputs: Outputs = this
 
         init {
-            val project = intent()
-                    .map { it.getParcelableExtra(IntentKey.PROJECT) as Project? }
-                    .ofType(Project::class.java)
+            val projectData = intent()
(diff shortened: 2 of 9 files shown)


app/src/main/java/com/kickstarter/libs/ActivityRequestCodes.java
```diff
@@ -8,4 +8,5 @@ private ActivityRequestCodes() {}
   public final static int CHECKOUT_ACTIVITY_WALLET_CHANGE_REQUEST = 592_11;
   public final static int CHECKOUT_ACTIVITY_WALLET_OBTAINED_FULL = 592_12;
   public final static int SAVE_NEW_PAYMENT_METHOD = 592_13;
+  public final static int SHOW_REWARDS = 592_14;
```
justinswart: What's a request code?

app/src/main/java/com/kickstarter/viewmodels/CampaignDetailsViewModel.kt
```diff
@@ -1,42 +1,79 @@
...
+        /** Call when the user clicks the pledge action button. */
+        fun pledgeActionButtonClicked()
+    }
 
     interface Outputs {
+        /** Emits when we should return to the [com.kickstarter.ui.activities.ProjectActivity] with the rewards visible. */
+        fun goBackToProject(): Observable<Void>
+
+        /** Emits a boolean determining if the faux pledge container is visible. */
+        fun pledgeContainerIsVisible(): Observable<Boolean>
```
justinswart: Ooh, not isGone? 😄

app/src/test/java/com/kickstarter/viewmodels/CampaignDetailsViewModelTest.kt
```diff
@@ -3,24 +3,98 @@ package com.kickstarter.viewmodels
...
+    @Test
+    fun testGoBackToProject() {
+        setUpEnvironment(environment(), ProjectDataFactory.project(ProjectFactory.project()))
+
+        this.vm.inputs.pledgeActionButtonClicked()
+        this.goBackToProject.assertValueCount(1)
+    }
+
+    @Test
+    fun testPledgeContainerIsVisible_whenProjectIsLiveNotBacked_control() {
```
justinswart: Couple questions about these tests:

1) This test appears to be testing the container visibility but asserts after `this.vm.inputs.pledgeActionButtonClicked()`, does the button click affect its visibility?
2) What do you think about `testPledgeContainerIsNotVisible` or `testPledgeContainerVisibility`? Usually I read the test name almost as an assertion that it's testing that _this thing is this_, so in this case I initially thought it was testing that the container _is_ visible whereas it's actually the opposite?

app/src/test/java/com/kickstarter/viewmodels/CampaignDetailsViewModelTest.kt
```diff
@@ -3,24 +3,98 @@ package com.kickstarter.viewmodels
...
+                })
+                .build()
+        setUpEnvironment(environment, ProjectDataFactory.project(ProjectFactory.project()))
+
+        this.vm.inputs.pledgeActionButtonClicked()
+        this.pledgeContainerIsVisible.assertValue(true)
+    }
+
+    @Test
+    fun testPledgeContainerIsVisible_whenProjectIsNotLive() {
```
justinswart: For these last two I'm guessing the container is not visible by default but not sure if we want to explicitly emit `false` in these cases 🤷‍♂

app/src/main/java/com/kickstarter/libs/ActivityRequestCodes.java
eoji: If you want to get a result back from a screen you started, you need to pass in a request code `Int`. https://developer.android.com/training/basics/intents/result#StartActivity. So if I want the `ProjectActivity` to know that the `CampaignDetailsActivity`'s pledge action button was clicked, I need to use a unique request code.

app/src/test/java/com/kickstarter/viewmodels/CampaignDetailsViewModelTest.kt
eoji: 1. No, it's copy pasta I need to update 😅 
2. We've been trying to name tests `testOutputName_whenCondition`

app/src/test/java/com/kickstarter/viewmodels/CampaignDetailsViewModelTest.kt
eoji: It is initially hidden but I can make it emit in those cases 👍

app/src/main/java/com/kickstarter/viewmodels/CampaignDetailsViewModel.kt
eoji: I think `isVisible` is less confusing but should I do `isGone` to be consistently confusing?

justinswart: Yeah, in our case we just match it to the property it's setting which is `isHidden`. Whatever works!

PR #773, opened 2020-03-02
[0️⃣] NT-949 Disabling updates row when there are no updates 

eoji:
# 📲 What
Disabling project updates container in Project page when there are no updates.

# 🤔 Why
If there are no updates, there's nothing 2 c.

# 🛠 How
- Disabling updates row in ProjectViewHolder when a project doesn't have any updates.
- Added helper method `hasUpdates` to `Project` model.
- Added output `updatesContainerIsEnabled` to `ProjectHolderViewModel` that emits a `Boolean` determining if the updates row should be enabled
- Cleaned up unnecessary dividers and strange padding in `project_creator_info`
- Tests for `updatesContainerIsEnabled`

# 👀 See
| After 🦋 | Before 🐛 |
| --- | --- |
| screenshot-2020-03-02_113711 | screenshot-2020-03-02_113415 |

# 📋 QA
Click on the updates row.

# Story 📖
[NT-949]

[NT-949]: https://kickstarter.atlassian.net/browse/NT-949

### DIFF
--- a/app/src/main/res/layout/project_creator_info.xml
+++ b/app/src/main/res/layout/project_creator_info.xml
@@ -19,8 +19,8 @@
       android:id="@+id/campaign"
       android:layout_width="match_parent"
       android:layout_height="wrap_content"
-      android:background="@drawable/click_indicator_light_masked"
       android:focusable="true"
+      android:foreground="@drawable/click_indicator_light_masked"
       android:paddingTop="@dimen/grid_2"
       android:paddingBottom="@dimen/grid_2">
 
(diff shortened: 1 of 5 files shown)


app/src/main/res/layout/project_creator_info.xml
```diff
@@ -1,123 +1,112 @@
...
-        android:layout_alignParentEnd="true"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        tools:text="1,000"/>
+        android:layout_alignParentEnd="true"
+        android:layout_centerVertical="true"
+        android:background="@drawable/text_view_rect_background"
+        android:textColor="@color/ksr_soft_black"
+        android:textSize="@dimen/body"
+        tools:text="1,000" />
```
jgsamudio: Is this just a temporary value to preview the design?

eoji: yep! any properties with the `tools` namespace are only visible when previewing the layout

jgsamudio: Awesome! 👍

PR #775, opened 2020-03-02
[␡] NT-897 Removing go rewardless feature flag

eoji:
# 📲 What
Removing go rewardless feature flag but keeping `Editorial` support.

# 🤔 Why
Because the feature is over!

# 🛠 How
Removing `android_go_rewardless_2` feature flag support:
- Removed `FeatureKey.ANDROID_GO_REWARDLESS`
- Removed `SharedPreferenceKey.GO_REWARDLESS`
- Removed go rewardless preference injection from `ApplicationModule`
## `FeatureFlagsActivity`
- Removing local toggle in `FeatureFlagsActivity`
- Removed `go_rewardless` row in `activity_feature_flags`
- Removed `Go_Rewardless` string.
## `Environment`
- Removed `environment.goRewardlessPreference`
## `NativeCheckoutRewardViewHolder`
- Removed `background` output from `NativeCheckoutRewardViewHolderViewModel`
- Deleted `bg_go_rewardless`
- Removed `reward_contents` id from `item_reward`
## `PledgeFragmentViewModel`
- `rewardTitle` for no reward or a reward with a `null` title now emits `Back it because you believe in it.`
## Tests
- Updated `ConfigUtilsTest`
- Removed go rewardless checks in `DiscoveryFragmentViewModelTest` and `NativeCheckoutRewardViewHolderViewModelTest`
- Updated title tests in `PledgeFragmentViewModelTest`

# 👀 See
| No reward unbacked  | No reward backed  | No reward pledge screen  |
| --- | --- |--- |
| screenshot-2020-03-02_170521 | screenshot-2020-03-02_170557 | screenshot-2020-03-02_170525 |

# 📋 QA
The go rewardless feature should not show up even if its turned on.

# Story 📖
[NT-897]

[NT-897]: https://kickstarter.atlassian.net/browse/NT-897


app/src/test/java/com/kickstarter/libs/utils/ConfigUtilsTest.kt
```diff
@@ -31,21 +31,15 @@ class ConfigUtilsTest : KSRobolectricTestCase() {
 
         assertEquals(JSONArray(), ConfigUtils.enabledFeatureFlags(ConfigFactory.configWithFeatureEnabled("ios_native_checkout")))
 
-        assertEquals(JSONArray().apply { put("android_go_rewardless_2") },
-                ConfigUtils.enabledFeatureFlags(ConfigFactory.configWithFeatureEnabled(FeatureKey.ANDROID_GO_REWARDLESS)))
+        assertEquals(JSONArray(),
+                ConfigUtils.enabledFeatureFlags(ConfigFactory.configWithFeaturesEnabled(mapOf(Pair(FeatureKey.ANDROID_NATIVE_CHECKOUT, false),
+                        Pair("ios_go_rewardless", true),
```
jgsamudio: Do we want to use the ios feature flags?

eoji: This test is making sure we don't return any feature flags that start with `ios_`, only the ones that start with `android_`. I could rename it to `ios_feature_1`, I just used a name we were familiar with. What do you think?

jgsamudio: Ahh, thank you for clarifying! You can do something like `ios_feature_that_wishes_it_was_android` lol. You can keep it as is since it's just a test.

PR #781, opened 2020-03-06
[🔢] NT-929 Adding ProjectCreatorDetails query

eoji:
# 📲 What
Adding `ProjectCreatorDetails` query and corresponding outputs in `ProjectHolderViewModel`.

# 🤔 Why
Laying the foundation for the `native_project_page_conversion_creator_details` experiment.

# 🛠 How
- Added `project.graphql` with `ProjectCreatorDetails` query.
- Added `creatorDetails` call to `ApolloClientType` interface and implemented in subtypes that returns a `CreatorDetails` object.
- And `creatorBackedProjectsCount` and `creatorLaunchedProjectsCount` outputs to `ProjectHolderViewModel`.
- Added `CreatorDetailsFactory` for tests.
- Added tests to `ProjectHolderViewModelTest`.

# 👀 See
Nothing 2 c

# 📋 QA
Does the code look legit?

# Story 📖
[NT-929]

[NT-929]: https://kickstarter.atlassian.net/browse/NT-929

### DIFF
--- a/app/src/main/java/com/kickstarter/mock/factories/CreatorDetailsFactory.kt
+++ b/app/src/main/java/com/kickstarter/mock/factories/CreatorDetailsFactory.kt
@@ -8,7 +8,6 @@ class CreatorDetailsFactory private constructor() {
             return CreatorDetails.builder()
                     .backingsCount(3)
                     .launchedProjectsCount(2)
-                    .name("Creator Name")
                     .build()
         }
     }
--- a/app/src/main/java/com/kickstarter/mock/factories/CreatorDetailsFactory.kt
+++ b/app/src/main/java/com/kickstarter/mock/factories/CreatorDetailsFactory.kt
@@ -1,0 +1,15 @@
+package com.kickstarter.mock.factories
+
+import com.kickstarter.models.CreatorDetails
+
+class CreatorDetailsFactory private constructor() {
+    companion object {
+        fun creatorDetails(): CreatorDetails {
+            return CreatorDetails.builder()
+                    .backingsCount(3)
+                    .launchedProjectsCount(2)
+                    .name("Creator Name")
+                    .build()
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/kickstarter/mock/factories/CreatorDetailsFactory.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import com.kickstarter.models.CreatorDetails
+
+class CreatorDetailsFactory private constructor() {
+    companion object {
+        fun creatorDetails(): CreatorDetails {
+            return CreatorDetails.builder()
+                    .backingsCount(3)
+                    .launchedProjectsCount(2)
+                    .name("Creator Name")
```
eoji: I actually don't use the name and maybe I should drop it?

eoji: I dropped it!

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
```diff
@@ -151,6 +152,34 @@ class KSApolloClient(val service: ApolloClient) : ApolloClientType {
...
+
+                        override fun onResponse(response: Response<ProjectCreatorDetailsQuery.Data>) {
+                            if (response.hasErrors()) {
+                                ps.onError(Exception(response.errors().first().message()))
+                            }
+
+                            response.data()?.project()?.creator()?.let {
+                                ps.onNext(CreatorDetails.builder()
+                                        .backingsCount(it.backingsCount())
+                                        .launchedProjectsCount(it.launchedProjects()?.totalCount() ?: 0)
```
justinswart: Makes sense to coalesce to `0` just noting that this _should_ only ever be 1 or more given that a creator shouldn't get into this state without having created a project 😄.

PR #785, opened 2020-03-11
[🔮] NT-985 Decreasing Optimizely event dispatch interval

eoji:
# 📲 What
Decreasing Optimizely event dispatch interval from 10 seconds to 2 seconds.

# 🤔 Why
We are playing telephone with Optimizely events coming from both the client and the backend and there is a race condition if a user completes a checkout on the backend before the client side events.

# 🛠 How
~`60L`~ `2L`

# 👀 See
There are no visual changes!

# 📋 QA
Smash that pledge button.

# Story 📖
[NT-985]

[NT-985]: https://kickstarter.atlassian.net/browse/NT-985

### DIFF
--- a/app/src/main/java/com/kickstarter/ApplicationModule.java
+++ b/app/src/main/java/com/kickstarter/ApplicationModule.java
@@ -639,7 +639,7 @@ public final class ApplicationModule {
     final OptimizelyManager optimizelyManager = OptimizelyManager.builder()
       .withSDKKey(optimizelyKey)
       .withDatafileDownloadInterval(60L * 10L)
-      .withEventDispatchInterval(60L)
+      .withEventDispatchInterval(2L)
       .build(context);
 
     optimizelyManager.initialize(context, null, optimizely -> {


app/src/main/java/com/kickstarter/ApplicationModule.java
```diff
@@ -639,7 +639,7 @@ ExperimentsClientType provideOptimizely(final @ApplicationContext @NonNull Conte
     final OptimizelyManager optimizelyManager = OptimizelyManager.builder()
       .withSDKKey(optimizelyKey)
       .withDatafileDownloadInterval(60L * 10L)
-      .withEventDispatchInterval(60L)
+      .withEventDispatchInterval(2L)
```
pcreux: Cool! Is the goal of the event dispatch interval to dispatch events by batch?

eoji: Yeah, that's the way `Optimizely X` for Android handles dispatching events https://docs.developers.optimizely.com/full-stack/docs/configure-event-dispatcher-android

PR #786, opened 2020-03-11
[🛤] NT-936 Creator Details Clicked event

eoji:
# 📲 What
Added `Creator Details Clicked` event for Optimizely.

# 🤔 Why
So we can get those sweet funnel numbers.

# 🛠 How
- Added `OptimizelyEvent.CREATOR_DETAILS_CLICKED` with value `"Creator Details Clicked"`
- Tracking `CREATOR_DETAILS_CLICKED` when the creator details are clicked in `ProjectViewModel`.
- Added tests to assert `CREATOR_DETAILS_CLICKED` is tracked when the button is clicked on live and not backed projects only.

# 👀 See
Nothing to see.

# 📋 QA
Check your local logs!
Check your data dog logs!

# Story 📖
[NT-936]

[NT-936]: https://kickstarter.atlassian.net/browse/NT-936


app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.kt
```diff
@@ -834,13 +832,18 @@ interface ProjectViewModel {
...
+            fullProjectDataAndCurrentUser
                     .compose<Pair<Pair<ProjectData, User?>, Boolean>>(combineLatestPair(shouldTrackCTAClickedEvent))
                     .filter { it.second }
                     .map { it.first }
                     .compose(bindToLifecycle())
                     .subscribe { this.optimizely.track(PROJECT_PAGE_PLEDGE_BUTTON_CLICKED, it.second, it.first.refTagFromIntent()) }
+
+            fullProjectDataAndCurrentUser
+                    .compose<Pair<ProjectData, User?>>(takeWhen(creatorInfoClicked))
+                    .filter { it.first.project().isLive && !it.first.project().isBacking }
```
ifbarrera: Do we not show the creator info if you've already backed the project? 🤔

eoji: This is only for tracking. I'm only tracking this for live not backed projects. Otherwise, I think the experiment would have really low conversion because it would fire for ended and ? But not sure because the experiment process has been pretty confusing.

ifbarrera: Ah ok that makes sense!

PR #793, opened 2020-03-16
[␡] NT-965 Removed `FeatureKey.ANDROID_NATIVE_CHECKOUT`

eoji:
# 📲 What
Removed `FeatureKey.ANDROID_NATIVE_CHECKOUT`

# 🤔 Why
Feature flag is over.

# 🛠 How
- Updated `ConfigUtilsTest` that used key.

# 👀 See
N/A

# 📋 QA
N/A

# Story 📖
[NT-965]

[NT-965]: https://kickstarter.atlassian.net/browse/NT-965

### DIFF
--- a/app/src/test/java/com/kickstarter/libs/utils/ConfigUtilsTest.kt
+++ b/app/src/test/java/com/kickstarter/libs/utils/ConfigUtilsTest.kt
@@ -1,7 +1,6 @@
 package com.kickstarter.libs.utils
 
 import com.kickstarter.KSRobolectricTestCase
-import com.kickstarter.libs.FeatureKey
 import com.kickstarter.mock.factories.ConfigFactory
 import org.json.JSONArray
 import org.junit.Test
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/kickstarter/libs/utils/ConfigUtilsTest.kt
```diff
@@ -32,13 +31,13 @@ class ConfigUtilsTest : KSRobolectricTestCase() {
         assertEquals(JSONArray(), ConfigUtils.enabledFeatureFlags(ConfigFactory.configWithFeatureEnabled("ios_native_checkout")))
 
         assertEquals(JSONArray(),
-                ConfigUtils.enabledFeatureFlags(ConfigFactory.configWithFeaturesEnabled(mapOf(Pair(FeatureKey.ANDROID_NATIVE_CHECKOUT, false),
```
jgsamudio: Was this feature key only used in tests? It might be a good idea to extract the checkout string

eoji: Yep, it was only used in these tests since now it's no longer used in the actual app

PR #795, opened 2020-03-18
[🆕] NT-996 Replacing tracking BackgroundServices with Workers

eoji:
# 📲 What
Replacing tracking `BackgroundService`s with `Worker`s

# 🤔 Why

# 🛠 How
## `TrackingClient`
- Removed abstract methods `backgroundServiceClass`, `eventKey`, `eventNameKey`, `tag` from `TrackingClient` and it's subclasses.
## `TrackingClientType`
- Added `Type` enum with values `KOALA` and `LAKE`
- Added abstract method `type` to return `Type` enum
- Renamed `cleanProperties` to `lakeProperties`
- Renamed `defaultProperties` to `koalaProperties`
## `TrackingWorker`
- Created base abstract class `TrackingWorker` because both of the background workers for tracking are very similar.
  - 4XX responses are considered hard failures and won't be retried
- Replaced `KoalaBackgroundService` with `KoalaWorker`
- Replaced `LakeBackgroundService` with `LakeWorker`
- Removed both services from `AndroidManifest`
## `IntentKey`
- Replaced `KOALA_EVENT`, `KOALA_EVENT_NAME`, `LAKE_EVENT`, and `LAKE_EVENT_NAME` with `EVENT_DATA` and `EVENT_NAME`
- Added `TRACKING_CLIENT_TYPE_TAG`
## Tests
- Initializing `MockTrackingClient` with helper method instead of repeating myself

# 👀 See
No UI changes but peep the logs:
Screen Shot 2020-03-18 at 6 12 16 PM

# 📋 QA
So many ways 2 QA:
- `ktk` the staging lake
- check the `Logcat` in Android Studio
- Look at the `dev` project in Amplitude

# Story 📖
[NT-996]

[NT-996]: https://kickstarter.atlassian.net/browse/NT-996


app/src/main/java/com/kickstarter/services/TrackingWorker.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    private fun logResponse() {
+        if (this.build.isDebug) {
+            Timber.d("Successfully tracked $tag event: $eventName")
+        }
+        Crashlytics.log(this.eventName)
+    }
+
+    private fun logTrackingError(code: Int) {
+        val errorMessage = "$code Failed to track $tag event: $eventName"
```
eoji: @yatriks I added the HTTP code to the log!

PR #796, opened 2020-03-20
[📲] NT-997 NT-998 Replacing push token Services with Workers

eoji:
# 📲 What
Replacing push token `Service`s with `Worker`s

# 🤔 Why

# 🛠 How

## `DeviceRegistrar`
- Converted this file to Kotlin to take advantage of the language
- Added `registerToken` method to schedule some work to register a push token
- Added `unregisterToken` method to schedule some work to unregister the current push token
## `ApiService`
- `registerPushToken` returns a `JsonObject` instead of an `Empty` so we can properly handle failures
## New `TokenWorker`s
- Replaced `RegisterService` with `RegisterTokenWorker`
  - 4XX responses are considered hard failures and won't be retried
- Replaced `UnregisterService` with `UnregisterTokenWorker`
- Removed both services from `AndroidManifest`
## `TokenListenerService`
- Calls `DeviceRegistrar.registerToken` when a new token is received instead of starting the `RegisterService`
## `IntentKey`
- Added `PUSH_TOKEN` as key for token string

# 👀 See
No UI changes but peep the logs after logging in and out:
Screen Shot 2020-03-20 at 7 03 55 PM

# 📋 QA
Log in (on prod, we don't send pushes to staging) and test that push notifications are still received!

# Story 📖
[NT-997]
[NT-998]

[NT-997]: https://kickstarter.atlassian.net/browse/NT-997


app/src/main/java/com/kickstarter/services/firebase/RegisterTokenWorker.kt
```diff
@@ -0,0 +1,80 @@
...
+        } else {
+            try {
+                val errorEnvelope = this.gson.fromJson(response, ErrorEnvelope::class.java)
+                logError(errorEnvelope)
+                when (errorEnvelope.httpCode()) {
+                    in 400..499 -> Result.failure()
+                    else -> Result.retry()
+                }
+            } catch (exception: JsonSyntaxException) {
+                Result.failure()
```
eoji: Let's log if this deserialization fails

app/src/main/java/com/kickstarter/libs/DeviceRegistrar.kt
```diff
@@ -0,0 +1,65 @@
...
+
+class DeviceRegistrar(private val playServicesCapability: PlayServicesCapability,
+                      @param:ApplicationContext @field:ApplicationContext private val context: Context) : DeviceRegistrarType {
+
+    /**
+     * If Play Services is available on this device, start a service to register it in the backend.
+     */
+    override fun registerDevice() {
+        if (this.playServicesCapability.isCapable) {
+            FirebaseInstanceId.getInstance().instanceId.addOnSuccessListener{
```
jgsamudio:
```suggestion
            FirebaseInstanceId.getInstance().instanceId.addOnSuccessListener {
```

app/src/main/java/com/kickstarter/libs/DeviceRegistrar.kt
```diff
@@ -0,0 +1,65 @@
...
+     */
+    override fun unregisterDevice() {
+        if (this.playServicesCapability.isCapable) {
+            unregisterToken()
+        }
+    }
+
+    private fun unregisterToken() {
+        val request = OneTimeWorkRequestBuilder<UnregisterTokenWorker>()
+                .setBackoffCriteria(BackoffPolicy.EXPONENTIAL, 30, TimeUnit.SECONDS)
```
jgsamudio: Can we have a constant for 30 here and the call below?

PR #797, opened 2020-03-25
[🤑] NT-1024 App Completed Checkout event

eoji:
# 📲 What
Tracking `App Completed Checkout` event on the Thanks page

# 🤔 Why
So we can track revenue from experiments!

# 🛠 How
## `ExperimentUtils`
- Added `ExperimentRevenueData` data class to hold models needed for checkout tags
- Added `checkoutTags` method to return map of checkout properties
  - Added tests for `checkoutTags` for US and non US project
## `ExperimentsClientType`
- Added `trackRevenue` method that takes in event key and `ExperimentRevenueData`
  - Implemented in `OptimizelyExperimentsClient` and `MockExperimentsClientType`
## `MockExperimentsClientType`
- Added `tags` field to `OptimizelyEvent` to helper `ExperimentsEvent` class
- Added `APP_COMPLETED_CHECKOUT` with value `"App Completed Checkout"`
## `ThanksViewModel`
- Tracking `APP_COMPLETED_CHECKOUT` if `CheckoutData` and `PledgeData` are present
  - Added tests for checking that `APP_COMPLETED_CHECKOUT` is tracked 
  - `ThanksViewModelTest.testTracking_whenCheckoutDataAndPledgeDataExtrasPresent` and `ThanksViewModelTest.testTracking_whenCheckoutDataAndPledgeDataExtrasNull` names were swapped 

# 👀 See
No visuals.

# 📋 QA
You can check your local Logcat filtered for `optly` to see Optimizely dispatching the event.

# Story 📖
[NT-1024]

[NT-1024]: https://kickstarter.atlassian.net/browse/NT-1024

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/ExperimentUtils.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/ExperimentUtils.kt
@@ -4,6 +4,7 @@ import com.google.firebase.iid.FirebaseInstanceId
 import com.kickstarter.libs.ApiEndpoint
 import com.kickstarter.libs.RefTag
 import com.kickstarter.models.User
+import com.kickstarter.ui.data.CheckoutData
 import java.util.*
 
 object ExperimentUtils {
@@ -23,3 +24,4 @@ object ExperimentUtils {
 }
 
 data class ExperimentData(val user: User?, val intentRefTag: RefTag?, val cookieRefTag: RefTag?)
+data class ExperimentRevenueData(val experimentData: ExperimentData, val checkoutData: CheckoutData)
--- a/app/src/main/java/com/kickstarter/libs/utils/ExperimentUtils.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/ExperimentUtils.kt
@@ -29,10 +29,12 @@ object ExperimentUtils {
         val project = experimentRevenueData.pledgeData.projectData().project()
         val fxRate = project.fxRate()
         val paymentType = experimentRevenueData.checkoutData.paymentType()
+        val revenue = (amount * fxRate * 100).roundToInt()
         return mutableMapOf(
                 Pair("checkout_amount", amount),
                 Pair("checkout_payment_type", paymentType.rawValue()),
-                Pair("checkout_revenue_in_usd_cents", (amount * fxRate * 100).roundToInt()),
+                Pair("checkout_revenue_in_usd_cents", revenue),
+                Pair("revenue", revenue),
                 Pair("currency", project.currency())
         )
     }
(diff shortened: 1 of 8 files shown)


app/src/test/java/com/kickstarter/libs/utils/ExperimentUtilsTest.kt
```diff
@@ -71,4 +73,32 @@ class ExperimentUtilsTest : KSRobolectricTestCase() {
...
+        val pledgeData = PledgeData.with(PledgeFlowContext.NEW_PLEDGE, ProjectDataFactory.project(ProjectFactory.caProject()), RewardFactory.reward())
+        val checkoutData = CheckoutDataFactory.checkoutData(10.0, 20.0)
+        val experimentRevenueData = ExperimentRevenueData(experimentData, checkoutData, pledgeData)
+        val checkoutTags = ExperimentUtils.checkoutTags(experimentRevenueData)
+        assertEquals(20.0, checkoutTags["checkout_amount"])
+        assertEquals("CREDIT_CARD", checkoutTags["checkout_payment_type"])
+        assertEquals(1500, checkoutTags["checkout_revenue_in_usd_cents"])
+        assertEquals("CAD", checkoutTags["currency"])
+        assertEquals(1500, checkoutTags["revenue"])
+    }
```
Setsun: Nice tests 👍

app/src/test/java/com/kickstarter/viewmodels/ThanksViewModelTest.java
```diff
@@ -351,16 +350,18 @@ public void testTracking_whenCheckoutDataAndPledgeDataExtrasNull() {
...
-  public void testTracking_whenCheckoutDataAndPledgeDataExtrasPresent() {
+  public void testTracking_whenCheckoutDataAndPledgeDataExtrasNull() {
     setUpEnvironment(environment());
 
     final Intent intent = new Intent()
             .putExtra(IntentKey.PROJECT, ProjectFactory.project());
     this.vm.intent(intent);
 
     this.lakeTest.assertNoValues();
+    this.experimentsTest.assertNoValues();
```
Setsun: nice and simple 👍

app/src/main/java/com/kickstarter/libs/utils/ExperimentUtils.kt
```diff
@@ -20,6 +23,22 @@ object ExperimentUtils {
...
         )
     }
+
+    fun checkoutTags(experimentRevenueData: ExperimentRevenueData): MutableMap<String, Any?> {
+        val amount = experimentRevenueData.checkoutData.amount()
+        val project = experimentRevenueData.pledgeData.projectData().project()
+        val fxRate = project.fxRate()
+        val paymentType = experimentRevenueData.checkoutData.paymentType()
+        val revenue = (amount * fxRate * 100).roundToInt()
+        return mutableMapOf(
```
Setsun: Would it be better to return a regular `Map` here, if we don't mutate this data?

eoji: Yeah, that makes sense! And I can do the same for `attributes`

PR #806, opened 2020-03-27
[🆕] Replacing FirebaseJobDispatcher with WorkManager

eoji:
**Note**: This is blocked by 2.4.0 

# 📲 What
Replacing `FirebaseJobDispatcher` with `WorkManager`

# 🤔 Why

# 🛠 How
- [🆕] NT-996 Replacing tracking BackgroundServices with Workers (#795) 
- [📲] NT-997 NT-998 Replacing push token Services with Workers (#796) 
- Removing library and `Dispatcher.kt` because they're no longer used.

# 👀 See
It's all in the background 😎 

# 📋 QA
After this is merged, tracking events and push token registration should not regress.

# Story 📖
[NT-898]

[NT-898]: https://kickstarter.atlassian.net/browse/NT-898


app/src/main/java/com/kickstarter/libs/TrackingClient.kt
```diff
@@ -43,28 +46,44 @@ abstract class TrackingClient(@param:ApplicationContext private val context: Con
...
+        if (type() == Type.KOALA) {
+            requestBuilder = OneTimeWorkRequestBuilder<KoalaWorker>()
+        } else if (type() == Type.LAKE) {
+            requestBuilder = OneTimeWorkRequestBuilder<LakeWorker>()
+        }
+
+        requestBuilder?.let {
+            val request = it
+                    .setInputData(data)
+                    .setBackoffCriteria(BackoffPolicy.EXPONENTIAL, 30, TimeUnit.SECONDS)
```
jgsamudio, line 74: Can we reuse the `BACKOFF_DELAY` from the DeviceRegistrar?

eoji: I'd rather give it its own, and I think I'll change it to `10` since tracking events will be happening more frequently.

eoji: But I think it makes the code less readable
```
            val backOffDelay = 10
            val request = it
                    .setInputData(data)
                    .setBackoffCriteria(BackoffPolicy.EXPONENTIAL, backOffDelay, TimeUnit.SECONDS)
                    .setConstraints(baseConstraints)
                    .build()
```

jgsamudio: If its only used in one place, then we don't need the constant. I didn't know if those values were the same. Changing it to 10 makes sense

PR #813, opened 2020-03-31
[⛔️] NT-1080 Pledge screen UI when fixing an errored backing's payment

eoji:
**Please note:** The `Toolbar` does not live in the `PledgeFragment` and is not part of this ticket.

# 📲 What
 Pledge screen UI when fixing an errored backing's payment

# 🤔 Why
So users **can only** change their payment method when fixing an errored backing.

# 🛠 How
## `PledgeFragmentViewModel`
- Adding support for `PledgeReason.FIX_PLEDGE` UI (it looks identical to updating your pledge).
  - The pledge summary is **visible** if the pledge had shipping.
  - The shipping summary is **visible** if the pledge had shipping.
  - The total is **visible**
  - The payment container is **visible**
  - The pledge section is **hidden**
  - The shipping section is **hidden**
  - The update pledge button is **hidden**
- Tests
  - The tests are almost identical to `PledgeReason.UPDATE_PAYMENT` but `"Update Payment Method Button Clicked"` is not tracked. There will be a new event.

# 👀 See

# 📋 QA
Smash that `Fix` button.
Selecting a payment method with call the `UpdateBacking` mutation.

# Story 📖
[NT-1080]

[NT-1080]: https://kickstarter.atlassian.net/browse/NT-1080

### DIFF
--- a/app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
+++ b/app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
@@ -529,6 +529,92 @@ class PledgeFragmentViewModelTest : KSRobolectricTestCase() {
     }
 
     @Test
+    fun testPledgeScreenConfiguration_whenFixingPaymentOfShippableReward() {
+        val shippableReward = RewardFactory.rewardWithShipping()
+        val unitedStates = LocationFactory.unitedStates()
+        val shippingRule = ShippingRuleFactory.usShippingRule().toBuilder().location(unitedStates).build()
+        val backing = BackingFactory.backing()
+                .toBuilder()
+                .amount(50.0)
+                .location(unitedStates)
+                .locationId(unitedStates.id())
+                .reward(shippableReward)
+                .rewardId(shippableReward.id())
+                .shippingAmount(shippingRule.cost().toFloat())
+                .build()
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/kickstarter/viewmodels/PledgeFragmentViewModelTest.kt
```diff
@@ -528,6 +528,92 @@ class PledgeFragmentViewModelTest : KSRobolectricTestCase() {
...
+        this.pledgeSummaryIsGone.assertValue(false)
+        this.shippingRulesSectionIsGone.assertValue(true)
+        this.shippingSummaryIsGone.assertValue(false)
+        this.totalDividerIsGone.assertValue(true)
+        this.updatePledgeButtonIsEnabled.assertNoValues()
+        this.updatePledgeButtonIsGone.assertValue(true)
+        this.updatePledgeProgressIsGone.assertNoValues()
+
+        this.koalaTest.assertNoValues()
+        this.lakeTest.assertNoValues()
```
yatriks, line 579: Just for my own curiousity, is this asserting that koala/lake events are _not_ fired here?

eoji: yep! And they don't fire because we don't have events for these yet

PR #822, opened 2020-04-04
[🔨] NT-1121 Stored cards redesign

eoji:
**Note:** This is an unfortunately big PR but there were previously 4 states for stored cards and now there are 2, so hopefully it's for the best!

# 📲 What
Stored cards redesign.

# 🤔 Why
@colleenmacd flexin

# 🛠 How
## Deletions and renames
- `FreezeLinearLayoutManager` is no longer needed to stop the list of cards from scrolling when one is selected
- `RewardLoadingCardViewHolder` and its layout `item_reward_loading_card.xml` is no longer needed because the progress state while a pledge is being created/updated is moving to the footer pledge button
- `item_reward_pledge_card_disabled.xml` is no longer needed because the pledge button lives in the footer
- `item_reward_credit_card.xml` is now `item_reward_unselected_card.xml`
- `item_reward_pledge_card.xml` is now `item_reward_selected_card.xml`
- `item_reward_add_card.xml` is now `item_add_card.xml`
- `RewardCardViewHolder` is now `RewardCardUnselectedViewHolder`
- `RewardCardViewHolderViewModel` is now `RewardCardUnselectedViewHolderViewModel`
- `RewardPledgeCardViewHolder` is now `RewardCardSelectedViewHolder`
- `RewardPledgeCardViewHolderViewModel` is now `RewardCardSelectedViewHolderViewModel`
## Style updates and additions
- `stored_card_radius` is now `grid_1`
- Added `unselected_card_stroke_width` with value `2dp`
- Added `circle_accent` to highlight which card is selected
- Added `circle_grey_500_stroke` to highlight which card is unselected
- `SelectButton` style is no longer used so it's been deleted
- `PledgeStoredCard`
  - no longer uses `cardUseCompatPadding` since the cards have no elevation
- `RewardCardDecoration` only sets `bottom` margin, the parent sets `start` and `end` margin
## Stored cards
- The stored cards got a makeover!
- `CardState`
  - There are now only 2 card states `SELECTED` AND `UNSELECTED`
- `RewardCardSelectedViewHolder`
  - Removed delegate since there are no actions you can take on this view
- `RewardCardUnselectedViewHolderViewModel`
  - Added outputs `isClickable`, `issuerImageAlpha`, `lastFourTextColor`, and `selectImageIsVisible`
  - Added tests for new outputs
- `PledgeFragmentViewModel`
  - Renamed `showPledgeCard` to `showSelectedCard`

# 👀 See
| After 🦋 | Before 🐛 | 
| --- | --- |
| screenshot-2020-04-03_200420 | screenshot-2020-04-03_200409 |
| screenshot-2020-04-03_200641 | screenshot-2020-04-03_200649 |
| screenshot-2020-04-03_200313 | screenshot-2020-04-03_200351 |

# 📋 QA
**Please note, for the sake of keeping this PR smaller:** 
- The new pledge button in the footer is still not wired up. That will be in a separate PR.
- I will autoselect the first payment method in a separate PR.
- I will do copy changes separately 

# Story 📖
[NT-1121]

[NT-1121]: https://kickstarter.atlassian.net/browse/NT-1121


app/src/main/java/com/kickstarter/viewmodels/RewardCardUnselectedViewHolderViewModel.kt
```diff
@@ -1,76 +1,66 @@
...
-            val isBackingPaymentAndAllowedType = isBackingPaymentSource
-                    .compose<Pair<Boolean, Boolean>>(combineLatestPair(allowedCardType))
+            allowedCardType
+                    .compose(bindToLifecycle())
+                    .subscribe(this.isClickable)
 
-            isBackingPaymentAndAllowedType
-                    .map { !it.first && it.second }
+            allowedCardType
+                    .map { if (it) 1.0f else .5f }
```
jgsamudio, line 56: Is this to fade the card image if there the card is disabled? I don't know if there are constants elsewhere in the app for this or if a similar behavior is found elsewhere?

app/src/main/res/drawable/circle_accent.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+  android:shape="oval">
+  <solid android:color="@color/accent" />
+  <padding
+    android:bottom="@dimen/grid_1_half"
```
jgsamudio, line 6: could you use `android:padding="@dimen/grid_1_half" `

app/src/main/res/layout/item_reward_unselected_card.xml
```diff
@@ -0,0 +1,48 @@
...
+      android:id="@+id/card_not_allowed_warning"
+      style="@style/Caption1SecondaryMedium"
+      android:layout_width="wrap_content"
+      android:layout_height="wrap_content"
+      android:layout_marginStart="@dimen/grid_8"
+      android:layout_marginEnd="@dimen/grid_2"
+      android:layout_marginBottom="@dimen/grid_2"
+      android:textColor="@color/ksr_red_400"
+      android:visibility="gone"
+      tools:text="@string/You_cant_use_this_credit_card_to_back_a_project_from_project_country"
```
jgsamudio, line 45: This is a really long string. Any way we can shorten it?

app/src/main/java/com/kickstarter/viewmodels/RewardCardUnselectedViewHolderViewModel.kt
eoji: Yeah! No constants, this is our first time disabling an image in this way (previously we disabled the button).

In the activity feed, we make project images `.2` alpha but that's also just a hardcoded number.

app/src/main/res/drawable/circle_accent.xml
eoji: yeah! will update

app/src/main/res/layout/item_reward_unselected_card.xml
eoji: No, unfortunately we make the entire string the key because when the string changes we also need to update the key

app/src/main/res/drawable/circle_accent.xml
eoji: Actually, you cannot! You have to specify each side https://developer.android.com/guide/topics/resources/drawable-resource#Shape

app/src/main/java/com/kickstarter/viewmodels/RewardCardUnselectedViewHolderViewModel.kt
jgsamudio: Ok lets keep it the same as the Figma designs

PR #827, opened 2020-04-08
[💳] NT-1123 Card not allowed copy

eoji:
# 📲 What
Updating the copy for when a card is not allowed.

# 🤔 Why
New copy party.

# 🛠 How
- Added `This_project_has_a_set_currency_that_cant_process_this_option` string and set it as the text for `card_not_allowed_warning`
## `RewardCardUnselectedViewHolderViewModel`
- Removed `projectCountry` output because it's no longer needed
## `PledgeFragmentViewModel` 🐛 
- Fixed bug where a card that is not allowed could be selected by default
- Added test for when the first card is not allowed
- Added test for when none of the cards are allowed

# 👀 See
| After 🦋 | Before 🐛|
| --- | --- |
| screenshot-2020-04-08_154955 | screenshot-2020-04-03_200313 |

# 📋 QA
Try to pledge to a project with a card that's not allowed. SIKE, you can't.

# Story 📖
[NT-1123]

[NT-1123]: https://kickstarter.atlassian.net/browse/NT-1123

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
@@ -1002,11 +1002,11 @@ interface PledgeFragmentViewModel {
         }
 
         private fun initialCardSelection(storedCards: List<StoredCard>, project: Project): Pair<StoredCard, Int>? {
-            val defaultIndex = 0
+            val defaultIndex = storedCards.indexOfFirst { ProjectUtils.acceptedCardType(it.type(), project) }
             val backingPaymentSourceIndex = storedCards.indexOfFirst { it.id() == project.backing()?.paymentSource()?.id() }
             return when {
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -1002,11 +1002,11 @@ interface PledgeFragmentViewModel {
         }
 
         private fun initialCardSelection(storedCards: List<StoredCard>, project: Project): Pair<StoredCard, Int>? {
-            val defaultIndex = 0
+            val defaultIndex = storedCards.indexOfFirst { ProjectUtils.acceptedCardType(it.type(), project) }
```
ifbarrera, line 1005: Lol we had this same bug.

eoji: such a small but TRICKY thing

PR #855, opened 2020-04-24
[🔄] NT-1192 Moving Creator Details Button Clicked event from Optimizely to Data 💧

eoji:
# 📲 What
Moving `Creator Details Button Clicked` from Optimizely to the Data Lake

# 🤔 Why
We're moving Optimizely tracking events into the Data Lake.

# 🛠 How
## `Koala`
- Added `trackCreatorDetailsClicked` that tracks `Creator Details Button Clicked` with Optimizely properties
- Added `experimentProperties` helper method since the 3 experiment tracking methods did the same thing
## `LakeEvent`
- Added `CREATOR_DETAILS_CLICKED` const with value `Campaign Details Button Clicked` in `native_project_page_conversion_creator_details` region
## `OptimizelyEvent`
- Removed `CREATOR_DETAILS_CLICKED` and `native_project_page_conversion_creator_details` region
## `ProjectViewModel`
- Tracking `creatorInfoClicked` event with `lake` instead of `optimizely`
- Updated tests

# 👀 See
No visual changes.

# 📋 QA
- Click the creator details
- Check your local logcat

# Story 📖
[NT-1192]

[NT-1192]: https://kickstarter.atlassian.net/browse/NT-1192

### DIFF
--- a/app/src/test/java/com/kickstarter/viewmodels/ProjectViewModelTest.kt
+++ b/app/src/test/java/com/kickstarter/viewmodels/ProjectViewModelTest.kt
@@ -479,7 +479,7 @@ class ProjectViewModelTest : KSRobolectricTestCase() {
 
         this.vm.inputs.creatorNameTextViewClicked()
         this.startCreatorBioWebViewActivity.assertValues(project)
-        this.experimentsTest.assertValue("Creator Details Clicked")
+        this.lakeTest.assertValues("Project Page Viewed", "Creator Details Clicked")
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


app/src/test/java/com/kickstarter/viewmodels/ProjectViewModelTest.kt
```diff
@@ -492,7 +492,7 @@ class ProjectViewModelTest : KSRobolectricTestCase() {
 
         this.vm.inputs.creatorNameTextViewClicked()
         this.startCreatorBioWebViewActivity.assertValues(project)
-        this.experimentsTest.assertNoValues()
+        this.lakeTest.assertValue("Project Page Viewed")
```
Arkariang, line 495: Could we have a constant for the "Project Page Viewd" ?

eoji: I don't think so. We don't test constants because if you spell it wrong, your test would not catch that.

PR #863, opened 2020-04-28
NT-658 ErroredPledge push notification

eoji:
# 📲 What
Adds support for push notifications containing `ErroredPledge`

# 🤔 Why
We will be turning on pushes for errored pledges.

# 🛠 How
## `PushNotificationEnvelope` 
- Added `ErroredPledge` model
- Added `erroredPledge` field
## `PushNotifications`
- Added new channel notification `CHANNEL_ERRORED_PLEDGES` with high priority
- Added a subscription for when the `PushNotificationEnvelope` contains an `ErroredPledge` object
- Created a helper method `projectIntent` to that returns an `Intent` for a `Project` so we can add more extras
## `ProjectViewModel`
- Fixed bug where the pledge sheet `ProgressBar` was still visible if the `ProjectActivity` was started with an `EXPAND_PLEDGE_SHEET` extra
## `DebugPushNotificationsView`
- Added a button for triggering an `ErroredPledge` push that will take you to the Project page with the pledge sheet expanded
# 👀 See
|  |  |
| --- | --- |
|  | device-2020-04-28-131801 2020-04-28 13_19_08 |

# 📋 QA
TBD

# Story 📖
[NT-658]

[NT-658]: https://kickstarter.atlassian.net/browse/NT-658

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -18,6 +18,7 @@
   <string name="Device_ID">Device ID</string>
   <string name="debug_push_notifications_register_device">Register device</string>
   <string name="debug_push_notifications_device_registration">Device registration</string>
+  <string name="debug_push_notifications_errored_pldege">Errored pledge</string>
   <string name="debug_push_notifications_friend_backing">Friend backing</string>
   <string name="debug_push_notifications_friend_follow">Friend follow</string>
   <string name="debug_push_notifications_message">Message</string>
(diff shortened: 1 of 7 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -18,6 +18,7 @@
   <string name="Device_ID">Device ID</string>
   <string name="debug_push_notifications_register_device">Register device</string>
   <string name="debug_push_notifications_device_registration">Device registration</string>
+  <string name="debug_push_notifications_errored_pldege">Errored pledge</string>
```
Arkariang, line 21: Just for information, this string is not gonna be translated right?

eoji: Correct, because it's only used in the Internal Tools!

PR #864, opened 2020-04-28
NT-1148: Fix pledge button clicked event

Arkariang:
# 📲 What
-  Added event for Fix pledge button inside View your pledge
Screen Shot 2020-04-28 at 11 57 07 AM

# 🤔 Why

New tracking events

# 📋 QA

When hitting that red button take a look into your logcat
Screen Shot 2020-04-28 at 12 03 05 PM

# Story 📖

[Fix pledge button clicked event](https://kickstarter.atlassian.net/browse/NT-1148)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/Koala.java
+++ b/app/src/main/java/com/kickstarter/libs/Koala.java
@@ -724,6 +724,12 @@ public final class Koala {
 
     this.client.track(LakeEvent.THANKS_PAGE_VIEWED, props);
   }
+
+  public void trackFixPledgeButtonClicked(final @NonNull ProjectData projectData) {
+    final Map<String, Object> props = KoalaUtils.projectProperties(projectData.project(), this.client.loggedInUser());
+
+    this.client.track(LakeEvent.FIX_PLEDGE_BUTTON_CLICKED, props);
+  }
   //endregion
 
   //region Log In or Signup
--- a/app/src/main/java/com/kickstarter/libs/Koala.java
+++ b/app/src/main/java/com/kickstarter/libs/Koala.java
@@ -727,6 +727,7 @@ public final class Koala {
 
   public void trackFixPledgeButtonClicked(final @NonNull ProjectData projectData) {
     final Map<String, Object> props = KoalaUtils.projectProperties(projectData.project(), this.client.loggedInUser());
+    props.put("context_pledge_flow", PledgeFlowContext.FIX_ERRORED_PLEDGE.getTrackingString());
 
     this.client.track(LakeEvent.FIX_PLEDGE_BUTTON_CLICKED, props);
   }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/libs/Koala.java
```diff
@@ -724,6 +724,12 @@ public void trackThanksPageViewed(final @NonNull CheckoutData checkoutData, fina
 
     this.client.track(LakeEvent.THANKS_PAGE_VIEWED, props);
   }
+
+  public void trackFixPledgeButtonClicked(final @NonNull ProjectData projectData) {
```
eoji, line 728: We need to add the `context_pledge_flow` property:
```
props.put("context_pledge_flow", PledgeFlowContext.FIX_ERRORED_PLEDGE.getTrackingString());
```

app/src/main/java/com/kickstarter/libs/LakeEvent.kt
```diff
@@ -20,6 +20,7 @@ const val PLEDGE_SUBMIT_BUTTON_CLICKED = "Pledge Submit Button Clicked"
 const val PROJECT_PAGE_PLEDGE_BUTTON_CLICKED = "Project Page Pledge Button Clicked"
 const val SELECT_REWARD_BUTTON_CLICKED = "Select Reward Button Clicked"
 const val THANKS_PAGE_VIEWED = "Thanks Page Viewed"
+const val FIX_PLEDGE_BUTTON_CLICKED = "Fix Pledge Button Clicked"
```
eoji, line 23: This event actually is in the `Manage a Pledge` set of events so we need a new region 🆕 
```
// region Manage a Pledge
const val FIX_PLEDGE_BUTTON_CLICKED = "Fix Pledge Button Clicked"
//endregion
```

app/src/main/java/com/kickstarter/viewmodels/ProjectViewModel.kt
```diff
@@ -777,6 +777,12 @@ interface ProjectViewModel {
                     .filter { it.project().isLive && !it.project().isBacking }
                     .compose(bindToLifecycle())
                     .subscribe { this.lake.trackCreatorDetailsClicked(it) }
+
+            projectData
+                    .compose<ProjectData>(takeWhen(this.fixPaymentMethodButtonClicked))
+                    .compose(bindToLifecycle())
+                    .subscribe{ this.lake.trackFixPledgeButtonClicked(it) }
```
eoji, line 784: So now that we have a tracking event, we need to add a test! `testShowUpdatePledge_whenFixingPaymentMethod`

Arkariang: I complety forgot about the test! sorry :anguished:

eoji: Np, you got this 🚀

PR #874, opened 2020-05-06
NT-1244 Listening for Optimizely initialization in BaseFragment

eoji:
# 📲 What
Listening for Optimizely initialization in `BaseFragment`

# 🤔 Why
`optimizely.isFeatureEnabled` can be called before `Optimizely` has been initialized and we might miss if the user is actually supposed to be in a feature.

# 🛠 How
## `ApplicationModule`
- When Optimizely is `valid`, we send a broadcast `Intent` with the action `"experiments_client_ready"`
## `BaseFragment`
- Added `optimizelyReadyReceiver` that let's the VM know Optimizely is Ready™ 
- Registering and unregistering `optimizelyReadyReceiver` in `onResume`
- Unregistering `optimizelyReadyReceiver` in `onPause`
## `FragmentViewModel`
- Added `PublishSubject` `optimizelyReady` that's called when we receive the broadcast `"experiments_client_ready"`. 
## `DiscoveryFragmentViewModel`
- When `optimizelyReady` emits we check if `LIGHTS_ON` is enabled.
- Added a test!

## bug
Fixed bug where `EditiorialActivity` was started twice

# 👀 See
N/A

# 📋 QA
Logged out and logged in users should see Lights On if they're in the feature.

# Story 📖
[NT-1244]

[NT-1244]: https://kickstarter.atlassian.net/browse/NT-1244


app/src/main/java/com/kickstarter/ui/viewholders/EditorialViewHolder.kt
```diff
@@ -38,12 +37,7 @@ class EditorialViewHolder(val view: View, val delegate: Delegate) : KSViewHolder
         this.vm.outputs.editorial()
                 .compose(bindToLifecycle())
                 .compose(Transformers.observeForUI())
-                .subscribe { delegate.editorialViewHolderClicked(it) }
```
Arkariang, line 41: 💖

app/src/test/java/com/kickstarter/viewmodels/DiscoveryFragmentViewModelTest.java
```diff
@@ -193,7 +193,7 @@ public void testShouldShowEditorial_defaultParams() {
 
 
   @Test
-  public void testShouldShowLightsOne_featureEnabled(){
```
Arkariang, line 196: 👍

PR #892, opened 2020-05-21
NT-1259: Firebase Analytics dependency

Arkariang:
# 📲 What

- Added Firebase Analytics Kotlin dependency.
- Updated Firebase Messaging (it's also a dependency for Firebase analytics, so I'm force to use the same version as them)
- Updated some gems 
-- Rubyzip & excon to fix the vulnerability issue in the repo
-- The rest of them needed for fastlane plugin

# 🤔 Why
To be able to track more data from Android app.
In that capture as you can see no crash free user information and hitting on the lastest release button just gave us information from the las one available one 1.18.1
Screen Shot 2020-05-19 at 11 04 02 AM

# 👀 See

Now we have our data, in internal release for now 😆 !!
Screen Shot 2020-05-20 at 5 17 31 PM
Screen Shot 2020-05-20 at 2 29 39 PM
* Including the app version adoption, which will help us decide the best moment for updating rollouts 
Screen Shot 2020-05-20 at 5 27 15 PM

# 📋 QA
- Execute in your environment the next scripts and nothing should fail:
```bundle install```
```bundle exec fastlane external|internal ``` --> depending on the type you wanna build
```./milkrun update-strings```
- Try to message someone using the app

# Story 📖

[Analytics dependency](https://kickstarter.atlassian.net/browse/NT-1259)

### CODE REVIEW

File: app/src/main/assets/json/server-config.json [hunk 1]
```diff
@@ -620,7 +620,7 @@
       "Subscribes_to_upcoming_lives_streams": "Subscribes to upcoming live streams.",
       "Successful_projects_have_been_created_on_kickstarter": "Successful projects have been created on Kickstarter.",
       "Support_creative_independence": "Support creative independence",
-      "Support_creative_spaces_and_businesses_affected_by": "Support creative spaces and businesses affected by COVID-19. See projects near you. ▸",
```
jgsamudio, line 623: Did we also want to merge these changes in this PR?

Arkariang: yeap, it's just updating the translations for the updated copy for lights on, and the result of executing  ```./milkrun update-strings``` . With this and the releases firebase internal you can see that all the scripts work perfectly after updating the gems :+1:

justinswart: For what it's worth - what we typically do on iOS is only commit string updates that are relevant to the changes being made in the PR. And then from time-to-time we will run an app-wide translations update and open a PR for that specifically.

app/src/test/java/com/kickstarter/libs/utils/ExperimentUtilsTest.kt
```diff
@@ -15,7 +15,7 @@ class ExperimentUtilsTest : KSRobolectricTestCase() {
                 .backedProjectsCount(10)
                 .build()
         val experimentData = ExperimentData(user, RefTag.discovery(), RefTag.search())
-        val attributes = ExperimentUtils.attributes(experimentData, "9.9.9", "10", OptimizelyEnvironment.STAGING)
+        val attributes = ExperimentUtils.attributes(experimentData, "9.9.9", "10", OptimizelyEnvironment.DEVELOPMENT)
```
justinswart, line 18: Why this change?

Arkariang: The test is checking attributes for a user not logged in production, I changed it to Development instead of staging because for staging using internally a FirebaseID, which requires have Firebase active in a unit tes and an API key, no unit test should need that.
This rise a question, why for staging we need to add the FirebaseID into this experiment params, (for production is null)? which I cannot answer tbh  😞.
You can see it here: ExperimentUtils.kt:13

PR #895, opened 2020-06-03
NT-1231 Creator Perspective query + View your pledge updated

Arkariang:
# 📲 What
- New query added for retrieving the backings with a backingId.
- Refactor BackingActivity + BackingActivityViewModel to just load BackingFragment with the updated information.
:warning: Sorry for the big PR, feel free to ping me in case of any doubt 

# 🤔 Why
- We will need this for retrieving Add-ons from a creator perspective
- Unify UI removing the old UI from messages to see your pledge.

# 🛠 How

- New query on checkout.graphql
- Minor adjustments to send the BackingId from MessagesActivity, in order to have the backing information when pressing "view pledge" button
- BackingActivity + BackingViewModel have been refactored to hold the BackingFragment and launch with the new query information.
- Added new header, the header you see on BackingFragment when loading from the project page, is in fact the project header. No access to that header from this context.

# 👀 See

| Before 🐛 | After 🦋 |
| 
Screen Shot 2020-06-03 at 8 42 05 AM

|
|  |  |

# 📋 QA

- For testing the creator facing view go to PRO, log in as therealnativesquad@gmail.com, go the dashboard, select the project bugs in the office, and look fo a message from luisa, open the message, and hit the view pledge button. In this case your a seeing a Pledge from a creator perspective :) 

# Story 📖

[NT-1231 Creator Perspective query + View your pledge view](https://kickstarter.atlassian.net/browse/NT-1231)


app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
```diff
@@ -171,13 +171,13 @@ interface BackingFragmentViewModel {
 
             val backedProject = this.projectDataInput
                     .map { it.project() }
-                    .filter { it.isBacking }
 
-            val backing = backedProject
-                    .switchMap { it.slug()?.let { slug -> this.apolloClient.getProjectBacking(slug) } }
+            val backing = this.projectDataInput
+                    .switchMap { getBackingInfo(it) }
+                    .distinctUntilChanged()
```
justinswart, line 177: Why do we need to check for uniqueness here?

justinswart:
Cool! I was able to navigate to the new backing view:

image

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
Arkariang: You can  hit the view pledge button, then the back arrow and hit again view pledge really quick. that's why I added the check for uniqueness but honestly to not repeat the network call uneccessarly. But if you have a better way for avoiding that second call let me know!

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
```diff
@@ -520,6 +548,56 @@ class KSApolloClient(val service: ApolloClient) : ApolloClientType {
...
+            .backerName(name)
+            .reward(reward)
+            .id(backerId)
+            .backer(backer)
+            .locationId(locationId)
+            .locationName(backingGr?.location()?.displayableName())
+            .pledgedAt(backingGr?.pledgedOn())
+            .projectId(projectId)
+            .sequence(backingGr?.sequence()?.toLong() ?: 0)
+            .shippingAmount(backingGr?.shippingAmount()?.amount().toString().toFloat())
```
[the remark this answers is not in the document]
Arkariang: I cast them to the exact same type the it's needed for the builder, the unconsistency on the types it's something we can explore as techDebt at some point I think, not sure about the impact to refactor them to use the same type at this point.
Screen Shot 2020-06-04 at 10 14 50 AM
Screen Shot 2020-06-04 at 10 15 04 AM

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
justinswart: Ok, perhaps I'm misunderstanding the code, what I'm reading is:

1. `.distinctUntilChanged()` is placed after the call would have happened so I don't think it would prevent repeating the network call in this case?
2. `getBackingInfo(it:)` appears to return the backing immediately if it's passed in so if we already have one it won't make another request - would we already have one in the scenario you're describing?

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
```diff
@@ -520,6 +548,56 @@ class KSApolloClient(val service: ApolloClient) : ApolloClientType {
...
+    val rewardSingleLocation = Reward.SingleLocation.builder()
+            .localizedName(backingGr?.location()?.displayableName())
+            .id(decodeRelayId(backingGr?.location()?.id())?: 0)
+            .build()
+
+    val reward = Reward.builder()
+            .minimum(rewardAmount?: 0.0)
+            .description(backingGr?.reward()?.description())
+            .estimatedDeliveryOn(DateTime(backingGr?.reward()?.estimatedDeliveryOn()))
+            .convertedMinimum(rewardAmount ?: 0.0)
```
justinswart, line 572: The `convertedMinimum` is based on the projects currency and exchange rate, we might need this to come through from GraphQL too.

Arkariang: got it!

PR #896, opened 2020-06-05
NT-1294: RecyclerView structure for reward & Add-ons

Arkariang:
# 📲 What
The UI in the backing fragment has change to hold a RecyclerView, now we can have a list of rewards & add-os.

# 🤔 Why

We will need to display the list of add-ons a backer has selected.

# 👀 See

 Before 🐛 

Screen Shot 2020-06-03 at 3 48 12 PM

##  After 🦋 | I forced and empty list of add-ons 
If you want to see the list as in the gif below, you can call populateDataForAddOns in BackingFragment.kt like in the screenshoot
Screen Shot 2020-06-05 at 7 56 48 AM

rwards addons
 

# 📋 QA

What you should see now when in the view/manage my pledge screen is the reward card loading exactly the same as it does in the rewards carrousel,  larger than before and sowing the select button de-activated. The logic for adapting that UI will come later on.

# Story 📖

[Adapt Reward UI in View/Manage pledge to a List](https://kickstarter.atlassian.net/browse/NT-1294)


app/src/main/java/com/kickstarter/ui/adapters/RewardAndAddOnsAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+        setSection(SECTION_ADD_ONS_CARD, rewards)
+        notifyDataSetChanged()
+    }
+
+    fun populateDataForRewad(reward: Pair<ProjectData, Reward>) {
+        setSection(SECTION_REWARD_CARD, listOf(reward))
+        notifyDataSetChanged()
+    }
+
+    companion object {
```
jgsamudio, line 43: Random question, is there a standard for where the companion object should be placed in the file? For example it should always be placed toward the top or bottom of the file?

Arkariang: I'be mostly see it on the bottom of the classes -> [Kotlin code conventions](https://kotlinlang.org/docs/reference/coding-conventions.html)
Screen Shot 2020-06-05 at 1 32 32 PM

app/src/main/java/com/kickstarter/ui/adapters/RewardAndAddOnsAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.util.Pair
+import android.view.View
+import com.kickstarter.R
+import com.kickstarter.models.Reward
+import com.kickstarter.ui.data.ProjectData
+import com.kickstarter.ui.viewholders.*
+
+class RewardAndAddOnsAdapter(private val viewListener: RewardAndAddOnsAdapter.ViewListener) : KSAdapter() {
+
+    interface ViewListener:RewardViewHolder.Delegate, AddOnViewHolder.ViewListener
```
JadeByfield89, line 12: It looks like both of these interfaces do the same thing (listen for user input), I found Delegate to be confusing at first. ViewListener makes sense. Is that why you changed it?

app/src/main/java/com/kickstarter/ui/adapters/RewardAndAddOnsAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    override fun layout(sectionRow: SectionRow): Int = when (sectionRow.section()){
+        SECTION_REWARD_CARD -> R.layout.item_reward
+        SECTION_ADD_ONS_CARD -> R.layout.item_add_on
+        else -> 0
+    }
+
+    override fun viewHolder(layout: Int, view: View): KSViewHolder {
+        return when(layout) {
```
JadeByfield89, line 26: Nice!

app/src/main/res/layout/item_add_on.xml
```diff
@@ -0,0 +1,22 @@
...
+  xmlns:app="http://schemas.android.com/apk/res-auto"
+  xmlns:tools="http://schemas.android.com/tools"
+  android:layout_width="@dimen/item_reward_width"
+  android:layout_height="wrap_content"
+  android:orientation="vertical">
+
+    <androidx.cardview.widget.CardView
+      android:id="@+id/add_on_reward_card"
+      android:layout_width="match_parent"
+      android:layout_height="300dp"
```
JadeByfield89, line 14: I'm assuming this doesn't need to be a dimens field because it's just a placeholder for now?

app/src/main/java/com/kickstarter/ui/adapters/RewardAndAddOnsAdapter.kt
Arkariang: For now it's empty and doing the same thing because I've implemented the same "action" as the rewards, that's quite sure gonna change. I just what to handle both view listener one for rewards and the other for AddOns, I didn't really change the name because AddOns are in essence  rewards ... anyway, not sure yet how to call that action, any suggestion is welcome :) 
And the Delegate part comes from RewardViewHolder, which is already implemented, the naming is because that code was started by iOS developers and that's how the call the viewListeners 😏

app/src/main/res/layout/item_add_on.xml
Arkariang: correct! I just want it to have some space to be able to see int the UI, for showing it in the gif so just QA purposes

PR #897, opened 2020-06-08
NT-1141: Refactor existing queries

Arkariang:
# 📲 What

Refactor in queries Project.backing and Backing.backing to use fragments for each subtype 

# 🤔 Why

Reusability - Readability - No code smells 

# 🛠 How
Creating separate fragments for each type and not attaching them to any high level query.

# 👀 See
No UI chenges here 

# 📋 QA

- Check one of your backed projects, hit the manage pledge button, all data should remain the same as before
- Go to any message os your messages with a creator of a project you already backed, go to the thread view, hit the view pledge button, all data should remain as before.
- Log in as creator if you can, go to any message you have with a baker to one of your projects, hit the view pledge, you should see the pledge data as before.

### DIFF
--- a/app/src/main/java/com/kickstarter/services/KSApolloClient.kt
+++ b/app/src/main/java/com/kickstarter/services/KSApolloClient.kt
@@ -570,11 +570,11 @@ private fun createBackingObject(backingGr: fragment.Backing?): Backing {
     val shippingAmount = backingGr?.shippingAmount()?.fragments()
 
     val reward = backingGr?.reward()?.fragments()?.reward()?.let { reward ->
-        val rewardId = decodeRelayId(reward.id()) ?: 0
+        val rewardId = decodeRelayId(reward.id()) ?: -1
         val rewardAmount = reward.amount().fragments().amount().amount()?.toDouble()
         val rewardSingleLocation = Reward.SingleLocation.builder()
                 .localizedName(location?.displayableName())
-                .id(decodeRelayId(location?.id())?: 0)
+                .id(decodeRelayId(location?.id())?: -1)
                 .build()
 
         return@let Reward.builder()
@@ -588,7 +588,7 @@ private fun createBackingObject(backingGr: fragment.Backing?): Backing {
 
     val backerData = backingGr?.backer()?.fragments()?.user()
     val nameBacker = backerData?.name()
-    val backerId= decodeRelayId(backerData?.id()) ?: 0
+    val backerId= decodeRelayId(backerData?.id()) ?: -1
     val avatar = Avatar.builder()
             .medium(backerData?.imageUrl())
             .build()
--- a/app/src/main/java/com/kickstarter/services/KSApolloClient.kt
+++ b/app/src/main/java/com/kickstarter/services/KSApolloClient.kt
@@ -122,6 +122,7 @@ class KSApolloClient(val service: ApolloClient) : ApolloClientType {
                         override fun onResponse(response: Response<GetBackingQuery.Data>) {
                             response.data()?.let {data ->
                                 Observable.just(data.backing())
+                                        .filter { it?.fragments()?.backing() != null }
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/services/KSApolloClient.kt
```diff
@@ -548,88 +549,74 @@ class KSApolloClient(val service: ApolloClient) : ApolloClientType {
...
-            .estimatedDeliveryOn(DateTime(backingGr?.reward()?.estimatedDeliveryOn()))
-            .shippingSingleLocation(rewardSingleLocation)
-            .id(rewardId)
+    val location = backingGr?.location()?.fragments()?.location()
+    val locationId = decodeRelayId(location?.id())
+    val projectId = decodeRelayId(backingGr?.project()?.fragments()?.project()?.id()) ?: 0
+    val shippingAmount = backingGr?.shippingAmount()?.fragments()
+
+    val reward = backingGr?.reward()?.fragments()?.reward()?.let { reward ->
+        val rewardId = decodeRelayId(reward.id()) ?: 0
```
justinswart, line 573: I would coalesce this to `-1` just in case. There is some historical behaviour where a `Reward` with and ID of `0` is treated as "No Reward" 😄

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
```diff
@@ -548,88 +549,74 @@ class KSApolloClient(val service: ApolloClient) : ApolloClientType {
...
+            .reward(reward)
+            .rewardId(reward?.id())
             .locationId(locationId)
-            .locationName(backingGr?.location()?.displayableName())
+            .locationName(location?.displayableName())
             .pledgedAt(backingGr?.pledgedOn())
             .projectId(projectId)
             .sequence(backingGr?.sequence()?.toLong() ?: 0)
-            .shippingAmount(backingGr?.shippingAmount()?.amount().toString().toFloat())
+            .shippingAmount(shippingAmount?.amount()?.amount()?.toFloat() ?: 0f)
```
jgsamudio, line 617: Should this still be optional or should be force to 0?

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
```diff
@@ -189,6 +189,7 @@ interface BackingFragmentViewModel {
                     .subscribe(this.backerAvatar)
 
             this.projectDataInput
+                    .filter { it.project().isBacking || ProjectUtils.userIsCreator(it.project(), it.user()) }
```
jgsamudio, line 192: Should we add a test for this or is it already covered by another test?

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
Arkariang: got it!

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
Arkariang: It's really never gonna 0 but retrieving it from Graph force it to be optional because of the unwrapping. I'll change the defaults as Justin suggested to -1, that would be clear that if we encounter that value something went wrong

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
Arkariang: Great point, I'll add a test

Arkariang: Done! two new tests, 
- one checking as creator being able to see the reward of the backer.
- one for checking no rewad information if the user is neither a creator or a backer

PR #898, opened 2020-06-08
Disable pledge button if shipping selection is invalid

JadeByfield89:
# 📲 What

On the pledge screen, the pledge button was still active even if the user cleared their shipping selection, making it invalid.

# 🤔 Why

Seems to be a possible cause for errors, if we're allowing users to make a pledge(spend money), without first confirming that a reward can even be shipped to their location.

# 🛠 How

I added an OnTextChangedListener to the shipping address AutoCompleteTextView. Each time the text is changed, I check if it's null, empty, or blank (only blank spaces). If so, I disable the pledge button. When the user selects a valid location from the shipping dropdown, I re-enable the pledge button.

I'm not 100% sure if this correct as I couldn't see where we were getting the shipping locations from. So, in the case that the user can select a location from the dropdown, and that location is "invalid"...this fix won't work. My assumption here is that the shipping location dropdown only shows valid locations that a reward could be sent to.

# 👀 See
| Before 🐛 
Screen Shot 2020-06-08 at 7 48 10 PM

| After 🦋 |
Screen Shot 2020-06-08 at 7 37 30 PM

|  |  |

# 📋 QA

1. On the android app, open a project with rewards that have shipping 

2. Select a reward with shipping 

3. On the pledge screen, select a payment method if one is not auto-selected

4. Tap on the shipping location field and delete the entry 

5. Tap outside the field and observe the Pledge button

6. Observe that the pledge button is disabled

# Story 📖

https://kickstarter.atlassian.net/secure/RapidBoard.jspa?rapidView=9&modal=detail&selectedIssue=NT-1217

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
+++ b/app/src/main/java/com/kickstarter/ui/fragments/PledgeFragment.kt
@@ -14,7 +14,6 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.TextView
-import android.widget.Toast
 import androidx.annotation.NonNull
 import androidx.core.content.ContextCompat
 import androidx.recyclerview.widget.LinearLayoutManager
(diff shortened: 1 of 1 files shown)


Arkariang:
This assumption is correct! 
We do a network call to V1 for getting the rules, and just load into the view the correct ones
Screen Shot 2020-06-09 at 8 04 54 AM

PR #903, opened 2020-06-18
NT-1346: Remove minimum pledge amount from no-reward cards 

JadeByfield89:
# 📲 What

Hide the minimum pledge amount from no-reward cards.

# 🤔 Why

We want to run experiments with suggested minimum amounts for different audiences, and this change is needed in order to efficiently run that experient.

# 🛠 How

I created a new Observable in RewardViewModelViewhHolder that emits whether or not a reward item is of the "no-reward" type. If that value is true, and it's a no-reward, we hide the amount and conversion text fields.

# 👀 See

Screen Shot 2020-06-17 at 8 23 03 PM

# 📋 QA

Check to ensure that the $ amount indicator has been removed from the “Pledge without a reward” card in the Reward Carousel for the control as well as all experimental variants 

# Story 📖

https://kickstarter.atlassian.net/secure/RapidBoard.jspa?rapidView=9&modal=detail&selectedIssue=NT-1346


JadeByfield89:
Closed due to incorrect base branch. Cherry picked commit onto proper branch instead

PR #920, opened 2020-07-09
NT-1411: Fix crash when creator refreshes while viewing a backer's pledge

JadeByfield89:
# 📲 What

When a creator goes to view details about a pledge to their project, the app was crashing on pull to refresh.

# 🤔 Why

The current logic in BackingFragment expects the parent activity to call the appropriate delegate function when refresh is triggered. This is implemented correctly in ProjectActivity, where the BackingFragment is loaded from in the user-context of a backer. For creators, BackingFragment is loaded by BackingActivity, which did not properly implement these delegate functions, so we were crashing due to a ClassCastException. 

# 🛠 How

I implemented the BackingFragment.BackingDelegate interface in BackingActivity, and updated BackingViewModel to re-fetch the backing on pull to refresh, and then update the UI.

# 👀 See

Hard to display this one, so no screenshots.

# 📋 QA

Log in as a creator, go to the Messages for a project you created, view a pledge, and refresh the layout. Confirm that no crash occurs.
# Story 📖

https://kickstarter.atlassian.net/secure/RapidBoard.jspa?rapidView=9&modal=detail&selectedIssue=NT-1411


Arkariang:
This should be merged against master, sorry for the branch mess 🥺 . I'll add the branches in the Jira ticket if that's useful

PR #934, opened 2020-07-23
[native] Update add-on strings with translations

leighdouglas:
# 📲 What

Updating strings with correct translations, with the exception of one string that has yet to be finalized. 

# 👀 See

Screenshot_1595535182

# 📋 QA

Change system language to any of the languages we support.
Open Kickstarter app, select your profile, and then select a backed project with add-ons. 
Tap manage. 
Tap the overflow menu and tap update pledge, for example. 
There you should see that all of the strings are translated appropriately.


leighdouglas:
Pushing against master instead of the feature branch, my mistake.

PR #956, opened 2020-08-24
NT-1390: Previously selected reward logic

Arkariang:
# 📲 What

- New logic for the CTA button string and when it should be activated.

# 👀 See
choose_another_reward
Screen Shot 2020-08-24 at 4 05 29 PM

|  |  |

# 📋 QA

- Change reward from one without Add-ons to another one without add-ons, no changes should be appreciated there on the button logic
- Change reward to the same reward with available add-ons, the button should be enabled
- Change reward to the same reward withOut available add-ons, the button should be disabled

# Story 📖

[NT-1390: Previously selected reward logic](https://kickstarter.atlassian.net/browse/NT-1390)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/RewardViewUtils.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/RewardViewUtils.kt
@@ -29,8 +29,8 @@ object RewardViewUtils {
 
         return when {
             BackingUtils.isBacked(project, reward) && !hasAddOnsSelected -> R.string.Selected
-            RewardUtils.isAvailable(project, reward) && !hasAddOnsSelected -> R.string.Select
-            reward.hasAddons() && hasAddOnsSelected && !hasUnavailableAddOn -> R.string.Continue
+            !BackingUtils.isBacked(project, reward) && RewardUtils.isAvailable(project, reward) -> R.string.Select
+            BackingUtils.isBacked(project, reward) && reward.hasAddons() && hasAddOnsSelected && !hasUnavailableAddOn -> R.string.Continue
             else -> R.string.No_longer_available
         }
     }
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/kickstarter/libs/utils/RewardViewUtils.kt
```diff
@@ -22,9 +22,15 @@ object RewardViewUtils {
      */
     @StringRes
     fun pledgeButtonText(project: Project, reward: Reward): Int {
+        val hasAddOnsSelected = project.backing()?.addOns()?.isNotEmpty() ?: false
+        val hasUnavailableAddOn:Boolean = project.backing()?.addOns()?.firstOrNull {
+            !RewardUtils.isAvailable(project, it)
+        }?.let { true } ?: false
```
leighdouglas, line 28: Would love to go over this with you to understand what exactly is happening here.

app/src/main/java/com/kickstarter/mock/factories/RewardFactory.java
```diff
@@ -16,6 +16,7 @@ private RewardFactory() {}
   public static @NonNull Reward addOn() {
     return reward().toBuilder()
             .isAddOn(true)
+            .limit(10)
```
leighdouglas, line 19: what is the limit field and why is it 10?

Arkariang: Is the addOns Limit, it wont go over 10. 
RewardFactory holds mocks for tests, I added the limit for the addOn object because there are some unit test checking availability for addOns using the limit.

app/src/main/java/com/kickstarter/libs/utils/RewardViewUtils.kt
Arkariang: In there I'm checking is any of the addOns is unavailable.
- backing.addOns is an optional list of rewards
- I use firstOrNull so it wont iterate over the entire list in case some addOn is unavailable, if there is some unavailable addOn on the let block return true. Null in case all the addOns are available so with the elvis operator `?:` sets the `hasUnavailableAddOn` to false

PR #961, opened 2020-08-27
NT-1440:Display Empty State when add-ons unavailable

leighdouglas:
# 📲 What

Update UI to display an empty state if the user selects a shipping location on the add-ons screen that has no available add-ons.

# 🤔 Why

To give context and instructions to user as to why add-ons are unavailable for selected reward. 

# 🛠 How

Backing addon fragment subscribes to a new stream that emits the current state depending on if the emitted add-ons list for selected location is empty. When this list is empty, the add-ons adapter displays an error message. 

# 👀 See
Screenshot_1598548351

# 📋 QA

- Search for the project "Hovering Album Art"
- Select "The full reward package!" reward
- Select "spain" as the location

# Story 📖

[Jira Ticket](https://kickstarter.atlassian.net/secure/RapidBoard.jspa?rapidView=9&projectKey=NT&modal=detail&selectedIssue=NT-1440)


app/src/test/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModelTest.kt
```diff
@@ -522,6 +524,77 @@ class BackingAddOnsFragmentViewModelTest : KSRobolectricTestCase() {
...
+        val project = ProjectFactory.project().toBuilder().rewards(listOf(rw)).build()
+        val projectData = ProjectDataFactory.project(project, null, null)
+        val pledgeReason = PledgeFlowContext.forPledgeReason(PledgeReason.PLEDGE)
+
+        val bundle = Bundle()
+        bundle.putParcelable(ArgumentsKey.PLEDGE_PLEDGE_DATA, PledgeData.with(pledgeReason, projectData, rw))
+        bundle.putSerializable(ArgumentsKey.PLEDGE_PLEDGE_REASON, PledgeReason.PLEDGE)
+        this.vm.arguments(bundle)
+
+        this.addOnsList.assertValue(Triple(projectData, emptyList(), shippingRuleRw))
```
Arkariang, line 559: something off with this test. If the empty state is emiting true, the addOnslist should be empty

Arkariang: My bad, I did't realize the second value is the empty one! that's a great work :)

PR #964, opened 2020-09-01
NT-1489: Fix Crash when a user views their Account

jgsamudio:
# 📲 What

Fixed a crash when user's try to access their account.

# 🤔 Why

In Firebase, we received crashes related to the ArrayAdapter. 
[Link to Crash](https://console.firebase.google.com/u/0/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/58ec151aaa1b8752aa817cdf9325fbb7?time=last-seven-days&sessionId=5F4E6DF500E8000157AD6BA2371B8A38_DNE_0_v2)

# 🛠 How

Instead of using the CurrencyCode enum that is generated by graphQL, we are going to use a list a supported enums so in the case where a new currency is added to the backend CurrencyCode we can still display the currencies without a crash.

The reason for the crash was do to a new currency that was added (PLN). The team updated the graphQL models and the next release was unable to handle the new currency. 

[Link to Currency Update PR](https://github.com/kickstarter/kickstarter/pull/19698)

# 📋 QA

Navigate to Settings, then Account and change the user's selected currency

# Story 📖

[NT-1489](https://kickstarter.atlassian.net/browse/NT-1489)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/AccountActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/AccountActivity.kt
@@ -28,6 +28,25 @@ class AccountActivity : BaseActivity<AccountViewModel.ViewModel>() {
 
     private lateinit var ksString: KSString
 
+    private val supportedCurrencies: List<CurrencyCode> by lazy {
+        arrayListOf(
+                CurrencyCode.AUD,
+                CurrencyCode.CAD,
+                CurrencyCode.CHF,
+                CurrencyCode.DKK,
+                CurrencyCode.EUR,
+                CurrencyCode.GBP,
+                CurrencyCode.HKD,
+                CurrencyCode.JPY,
+                CurrencyCode.MXN,
+                CurrencyCode.NOK,
+                CurrencyCode.NZD,
+                CurrencyCode.SEK,
+                CurrencyCode.SGD,
+                CurrencyCode.USD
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/kickstarter/ui/activities/AccountActivity.kt
```diff
@@ -143,8 +163,8 @@ class AccountActivity : BaseActivity<AccountViewModel.ViewModel>() {
             override fun onNothingSelected(parent: AdapterView<*>?) {}
 
             override fun onItemSelected(parent: AdapterView<*>?, view: View?, postion: Int, id: Long) {
-                if (currentCurrencySelection != null && currentCurrencySelection!!.ordinal != postion) {
-                    newCurrencySelection = CurrencyCode.values()[postion]
+                if (currentCurrencySelection != null && supportedCurrencies.indexOf(currentCurrencySelection!!) != postion) {
```
Arkariang, line 166: Can you unwrap the currentCurrencySelection, so we don't need to use that forcing non optional `!!`
`currentCurrencySelection?.let {
                    if (supportedCurrencies.indexOf(it) != postion) {
                        newCurrencySelection = supportedCurrencies[postion]
                        lazyFollowingOptOutConfirmationDialog().show()
                    }
                }`

PR #965, opened 2020-09-01
NT-1417: Show a native alert during Edit Reward if changes would remove add ons

Arkariang:
# 📲 What

- Let the backer know when changing rewards that selection could potentially loose some the the selected addOns

# 🛠 How

- Create alert dialog on RewardsFragment
- Make RewardsFragmentViewModel aware of the previous selection state

# 👀 See

| Before 🐛 | After 🦋 |
Screen Shot 2020-09-01 at 10 07 27 AM

|  |  |

# 📋 QA

- Change from a reward with Add-ons to a reward not offering them
- Edit your current reward to edit your addOns selection
- Change from a reward without AddOns to a reward with AddOns

# Story 📖

[NT-1417: Show a native alert during Edit Reward if changes would remove add ons](https://kickstarter.atlassian.net/browse/NT-1417)


app/src/main/java/com/kickstarter/ui/fragments/RewardsFragment.kt
```diff
@@ -48,25 +51,54 @@ class RewardsFragment : BaseFragment<RewardsFragmentViewModel.ViewModel>(), Rewa
...
+                    .setMessage(getString(R.string.It_may_not_offer_some_or_all_of_your_add_ons))
+                    .setNegativeButton("             ${getString(R.string.No_go_back)}") { _, _ -> {} }
+                    .setPositiveButton("             ${getString(R.string.Yes_continue)}") { _, _ ->
+                        this.viewModel.inputs.alertButtonPressed()
+                    }.create()
+        }
+    }
+
+    private fun showAlert() {
+        dialog.create()
```
leighdouglas, line 98: do we need to call `create()` here? We are also calling it in the `createDialog()` function, so it would be called twice.

app/src/main/java/com/kickstarter/viewmodels/RewardsFragmentViewModel.kt
```diff
@@ -119,6 +184,8 @@ class RewardsFragmentViewModel {
             this.rewardClicked.onNext(reward)
         }
 
+        override fun alertButtonPressed() = this.alertButtonPressed.onNext(null)
```
leighdouglas, line 187: Just curious about sending a null value through this stream, instead of `any`

app/src/main/java/com/kickstarter/ui/fragments/RewardsFragment.kt
Arkariang: You are completely right! removing it 👍

app/src/main/java/com/kickstarter/viewmodels/RewardsFragmentViewModel.kt
Arkariang: Any implies all possible types (including null), and you don't need to create a new object sending null. I won't mark as incorrect sending an any object though

PR #967, opened 2020-09-01
NT-1411: Crash on Creator Pledge Screen

jgsamudio:
# 📲 What

Added a basic add-on if the backer pledge without a reward. 

# 🤔 Why

The crash was due to the basic reward missing "id" and "minimum". Without this information the builder crashed.

# 🛠 How

In order to fix the issue the "id" and "minimum" was added to the reward object to correct pass a completed reward object.

# 📋 QA

Using the credentials for the creator account in the ticket below, navigate to the Dashboard, then messages and then select the first message. After that select the "view pledge" CTA

# Story 📖

[NT-1411](https://kickstarter.atlassian.net/browse/NT-1411)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/fragments/BackingFragment.kt
+++ b/app/src/main/java/com/kickstarter/ui/fragments/BackingFragment.kt
@@ -210,7 +210,7 @@ class BackingFragment : BaseFragment<BackingFragmentViewModel.ViewModel>() {
                     pledge_details_label.text = getString(R.string.Pledge_details)
                     ViewUtils.setGone(received_section, true)
                     ViewUtils.setGone(delivery_disclaimer_section, it)
-                    ViewUtils.setGone(estimated_delivery_label_2, false)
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kickstarter/ui/fragments/BackingFragment.kt
```diff
@@ -210,7 +210,7 @@ class BackingFragment : BaseFragment<BackingFragmentViewModel.ViewModel>() {
                     pledge_details_label.text = getString(R.string.Pledge_details)
                     ViewUtils.setGone(received_section, true)
                     ViewUtils.setGone(delivery_disclaimer_section, it)
-                    ViewUtils.setGone(estimated_delivery_label_2, false)
+                    ViewUtils.setGone(estimated_delivery_label_2, true)
```
jgsamudio, line 213: Change this to `true` so that the estimated label would also disappear when the disclaimer section is gone

PR #971, opened 2020-09-03
NT-1387: Fix - Select AddOns for digital reward

Arkariang:
# 📲 What

- A digital reward can have digital AddOns.

# 🤔 Why

- the current change is needed because we first assumed that AddOns where available just for shippable rewards.

# 🛠 How

- And empty shipping location has been created in case a digital reward is selected and it has add-Ons, doing so the extra logic needed for the AddOns screen is minimum
- Modified old test, and added new ones to cover this situation
- the RewardUtils.isDigital() method is now covering the logic for both models V1 and GraphQL rewards, before it was giving us false negatives, keep in mind the selected reward is from V1, the loaded addOns from GraphQL

# 👀 See

select_digital_addOn

|  |  |

# 📋 QA

- Use any project with AddOns, this concrete one has a digital reward with digital addOns: [one inch straw - Loose reusable units - GROW IT from 1"- 10"](https://staging.kickstarter.com/projects/eserrano/one-inch-straw-loose-reusable-units-grow-it-from-1-10?ref=nav_search&result=project&term=one%20inch%20straw)
- Check we haven't introduce any regression for all the other use cases, this project has a bunch of AddOns to test with [GHOST BAND: A 96 page graphic novel performance!](https://staging.kickstarter.com/projects/ramonesome/ghost-band?ref=nav_search&result=project&term=ghost%20band)

# Story 📖

[Filter AddOns by location](https://kickstarter.atlassian.net/browse/NT-1387)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/RewardUtils.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/RewardUtils.java
@@ -17,6 +17,8 @@ import java.util.List;
 
 import androidx.annotation.NonNull;
 
+import static com.kickstarter.models.Reward.SHIPPING_TYPE_NO_SHIPPING;
+
 public final class RewardUtils {
   private RewardUtils() {}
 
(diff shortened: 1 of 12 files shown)


leighdouglas (approved):
👍 lgtm!

PR #973, opened 2020-09-04
NT-1482: Add-Ons Analytics Events

jgsamudio:
# 📲 What

Add the Add-Ons analytics events to the app. 

# 🤔 Why

Needed to track the impact of Add-Ons

# 🛠 How

Added Add-Ons to the backing add-on fragment with tests.

# Story 📖

[NT-1482](https://kickstarter.atlassian.net/browse/NT-1482)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
@@ -82,7 +82,7 @@ class BackingAddOnsFragmentViewModel {
         private val addOnsListFiltered = PublishSubject.create<Triple<ProjectData, List<Reward>, ShippingRule>>()
         private val isEmptyState = PublishSubject.create<Boolean>()
         private val totalSelectedAddOns = BehaviorSubject.create(0)
-        private val continueButtonPressed = BehaviorSubject.create<Void>()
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
```diff
@@ -252,12 +257,17 @@ class BackingAddOnsFragmentViewModel {
                                 .build()
                     }
                 }
-
                 return@combineLatest Pair(updatedPledgeData, pledgeReason)
             }
+
+            this.continueButtonPressed
+                    .compose<Pair<Boolean, Pair<PledgeData, PledgeReason>>>(combineLatestPair(pledgeDataAndReason))
+                    .filter { it.first }
```
jgsamudio, line 265: Fixed an issue where the user selects an add-on, presses continue, then goes back to the add-ons screen and changes their add-on selection. When the add-ons selection is changed, the view automatically transitions to the pledge screen. The transition should only happen when selecting the continue button

Arkariang: that one is also fixed on my PR, I'll add you to it

Arkariang (approved):
LGTM! great work as usual @jgsamudio go ahead and merge it, I'll fix the conflicts in my PR once this one is merged

PR #975, opened 2020-09-08
[NT-1417][NT-1510]:Fix change reward flow

Arkariang:
# 📲 What

- We were using the backing shipping location when choosing a new reward, that's incorrect, the shippingLocation should be always the one from the selected reward, except if it's digital, then hide shipping selector.
- Unified shippingType & shippingPreference for both Reward from V1 and Reward from Graph. When doing a pledge for the first time the model for Reward is from V1, when taking the reward from the backingObject the model is from Graph there where some discrepancies on the ShippingPreference and shippingType.

# 🛠 How

- In case the reward has restrited shipping rules, and none of those rules matches our default location, we load the first one of the Reward shipping rules
- In case the reward is unrestricted, load the defaultLocation
Those two AC are true for both flows new pledge and choosing new reward.
- Updated logic for showing the Native alert on choosing another reward flow to check the id of the selected reward not the same as the backed one (Do not show alert in this case)
- Updated logic for showing the Native alert to be shown if the selected reward has different shipping rules from the backed one.
- show the alert in case we move from a reward with AddOns to a NoReward reward.
- showing back the name of the selected shippingrule in PledgeFragment.

# 👀 See

| Before 🐛 | After 🦋 |
change-reward-flow

|  |  |

# 📋 QA

- Update reward from digital to unrestricted with addOns
- update reward from unrestricted to restricted
- update reward from restricted to digital
- updated from any of the previous one to a digital without addOns
- updated from any of the previous one to a regular reward without addOns
- updated from any of the previous one to a NoReward

# Story 📖

[NT-1417](https://kickstarter.atlassian.net/browse/NT-1417)
[NT-1510](https://kickstarter.atlassian.net/browse/NT-1510)


jgsamudio:
When selecting a reward in the "one inch straw" project, I noticed that the rewards headers scrolls because the price is moved to a multiple line. Is this related to this PR or should we create a ticket for this?

Screen Shot 2020-09-08 at 7 27 10 PM

jgsamudio:
When selecting another reward that is not the currently selected reward, the app transitions to the manage pledge screen and also presents the alert. We want to present the alert without transitioning to the next screen. Only when the user selects, "yes, continue" that we want to transition to the next screen.

app/src/main/java/com/kickstarter/viewmodels/RewardsFragmentViewModel.kt
```diff
@@ -158,6 +179,12 @@ class RewardsFragmentViewModel {
                     }
         }
 
+        private fun differentShippingTypes(newRW: Reward, backedRW: Reward): Boolean =
```
jgsamudio, line 182: This is a little more readable. I normally only like to use the inline notation for one liners. Code that requires more than one line should be in a block.

```
private fun differentShippingTypes(newRW: Reward, backedRW: Reward): Boolean {
    if (newRW.id() != backedRW.id()) {
        return newRW.shippingType()?.toLowerCase() ?: "" != backedRW.shippingType()?.toLowerCase() ?: ""
    }
    return false
}
```

Arkariang: I don't like the two returns but ok with it being in a block 👍

Arkariang:
> When selecting another reward that is not the currently selected reward, the app transitions to the manage pledge screen and also presents the alert. We want to present the alert without transitioning to the next screen. Only when the user selects, "yes, continue" that we want to transition to the next screen.

Fixed! and simplified logic :)

jgsamudio (approved):
LGTM! Great Work!

PR #976, opened 2020-09-09
[NT-1474]: Refresh backing fragment

Arkariang:
# 📲 What

- The backing fragment was not reflecting the update pledge changes, or changing reward changes unless fragment re-created again

- Allow to load emptyList on AddOns section for the rewards and addOns recyclerView (changing your addOns selection to 0 addOns, or changing from a reward with addOns to one without it was not updating the RecyclerView UI).

- Do not present Any fragment or Dialog unless the current fragment is visible.

# 🛠 How

- Added .materialize() operator when fetching the BackingObject in ProjectViewModel

# 👀 See
refreshed-backing-fragment
|  |  |

# 📋 QA

- Change from a reward with AddOns to a Reward without AddOns, the BackingFragment should change without any user interaction to reflect the new selection (look at the gif if any doubts)

# Story 📖

[NT-1474](https://kickstarter.atlassian.net/browse/NT-1474)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/adapters/RewardAndAddOnsAdapter.kt
+++ b/app/src/main/java/com/kickstarter/ui/adapters/RewardAndAddOnsAdapter.kt
@@ -29,10 +29,8 @@ class RewardAndAddOnsAdapter() : KSAdapter() {
     }
 
     fun populateDataForAddOns(rewards: List<Pair<ProjectData,Reward>>) {
-        if (rewards.isNotEmpty()) {
-            setSection(SECTION_ADD_ONS_CARD, rewards)
-            notifyDataSetChanged()
-        }
+        setSection(SECTION_ADD_ONS_CARD, rewards)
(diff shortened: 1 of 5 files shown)


jgsamudio (commented):
I noticed when I back a project, then change the backing to another reward and then cancel that reward that the "you backed" tag is not updated. I think the recycler view just needs to be updated again when its presented

Arkariang:
> I noticed when I back a project, then change the backing to another reward and then cancel that reward that the "you backed" tag is not updated. I think the recycler view just needs to be updated again when its presented

I'm gonna need another ticket for that, after looking at it deeply,  when canceling a pledge the fragments are not refreshed, I've tried backing a project and canceling pledge, and the same issue raised. Good catch @jgsamudio 👍 
It will be worked in here https://kickstarter.atlassian.net/browse/NT-1520

jgsamudio (approved):

PR #979, opened 2020-09-11
[NT-1520] - Update UI in RewardsFragment when canceling a pledge

Arkariang:
# 📲 What

- ProjectData was always updated with the outdated backing object

# 🤔 Why

- We fetch project from V1, but backing Object from Graph. In this situation the observable for backingObject still had the previous emission, so added additional logic to make sure we update the project data only if needed

# 👀 See

cancel-pledge-updating-fragments

# 📋 QA

- Pledge to a project, hit manage your pledge, cancel the pledge. Then hit again back this project button, the reward carousel should be in a non backed state without the "You backed" tag

# Story 📖

[NT-1520](https://kickstarter.atlassian.net/browse/NT-1520)


leighdouglas (approved):
Looks good!

PR #980, opened 2020-09-14
NT-1442: Android Display an error state if add-ons fail to load

leighdouglas:
# 📲 What

Show an error alert dialog when there is an error retrieving the add-ons. Also allow user to tap retry to resend the networks calls to get the add-ons. 

# 🤔 Why

To give user more context on why add-ons are not loading correctly. 

# 🛠 How

Created a new stream that triggers the dialog when the network response is an error. Retry button resends the project data and reason to make a network call.

# 👀 See

Screenshot_1600286472

# 📋 QA

- Search for the project "one inch straw"
- Select "back project"
- Turn off wifi or put phone in airplane mode
- Select a pledge with add-ons
- Wait to see the error dialog
- Turn on wifi or turn off airplane mode
- Press "retry"
- After network call finishes, should show add-ons

# Story 📖

[NT-1442](https://kickstarter.atlassian.net/secure/RapidBoard.jspa?rapidView=9&projectKey=NT&modal=detail&selectedIssue=NT-1442&assignee=5efe5a70e0043f0bacaf7706&assignee=5eda9c57222d4d0b8296577e)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/fragments/BackingAddOnsFragment.kt
+++ b/app/src/main/java/com/kickstarter/ui/fragments/BackingAddOnsFragment.kt
@@ -1,6 +1,7 @@
 package com.kickstarter.ui.fragments
 
 import android.os.Bundle
+import android.util.Log
 import android.util.Pair
 import android.view.LayoutInflater
 import android.view.View
@@ -121,6 +122,8 @@ class BackingAddOnsFragment : BaseFragment<BackingAddOnsFragmentViewModel.ViewMo
     }
 
     private fun showErrorDialog() {
+        Log.d("HELLOWORLD", "SHOW Error")
+
         if (!errorDialog.isShowing) {
             errorDialog.show()
         }
--- a/app/src/main/java/com/kickstarter/ui/fragments/BackingAddOnsFragment.kt
+++ b/app/src/main/java/com/kickstarter/ui/fragments/BackingAddOnsFragment.kt
@@ -46,7 +46,6 @@ class BackingAddOnsFragment : BaseFragment<BackingAddOnsFragmentViewModel.ViewMo
         super.onViewCreated(view, savedInstanceState)
         setUpShippingAdapter()
         setupRecyclerView()
-
         setupErrorDialog()
 
         this.viewModel.outputs.showPledgeFragment()
@@ -122,7 +121,9 @@ class BackingAddOnsFragment : BaseFragment<BackingAddOnsFragmentViewModel.ViewMo
     }
 
     private fun showErrorDialog() {
-        errorDialog.show()
+        if (!errorDialog.isShowing) {
+            errorDialog.show()
+        }
     }
 
     private fun dismissErrorDialog() {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/ui/fragments/BackingAddOnsFragment.kt
```diff
@@ -1,10 +1,12 @@
 package com.kickstarter.ui.fragments
 
 import android.os.Bundle
+import android.util.Log
```
jgsamudio, line 4: Remove log if not needed

app/src/main/java/com/kickstarter/ui/fragments/BackingAddOnsFragment.kt
```diff
@@ -150,6 +168,18 @@ class BackingAddOnsFragment : BaseFragment<BackingAddOnsFragmentViewModel.ViewMo
                 ?.commit()
     }
 
+    private fun setupErrorDialog() {
+        context?.let { context ->
+            errorDialog = AlertDialog.Builder(context, R.style.AlertDialog)
+                    .setCancelable(false)
+                    .setTitle(getString(R.string.Something_went_wrong_please_try_again))
+                    .setPositiveButton("                     ${getString(R.string.Retry)}") { _, _ ->
```
jgsamudio, line 176: Is there a way of setting the text of the positive and negative buttons without the front space?

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
```diff
@@ -1,11 +1,11 @@
 package com.kickstarter.viewmodels
 
+import android.util.Log
```
jgsamudio, line 3: Remove log if not needed

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
```diff
@@ -42,6 +41,9 @@ class BackingAddOnsFragmentViewModel {
 
         /** Emits the quantity per AddOn Id selected */
         fun quantityPerId(quantityPerId: Pair<Int, Long>)
+
+        /**Envoked when the retry button on the addOn Error alert dialog is pressed */
```
jgsamudio, line 45:
```suggestion
        /** Invoked when the retry button on the add-on Error alert dialog is pressed */
```

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
```diff
@@ -68,26 +70,34 @@ class BackingAddOnsFragmentViewModel {
 
         /** Emits whether or not the empty state should be shown **/
         fun isEmptyState(): Observable<Boolean>
+
+        /**Emits an alert dialog when add-ons request results in error **/
```
jgsamudio, line 74:
```suggestion
        /** Emits an alert dialog when add-ons request results in error **/
```

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
```diff
@@ -175,11 +185,35 @@ class BackingAddOnsFragmentViewModel {
             shippingRules
                     .filter { it.isNotEmpty() }
                     .compose<Pair<List<ShippingRule>, PledgeReason>>(combineLatestPair(pledgeReason))
+                    .filter { it.second == PledgeReason.PLEDGE }
```
jgsamudio, line 188: Why was this added?

jgsamudio (changes requested):

app/src/main/java/com/kickstarter/ui/fragments/BackingAddOnsFragment.kt
leighdouglas: This is the way that we set them so that they stack on top of one another as opposed to next to one another. Not sure there is another way to do this unless we create a custom layout for this alert dialog.

app/src/main/java/com/kickstarter/ui/fragments/BackingAddOnsFragment.kt
leighdouglas: 👍

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
leighdouglas: 👍

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
leighdouglas: Removed

app/src/main/java/com/kickstarter/ui/fragments/BackingAddOnsFragment.kt
jgsamudio: I found this stack overflow article 
https://stackoverflow.com/questions/35880320/how-to-stack-alertdialog-buttons-vertically
and I agree with it, we shouldn't add spaces since the OS can change this. What happens if on a Galaxy phone the spaces don't look right? 

There's an example of a custom layout. What do you think about implementing that custom layout? The alert is simple. A text and 2 buttons.

leighdouglas: Will resolve in subsequent PR

jgsamudio (commented):
I noticed that it takes a long time to present when the internet is off. It could be because of the behaviour / publish subject issue that we saw before? Can you verify this?

leighdouglas:
@jgsamudio When putting it in airplane mode on the virtual device, the error dialog popped up immediately. I then tested this behavior by turning off my wifi and the error dialog took longer to popup - consistent with your comment. I tested this with both `BehaviorSubject` and `PublishSubject`, and both behaved the same.

jgsamudio (approved):

PR #981, opened 2020-09-21
NT-1516: Update the Bonus Support Base Amount

Arkariang:
# 📲 What

- The minimum pledge allowed is specified by country, but when we added the RewardMinimum experiment we specified for the Control variant the amount "1". Changing involuntarily for some countries like Mexico (with minPledge 10MX ) the NoReward amount when control variant active to 1MX.
- Previously the stepper increased amount was the minPledge, now is always increased by 1.

# 🤔 Why

If control variant active the No Reward pledge amount was incorrect for some countries with minimum pledge higher than the control variant (IE: Mexico or japan with minPledges 10MX, and 100¥)

# 🛠 How

- When the stepper button is pressed just update by 1 the amount
- In RewardUtils if the amount active in the experiment is lower than the minimumPledge, use the minPledge value.

# 👀 See
- Mexico 
min_pledge_and_stepper

- Japan
japan_min

|  |  |

# 📋 QA

- Pledge to a Mexican or japanese project, with a No Reward reward, make sure the amount we show in the pledgeInput is 10Mx or 100¥
- make sure the stepper increase the amount by 1.

# Story 📖

[Update the Bonus Support Base Amount](https://kickstarter.atlassian.net/browse/NT-1516)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/RewardUtils.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/RewardUtils.java
@@ -202,14 +202,16 @@ public final class RewardUtils {
   }
 
   /**
-   * Returns the amount value for each variant, being Control the original value
+   * Returns the amount value for each variant, being Control the original value, and the minimum
+   * the minPledge defined by country
    *
    * @param variant the variant for which you want to get the value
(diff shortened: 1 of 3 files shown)


leighdouglas (approved):
LGTM!

PR #986, opened 2020-09-23
Android NT-1442: Hide shipping selector on network error

leighdouglas:
# 📲 What

We want to hide the shipping selector when there is a network call error to fetch addons, and only show when addons are fetched correctly. 

# 🛠 How

Update `shippingSelectorGone` stream when there is a network error. 

# 👀 See

shippingSelectorGone

# 📋 QA

For error alert dialog:

- Search for the project "fadoodle"
- Select "back project"
- Turn off wifi or put phone in airplane mode
- Select a pledge with add-ons
- Upon error dialog popping up, location selector should disappear
- Before tapping "retry", turn wifi on or take phone off airplane mode
- Press "retry"
- Network call should be successful and location selector should reappear

# Story 📖

[Android Display an error state if add-ons fail to load
](https://kickstarter.atlassian.net/browse/NT-1442)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
@@ -8,6 +8,7 @@ import com.kickstarter.libs.KSString
 import com.kickstarter.libs.rx.transformers.Transformers.combineLatestPair
 import com.kickstarter.libs.rx.transformers.Transformers.takeWhen
 import com.kickstarter.libs.utils.ObjectUtils
+import com.kickstarter.libs.utils.RewardUtils
 import com.kickstarter.libs.utils.RewardUtils.isDigital
 import com.kickstarter.libs.utils.RewardUtils.isShippable
 import com.kickstarter.mock.factories.ShippingRuleFactory
@@ -218,7 +219,6 @@ class BackingAddOnsFragmentViewModel {
                     .filter { ObjectUtils.isNotNull(it) }
                     .subscribe {
                         addOnsFromGraph.onNext(it)
-                        this.shippingSelectorIsGone.onNext(false)
                     }
 
             val filteredAddOns = Observable.combineLatest(addonsList, projectData, this.shippingRuleSelected, reward, this.totalSelectedAddOns) { list, pData, rule, rw,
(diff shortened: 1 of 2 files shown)


colleenmacd (approved):

Arkariang (approved):
LGTM!

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
```diff
@@ -191,27 +191,35 @@ class BackingAddOnsFragmentViewModel {
...
+                    }
 
             Observable
                     .combineLatest(this.retryButtonPressed.startWith(false), project) { _, pj ->
                         return@combineLatest this.apolloClient
                                 .getProjectAddOns(pj.slug()?.let { it } ?: "")
-                                .doOnError { this.showErrorDialog.onNext(true) }
+                                .doOnError {
+                                    this.showErrorDialog.onNext(true)
+                                    this.shippingSelectorIsGone.onNext(true)}
```
Arkariang, line 214: No need to update the shippingSelector here

leighdouglas: removed!

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
```diff
@@ -191,27 +192,34 @@ class BackingAddOnsFragmentViewModel {
...
                                 .onErrorResumeNext(Observable.empty())
                     }
                     .switchMap { it }
                     .filter { ObjectUtils.isNotNull(it) }
-                    .map { listAddOns ->
```
Arkariang, line 211: This should remain as it was. This is the filter in case is not available the addOn

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnsFragmentViewModel.kt
Arkariang: Sorry my bad! that is actually correct, we update the visibility of the shipping selector if any of the two network calls fail. You're where doing it correctly ignore my previous comment

PR #988, opened 2020-09-23
NT-1542: Pledge Amount Missing from Manage Pledge

jgsamudio:
# 📲 What

In the manage pledge view, the pledge line item does not show up when the user backs for a digital reward.

# 🤔 Why

The user should be able to see what their pledged amount is and how that value is separate from the total value and the bonus support.

# 🛠 How

The logic to determine if the pledge line should be visible is no longer based on the location id but on if the reward on the backing object is null.

# 📋 QA

View a project that has a digital reward and back it. Navigate to the manage pledge view to verify the pledge line item.

# Story 📖

[NT-1542](https://kickstarter.atlassian.net/browse/NT-1542)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
@@ -235,8 +235,15 @@ interface BackingFragmentViewModel {
                     .distinctUntilChanged()
                     .compose(bindToLifecycle())
                     .subscribe {
-                        this.pledgeSummaryIsGone.onNext(it)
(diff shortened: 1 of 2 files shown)


Arkariang (approved):
LGTM

Arkariang (commented):
Could you add a check for isNoReward, or consult with @colleenmacd  if the Reward no Reward should show the same text as the regular rewards

jgsamudio:
> Could you add a check for isNoReward, or consult with @colleenmacd if the Reward no Reward should show the same text as the regular rewards

Confirmed with Colleen that the text can be the same as the regular rewards.

PR #989, opened 2020-09-23
NT-1497: Preserve Bonus Support when Editing a Reward with Add-ons

jgsamudio:
# 📲 What

Updated the PledgeFragment to check the backing object on the main PledgeData as well as the project object. 

# 🤔 Why

As a user, I may only want to change a portion of the backing and resetting the bonus support is an action that the user might not expect.

# 🛠 How

In the PledgeFragment view, the bonus amount is only checked on the top backing object where in some cases that is null, in those instances we check the project's backing model for the bonus support.

# 📋 QA

- Back a project with add-ons and add a bonus amount.
- Then manage that reward
- Select "choose another reward"
- Select the same reward and update the number of add-ons
- Continue and verify that the original bonus support is present before re-pledging

# Story 📖

[NT-1497](https://kickstarter.atlassian.net/browse/NT-1497)


Arkariang (approved):
LGTM!

PR #990, opened 2020-09-24
NT-1539: Unprompted Edit Rewards Alert

jgsamudio:
# 📲 What

Added a additional filter and check to prevent the alert from presenting when navigated away from the app.

# 🤔 Why

# 🛠 How

The project activity notifies the other sub fragments when the user resumes their session and this toggles an update to the reward clicked stream. If the user already prompted the modal before, this modal will come up again. The fix was to apply a filter on the reward clicked so that the alert only shows when it is prompted from a selection of the reward

# 📋 QA

# Story 📖


jgsamudio:
Will close this PR to allow for a simpler solution that Isabel found. We need to spend more time refactoring the streams as a whole

PR #991, opened 2020-09-24
NT-1471 | NT-1539 - Prevent Add-ons Alert & Add Pledge Label

jgsamudio:
# 📲 What

The modal will now only be presented when the user selects another reward while their current reward has add-ons. The Pledge label will be shown when the user wants to update their pledge.

# 🤔 Why

# 🛠 How

Added a check on the backing stream to only fire when it is a unique change. For the manage pledge view, we are showing the pledge label. 

# 📋 QA

# Story 📖

[NT-1471](https://kickstarter.atlassian.net/browse/NT-1471)


Arkariang (approved):

leighdouglas (approved):
🎉 🎉 🎉

PR #992, opened 2020-09-25
NT-1390: Disabled button first load changing reward with addOns

Arkariang:
# 📲 What

- When no changes in your previously selected addOns continue button disabled

# 🛠 How

- Before the validation was not being triggered when backed reward is digital. Fixed that
- If same addOns selected and no updates, but new item is selected button should be enabled

# 👀 See
choose-another-rw

|  |  |

# 📋 QA

- pledge to a digital reward with add ons
- change reward and select same reward
- check in addOns fragment button is disabled until update preSelected addOns or added new one.

# Story 📖

[NT-1390](https://kickstarter.atlassian.net/browse/NT-1390?focusedCommentId=17873)


jgsamudio (approved):
LGTM! 👍

PR #994, opened 2020-09-25
NT-1539: Unprompted Edit Rewards Alert

jgsamudio:
# 📲 What

Added a additional filter and check to prevent the alert from presenting when navigated away from the app.

# 🤔 Why

# 🛠 How

The project activity notifies the other sub fragments when the user resumes their session and this toggles an update to the reward clicked stream. If the user already prompted the modal before, this modal will come up again. The fix was to apply a filter on the reward clicked so that the alert only shows when it is prompted from a selection of the reward

# 📋 QA

# Story 📖

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/RewardsFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/RewardsFragmentViewModel.kt
@@ -56,7 +56,7 @@ class RewardsFragmentViewModel {
     class ViewModel(@NonNull val environment: Environment) : FragmentViewModel<RewardsFragment>(environment), Inputs, Outputs {
 
         private val projectDataInput = PublishSubject.create<ProjectData>()
-        private var rewardClicked = PublishSubject.create<Pair<Reward, Boolean>>()
+        private val rewardClicked = BehaviorSubject.create<Pair<Reward, Boolean>>()
         private val alertButtonPressed = PublishSubject.create<Void>()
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kickstarter/viewmodels/RewardsFragmentViewModel.kt
```diff
@@ -78,42 +76,79 @@ class RewardsFragmentViewModel {
...
                     .map { indexOfBackedReward(it) }
                     .distinctUntilChanged()
                     .compose(bindToLifecycle())
                     .subscribe(this.backedRewardPosition)
 
-            val pledgeDataAndReason = this.projectDataInput
-                    .compose<Pair<ProjectData, Reward>>(Transformers.takePairWhen(this.rewardClicked))
-                    .map { pledgeDataAndPledgeReason(it.first, it.second) }
-                    .distinctUntilChanged()
+            subscribeRewardClicked()
```
jgsamudio, line 87: Split out the reward clicked setup

Arkariang: I'm gonna be hyper picky with this sorry 😞 , we can ha a chat if you want but, could we not split the init of the viewModel in functions? ... until now every viewModel has been following the same approach and just pure functions in the RXChains. Doing this we are breaking that consistency through the app.
To resume, could we not have the subscribeRewardClicked function and bring back to the constructor all the RXChains?

Arkariang (approved):

PR #995, opened 2020-09-25
NT-1541: Updated Add-ons Quantity Not Updated

Arkariang:
# 📲 What

- False negative when using .distinctUntilChanged()
- Clean up to avoid memory leaks in the fragments.

# 🛠 How
- Remove the distinctUntilChanged operator
- When onDetach is called remove the viewModel.

# 👀 See
updatedQuantity

|  |  |

# 📋 QA

- Choose the same reward, update your addOns, choose the same reward again without leaving manage your pledge screen, make sure the addOns screen is reflecting your changes.

# Story 📖

[Updated Add-ons Quantity Not Updated](https://kickstarter.atlassian.net/browse/NT-1541)


jgsamudio:
@Arkariang When I enter a project and I go back the app is crashing

jgsamudio (approved):

PR #996, opened 2020-09-25
NT-1534: Android Sold-out add-ons fix

Arkariang:
# 📲 What

- If you backed addOns, the Choose new reward let you ALWAYS edit that same reward to modify your addons selection, no matter the state of that reward.

# 🤔 Why
- You may have the reward unavailable and still be able to update your addOns selection because the are available (availability of addOns to offer is handled in the AddOns Screen)

# 🛠 How
- Modified pre-existing logic

# 👀 See
- No actual project to test with that I've found, but test covering that use case updated and passing.

|  |  |

# 📋 QA

- I haven't found a project to test with but follow this scenario: 
1 - Find a limited reward with just one reward left if possible and with addOns (pledge to the reward + addons)
2- Choose another reward flow -> 
BEFORE: Choosing same reward was impossible because that reward wasn't selectable (disabled button).
NOW: You can choose that same reward, and update the addOns selection 

# Story 📖

[Android Sold-out add-ons fix](https://kickstarter.atlassian.net/browse/NT-1534)

### DIFF
--- a/app/src/main/java/com/kickstarter/mock/factories/ProjectFactory.java
+++ b/app/src/main/java/com/kickstarter/mock/factories/ProjectFactory.java
@@ -94,10 +94,10 @@ public final class ProjectFactory {
       .build();
   }
 
-  public static @NonNull Project backedProjectWithAddOnsLimitReached() {
+  public static @NonNull Project backedProjectWithRewardAndAddOnsLimitReached() {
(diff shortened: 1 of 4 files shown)


jgsamudio (approved):

leighdouglas (approved):
LGTM! 💃

PR #1001, opened 2020-09-28
Max Pledge Fix

jgsamudio:
# 📲 What

When the user selects a reward that equals the max pledge for the currency, the validation logic for the bonus support prevents the user from pledging

# 🤔 Why

The user should be able to pledge a reward that is the maximum for the currency. No bonus support should be added.

# 🛠 How

The currency max is what is used to determine the max. A bonus support max stream is added to update the max error string for the bonus support

# 📋 QA

- Go to the "Picton Magazine" project
- Pledge for the maximum reward (8,500)
- Confirm pledge and ensure the user cannot pledge with bonus support.


leighdouglas (approved):
LGTM!

PR #1006, opened 2020-10-05
NT-606: FIX Display a string indicating backing state for creators on View/Manage Pledge screen

Arkariang:
# 📲 What

- if a creator was also an admin, the query from the creators perspective was getting this error for fetching the field `        requiresAction`. With this changes the fields related to the checkout mutation won't be fetched from any query either backer or creator perspective.

# 🤔 Why
- Backing object null if trying to query a checkout related field.
Screen Shot 2020-10-05 at 8 57 12 AM

# 🛠 How
- Modified the backing fragment to not fetch fields
 `     clientSecret
        requiresAction
        status
`
- Added new fragment `checkoutBacking` for those same fields 
- modified `CreateBacking` and `UpdateBacking` mutations to use the new fragment

# 👀 See
wathc_pledge_as_creator
|  |  |

# 📋 QA

- Log in as creator @louisaadjadja can help with this step, she provided the credentials : 
Screen Shot 2020-10-05 at 2 23 38 PM
- Go to dashboard -> hit messages -> hit view pledge button. The view/manager pledge screen should have content.
- Pledge to something to check the CreateBacking mutations works as usual
- Update some pledge to check the UpdateBacking mutations keeps working as usual.

# Story 📖

[NT-606](https://kickstarter.atlassian.net/browse/NT-606)

### DIFF
--- a/app/src/main/graphql/checkout.graphql
+++ b/app/src/main/graphql/checkout.graphql
@@ -3,9 +3,7 @@ mutation CreateBacking($projectId: ID!, $amount: String!, $paymentType: String!,
     checkout {
       id
       backing {
-        clientSecret
-        requiresAction
-        status
+        ... checkoutBacking
       }
     }
   }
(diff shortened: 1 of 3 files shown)


leighdouglas (approved):
LGTM! 👍

PR #1013, opened 2020-10-14
NT-1587: Improve performance

Arkariang:
# 📲 What

- **BEFORE** : every time the amount was updated, the addOn item was re-builded to reflact that amount, by doing that the rx-chains where triggered again.
- **NOW**: will keep the reference to the selected addOns. We will update the quantity of the items just once the continue button has been pressed.

# 🤔 Why

- On projects with large amount of addOns there was performance issues when selecting the addOns try for example the this project with 25 addOns ->  https://www.kickstarter.com/projects/jake61341/santa-claus-mug-enamel-pin?ref=nav_search&result=project&term=santa%20claus%20pin

# 👀 See
- Select many addOns from a list of 25 addOns
many-addOns-performance
- memory usage screenshoot
memory-consumption-with-change

|  |  |

# 📋 QA
- ⚠️ in order to about the current timeout, if testing this PR, query for now shippingRules instead of shippingRulesExtended in case not digital, here what you need to change 
Screen Shot 2020-10-13 at 5 03 16 PM
- For a digital addOns you can try this one with 31 addOns -> https://www.kickstarter.com/projects/cmdrfalcon/sierra-collectors-quest

- Pledge to a project with many addOns 
- Update that pledge to add/remove addOns -> take a look into the button disabled if same selection
- Update pledge to a digital reward 

# Story 📖

[NT-1587](https://kickstarter.atlassian.net/browse/NT-1587)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/BackingAddOnViewHolderViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/BackingAddOnViewHolderViewModel.kt
@@ -249,8 +249,8 @@ class BackingAddOnViewHolderViewModel {
             this.quantity
                     .compose<Pair<Int, Reward>>(combineLatestPair(addOn))
                     .map { data -> Pair(data.first, data.second.id()) }
-                    .distinctUntilChanged { item1, item2 ->
-                        item1.first == item2.first && item1.second == item2.second
+                    .distinctUntilChanged { quantityPerId1, quantityPerId2 ->
+                        quantityPerId1.first == quantityPerId2.first && quantityPerId1.second == quantityPerId2.second
                     }
                     .compose(bindToLifecycle())
                     .subscribe(this.quantityPerId)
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/viewmodels/BackingAddOnViewHolderViewModel.kt
```diff
@@ -249,6 +249,9 @@ class BackingAddOnViewHolderViewModel {
             this.quantity
                     .compose<Pair<Int, Reward>>(combineLatestPair(addOn))
                     .map { data -> Pair(data.first, data.second.id()) }
+                    .distinctUntilChanged { item1, item2 ->
```
leighdouglas, line 252: Could we rename `item1` and `item2` to be what the actual items are? Just to make it more clear.

jgsamudio (approved):
LGTM!

leighdouglas (approved):

app/src/main/java/com/kickstarter/viewmodels/BackingAddOnViewHolderViewModel.kt
Arkariang: in this case item type is Pair(int, long), not sure what name give them for it to be clearer, what do you think it would be helpful here @leighdouglas ?

PR #1015, opened 2020-10-15
NT-1587: Query with filtering for shippingRules

Arkariang:
# 📲 What

- the backend has provided has a new query for filtering shippingRulesExtended by locationId

# 🛠 How

- Modified Rewards graphql fragment to not retrieve shippingRulesExpanded
- modified GetProjectAddOns query receive a new parameter for locationId
- in addOns nodes item use the new ShippingRulesExtended query with the locationId as parameter
- removed unnecessary fields from location graphql fragment
- modified KsApolliClient to parser the new structure for shippingRulesExtended
- When updating backed addOns modify the bakedAddOn item to hold the shippingRules from that same item from graphql
- in pledge fragment do not use the backing shippingRule if the selected reward has backed addOns

# 👀 See

| Before 🐛 | After 🦋 |
- **Before networking**  10 Second
networking_before

- **Now networking** less than 1 second
networking_now

- In the gif you can see how for Canada shippingCost is 5, when changing to United states is 0 
filter-location-small

|  |  |

# 📋 QA

- Use any of the big projects to back addOns 
- After pledging modify your selection changing the shipping rule

# Story 📖

[NT-1587](https://kickstarter.atlassian.net/browse/NT-1587)

### DIFF
--- a/app/src/main/graphql/fragments.graphql
+++ b/app/src/main/graphql/fragments.graphql
@@ -75,11 +75,6 @@ fragment reward on Reward {
         ... amount
     }
     shippingPreference
-    shippingRulesExpanded {
-        nodes {
-            ... shippingRule
-        }
-    }
     remainingQuantity
     limit
     limitPerBacker
@@ -91,7 +86,6 @@ fragment shippingRule on ShippingRule {
     cost {
         ... amount
     }
-    id
     location {
         ... location
     }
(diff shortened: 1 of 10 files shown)


Arkariang:
@leighdouglas here updating a query from Graph and updating some queries, take the chance to ask any question regarding GraphQL

jgsamudio (approved):
Looks great! 👍

PR #1019, opened 2020-10-16
NT-1615: fix Edit reward when add-ons were available but not backed

Arkariang:
# 📲 What

- Allow a editing an available reward with available addOns but not backed addOns

# 👀 See
select-same-reward-no-addOns

|  |  |

# 📋 QA

- Pledge to a reward with available addOns, but skip the addOns

- **Before** Edit reward -> The button for selecting the same reward not available
- **NOW** Edit reward -> The button for selecting the same reward is available letting the user modify the addOns selection

# Story 📖

[NT-1615](https://kickstarter.atlassian.net/browse/NT-1615)


leighdouglas (approved):
LGTM!

PR #1021, opened 2020-10-19
NT-1616: Shipping location retention when changing reward

Arkariang:
# 📲 What

- Change shipping location for a reward with addOns, but not backed addOns 

# 🤔 Why

- There was a few use cases not reflected regarding shipping location 

# 👀 See
- **Before** When modifying addOns selection the selected shippingRule was the default one, instead of the backed shippingRule.
- **NOW** After pledging with addOns with shippingLocation, if you edit the addOns the shipping location you'll see at the AddOns screen will be the backed one.
persist-backed-shipping-addOns

- **Before** When pledging to a reward with available addOns but skipping them, updating reward, select addOn, go to pledge screen, the shipping was the default one, instead of the chosen on AddOns screen.
- **NOW** When pledging to a reward with available addOns but skipping them, updating reward, select addOn, go to pledge screen, the shipping is the chosen one for addOns screen
correct-shipping-location-secondpledge-has-addOns

- **Before** editting reward, delete all addOns selection, on pledge screen not able to edit shipping selection
- **Now** editting reward, delete all addOns selection, on pledge screen shipping rule selector available to edit shipping
skip-addOns-edit-shipping-pledgeFragment

|  |  |

# 📋 QA

**Case 1**
- Pledge to reward with addOns with shipping different to your default
- Update pledge, make sure the shipping rule in addOns screen is the backed one.

**Case 2**
- Pledge to reward with addOns but skip them.
- Update pledge, select addOns now, go to pledge screen, make sure the shipping location is the correct one.

**Case 3**
- Pledge to a reward with addOns, select some addOns 
- Update pledge, delete all addOns selection and skip them, go to pledge screen
- Make sure on pledge screen you can select a shipping location and pledge with it.

# Story 📖

[NT-1616](Thttps://kickstarter.atlassian.net/browse/NT-1616)


app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -1504,6 +1526,11 @@ interface PledgeFragmentViewModel {
             return shippingCost
         }
 
+        /**
+         * When retrieving the backing shipping information we do have available the LocationId on the backingObject
+         * and the list of shippingRules available for the selected reward, so we the backing shippingRule will be
```
leighdouglas, line 1531: nit: change `so we the backing shippingRule will be` to `the backing shippingRule will be`

PR #1022, opened 2020-10-21
NT-1624: IndexOutOfBoundsException

Arkariang:
# 📲 What

- This concrete bug was fix for another editText (pledge_amount one in fragment_pledge_section_pledge_amount.xml ), it happens when having an EditText with `android:maxLength` field
Screen Shot 2020-10-21 at 1 39 13 PM

# 🛠 How

- New kotlin Extension created for the class android.widget.EditText 
- Refactor fix for `pledge_amount` and `bonus_amount` to use the same solution to avoid code duplication

# 👀 See
- Take a look here: [Crash report](https://console.firebase.google.com/u/1/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/8088312ce78ee3850631c3b034775de7?time=last-twenty-four-hours&versions=2.9.0)
Screen Shot 2020-10-21 at 1 44 10 PM

|  |  |

# 📋 QA

- Pledge to a reward no reward use the edit text to put an amount with as many numbers as you can, then hit the stepper up button. (This reward uses the pledge_amount editText)
- Pledge regular reward no reward use the edit text to put an amount with as many numbers as you can, then hit the stepper up button.(This reward uses the bonus_amount editText)

# Story 📖

[NT-1624](https://kickstarter.atlassian.net/browse/NT-1624)


leighdouglas (approved):
LGTM!

PR #1023, opened 2020-10-21
NT-1630: Fix UpdateActivity crash 

Arkariang:
# 📲 What

Screen Shot 2020-10-21 at 4 03 49 PM

# 🤔 Why
- We cannot reproduce the error, we are just trying to avoid it.

# 🛠 How
- Avoid any possible memory leak from the webView or having the webview runnig but dettached from the Activity cleaning up those two components when onDestroy.

- This PR goes along with another one in [here](https://github.com/kickstarter/native-secrets/pull/75), more detail inside.
- Once that PR is merged run `make secrets` in your local environment.
- Start the QA specified in that PR
|  |  |

# 📋 QA

# Story 📖


leighdouglas (approved):
LGTM!

PR #1024, opened 2020-10-22
NT-1625: Fix crash 

Arkariang:
# 📲 What
- In case using the GraphQL query for backing we receive a backing object, without backer information, feed empty instead of null to the dataModel.

# 🛠 How

- avoid null on backer name
- avoid shadowing it parameter

# 👀 See
[Crash report](https://console.firebase.google.com/u/1/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/310732673fa17a383580c28e897e2993?time=last-twenty-four-hours&versions=2.9.0%20(2010151641)&sessionEventKey=5F9076F502EC00017613A6D13BB67AFF_1464580771640330662)
Screen Shot 2020-10-22 at 11 05 12 AM
|  |  |

# 📋 QA

- we could not reproduce it

# Story 📖

[NT-1625](https://kickstarter.atlassian.net/browse/NT-1625)


leighdouglas (approved):
LGTM!

PR #1038, opened 2020-11-10
NT-1650: The user can navigate to their inbox from the app

leighdouglas:
# 📲 What

When the user taps "Open Inbox", they will either be taken to their default application, or shown a list of apps that they can choose from to open their email. If they have no apps available for email, it will show an "empty state" bottom sheet.

# 🤔 Why

We want users to be able to open their inbox application immediately upon receiving the verification email from kickstarter. 

# 🛠 How

When the user taps on the "Open Inbox" button, the viewmodel responds by sending at intent with parameters attached that indicate we would like to open an email application. This is built into the Android framework. 

# 👀 See

Nov-10-2020 11-24-53

# 📋 QA

- Tap on ovreflow menu
- Tap on "Internal Tools"
- Scroll to bottom, tap "Email Verification Interstitial"
- Tap "Open Inbox"
- Default email app or a list of email applications should appear. 

# Story 📖

[NT-1650: The user can navigate to their inbox from the app](https://kickstarter.atlassian.net/browse/NT-1650)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/EmailVerificationInterstitialFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/EmailVerificationInterstitialFragmentViewModel.kt
@@ -1,6 +1,5 @@
 package com.kickstarter.viewmodels
 
-import android.content.Intent
 import androidx.annotation.NonNull
 import com.kickstarter.libs.Environment
 import com.kickstarter.libs.FragmentViewModel
--- a/app/src/main/java/com/kickstarter/viewmodels/EmailVerificationInterstitialFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/EmailVerificationInterstitialFragmentViewModel.kt
@@ -1,23 +1,44 @@
 package com.kickstarter.viewmodels
 
+import android.content.Intent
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/viewmodels/EmailVerificationInterstitialFragmentViewModel.kt
```diff
@@ -4,9 +4,40 @@ import androidx.annotation.NonNull
...
+        val inputs = this
+        val outputs = this
+
+        private val openInboxButtonPressed = BehaviorSubject.create<Void>()
+
+        private val startEmailActivity = BehaviorSubject.create<Void>()
+
+
+        init {
+            openInboxButtonPressed
```
Arkariang, line 30: add in here `this.openInboxButtonPressed` we do explicitly call `this.` for the global variables

Arkariang (approved):
This looks great! 👍

app/src/main/java/com/kickstarter/viewmodels/EmailVerificationInterstitialFragmentViewModel.kt
leighdouglas: got it 👍

PR #1039, opened 2020-11-13
NT-1648: Update copy and fix interstitial screen copy 

leighdouglas:
# 📲 What

- Update "Open Inbox" copy on email verification interstitial screen to be sentence case instead of title case.
- Fix but caused by deprecated `getUser` method from User object class. 

# 🤔 Why

Upon added a required `user` parameter to the interstitial fragment for the login and signup activity, opening the interstitial from the internal tools screen was crashing because the user object was being called by a deprecated method. The user object was null and it caused a `nullpointerexpection`.

# 🛠 How

In response to clicking the Interstitial screen button in the internal tools page, the activity subscribes to an observable provided by the `user` model that emits the current user. If this stream emits, the fragment is started. This means, if you are not logged in, you will not be able to see the interstitial screen from the internal tools page. 

# 👀 See

This is an example of the copy change I made: 
Before and After
Screenshot_1604659516 Screenshot_1605294349

# 📋 QA

Crash QA:
- Tap the overflow menu at the top left of the screen
- Select "Internal Tools"
- Scroll to the bottom and select "Email Verification Interstitial" 
        - Not logged in: There should be no reaction.
        - Logged in: The interstitial screen should appear

Copy QA:
- Make sure you are logged in and repeat the steps above
- Observe that the interstitial cta button says "Open inbox", in sentence case. 

# Story 📖

[Build interstitial UI](https://kickstarter.atlassian.net/browse/NT-1648)


app/src/internal/java/com/kickstarter/ui/activities/InternalToolsActivity.java
```diff
@@ -152,7 +153,14 @@ public void featureFlagsClick() {
...
   public void emailVerificationInterstitialClick() {
-    final User user = this.environment().currentUser().getUser();
+    this.environment()
+            .currentUser()
+            .observable()
+            .filter(ObjectUtils::isNotNull)
+            .subscribe(this::startInterstitialFragment);
+  }
+
+  private void startInterstitialFragment(@NonNull User user) {
```
jgsamudio, line 163: There is a checkstyle error that the user should be final

jgsamudio (approved):
Looks good after fixing the failed test error

colleenmacd:
thank you!!

colleenmacd (approved):

app/src/internal/java/com/kickstarter/ui/activities/InternalToolsActivity.java
```diff
@@ -152,7 +153,14 @@ public void featureFlagsClick() {
...
   public void emailVerificationInterstitialClick() {
-    final User user = this.environment().currentUser().getUser();
+    this.environment()
+            .currentUser()
+            .observable()
+            .filter(ObjectUtils::isNotNull)
+            .subscribe(this::startInterstitialFragment);
+  }
+
+  private void startInterstitialFragment(@NonNull final User user) {
```
Arkariang, line 163: nit: you can use the method on the LoginHelper https://github.com/kickstarter/android-oss/blob/8d01e7811e4bdb18b0f8780c0fe4fc6886073ea5/app/src/main/java/com/kickstarter/libs/utils/LoginHelper.kt#L50

that method was created to not have to re-write the logic every time we need to present the interstitial, here an example of use in Java https://github.com/kickstarter/android-oss/blob/8d01e7811e4bdb18b0f8780c0fe4fc6886073ea5/app/src/main/java/com/kickstarter/ui/activities/SignupActivity.java#L82

and here example of use in kotlin https://github.com/kickstarter/android-oss/blob/8d01e7811e4bdb18b0f8780c0fe4fc6886073ea5/app/src/main/java/com/kickstarter/ui/activities/LoginActivity.kt#L104

No need to change it, the goal of having the interstitial presented on the internal tools was to be able to work both of us in parallel, and made UAT for desing easier , this comment is just an FYI :)

Arkariang (approved):

app/src/main/res/values/strings_i18n.xml
```diff
@@ -388,7 +388,7 @@ from friends yet.</string>
   <string name="Opens_email_composer" formatted="false">Opens email composer.</string>
   <string name="Opens_filters" formatted="false">Opens filters.</string>
   <string name="Opens_help_sheet" formatted="false">Opens help sheet.</string>
-  <string name="open_inbox" formatted="false">Open Inbox</string>
```
Arkariang, line 391: I've just realised about this :sweat_smile:, but, is the string "Open inbox" translated? Or we haven't done the translations yet? @leighdouglas

PR #1043, opened 2020-11-26
NT-1673: Deep Linking from Email Verification

Arkariang:
# 📲 What

- Open App in Discovery activity when the user press the email verification link.

# 🛠 How

- Provided OkHttpClient with correct headers to the environment object
- Declare on Manifest.xml the intent-data for opening the app 
- Make network request to resolve the verification
- Simplified EmailVerificationFragment as the envelop is no longer needed for resending the email

# 👀 See
**Expired Token**
Expired-Token

**Unauthorized (when user is not logged in)**
UnAuthorized-token

|  |  |

# 📋 QA
- Expired Token will show a snackbar in Discovery with the 403 code
- Success validation will show a snackbar in Discovery with the 200 code
- Unauthorized will show a snackbar in Discovery with the 401 code

# Story 📖

[Deep Linking](https://kickstarter.atlassian.net/browse/NT-1673)


app/src/main/AndroidManifest.xml
```diff
@@ -1,336 +1,302 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest package="com.kickstarter"
-  xmlns:android="http://schemas.android.com/apk/res/android"
-  xmlns:tools="http://schemas.android.com/tools">
-
-  <!-- PERMISSIONS -->
-
-  <uses-permission android:name="android.permission.INTERNET" />
-  <!--
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
jgsamudio, line 2: Was this file just formatted?

app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.java
```diff
@@ -420,6 +464,18 @@ private int currentDrawerMenuIcon(final @Nullable User user) {
       }
     }
 
+    private Observable<Response> makeCall(final @NonNull Uri uri) {
```
jgsamudio, line 467: What do you think about renaming this to something else? `apiRequest` or `apiRequestForUri`?

app/src/main/java/com/kickstarter/viewmodels/LoginViewModel.kt
```diff
@@ -272,6 +273,6 @@ interface LoginViewModel {
 
         override fun tfaChallenge() = this.tfaChallenge
 
-        override fun showInterstitialFragment(): BehaviorSubject<AccessTokenEnvelope> = this.showInterstitial
```
jgsamudio, line 275: Do we need the access token on the interstitial screen for resending the email verification email?

jgsamudio (approved):

app/src/main/java/com/kickstarter/viewmodels/LoginViewModel.kt
Arkariang: It has been removed, we do not need the accessToken on the interstitial. The resend email mutation works with the backend session, so we don't need any data to pass any data into the interstitial, that point was unclear when that screen was builded so doing a bit of refactor in here :)

app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.java
Arkariang: it's not a request to an REST API endpoint, you could call any uri, that's why the name is so generic

app/src/main/AndroidManifest.xml
Arkariang: - added the deepling intent data on DiscoveryActviity tags
- and <activity ...></activity> has been changed to <activity .... /> for those activities without intent data

PR #1047, opened 2020-12-01
NT-1680: Send Email when interstitial presented

Arkariang:
# 📲 What

- When the interstitial is presented we send the validation email

# 👀 See
getEmailWhenPresented

|  |  |

# 📋 QA

- Every time you open the Interstitial with an user logged in you should get an email.

# Story 📖

[NT-1680](https://kickstarter.atlassian.net/browse/NT-1680)


jgsamudio (approved):
Great work!

PR #1048, opened 2020-12-01
NT-1467: Kotlin 1.4.20

Arkariang:
# 📲 What

- Updated kotlin version 
- Do not instantiate firebase dependencies on tests 

# 👀 See
scripts
|  |  |

# 📋 QA

**You need in your environment a Gradle version 6.+**
**You need in your environment the JDK version  1.8.+**
Execute on you environment the next scripts: 
- ` bundle exec fastlane external` -> this creates and publish to firebase an external release
- ` bundle exec fastlane internal` -> this creates and publish to firebase an internal release
- `./bin/milkrun update-strings ` -> -> this fetch the new translations and updates all the languages string files

# Story 📖

[NT-1467](https://kickstarter.atlassian.net/browse/NT-1467)

### DIFF
--- a/app/src/main/java/com/kickstarter/KSApplication.java
+++ b/app/src/main/java/com/kickstarter/KSApplication.java
@@ -46,14 +46,14 @@ public class KSApplication extends MultiDexApplication {
     }
 
     JodaTimeAndroid.init(this);
-    FirebaseCrashlytics.getInstance().setCrashlyticsCollectionEnabled(true);
 
     this.component = DaggerApplicationComponent.builder()
       .applicationModule(new ApplicationModule(this))
       .build();
     component().inject(this);
 
-    if (FirebaseApp.getApps(getApplicationContext()).isEmpty() && !!isInUnitTests()) {
+    if (FirebaseApp.getApps(getApplicationContext()).isEmpty() && !isInUnitTests()) {
+      FirebaseCrashlytics.getInstance().setCrashlyticsCollectionEnabled(true);
       FirebaseApp.initializeApp(getApplicationContext());
       FirebaseAnalytics.getInstance(getApplicationContext()).setAnalyticsCollectionEnabled(true);
     }
--- a/app/src/main/java/com/kickstarter/KSApplication.java
+++ b/app/src/main/java/com/kickstarter/KSApplication.java
@@ -53,8 +53,10 @@ public class KSApplication extends MultiDexApplication {
       .build();
     component().inject(this);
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/KSApplication.java
```diff
@@ -46,15 +46,17 @@ public void onCreate() {
...
-    FirebaseCrashlytics.getInstance().setCrashlyticsCollectionEnabled(true);
 
     this.component = DaggerApplicationComponent.builder()
       .applicationModule(new ApplicationModule(this))
       .build();
     component().inject(this);
 
-    FirebaseApp.initializeApp(this);
-    FirebaseAnalytics.getInstance(this).setAnalyticsCollectionEnabled(true);
+    if (FirebaseApp.getApps(getApplicationContext()).isEmpty() && !isInUnitTests()) {
```
jgsamudio, line 55: Question: Why do we need to check the app context?

app/src/main/java/com/kickstarter/ui/adapters/KSListAdapter.kt
```diff
@@ -12,6 +12,7 @@ import com.kickstarter.BuildConfig
 import com.kickstarter.libs.utils.ExceptionUtils
 import com.kickstarter.ui.viewholders.KSViewHolder
 import com.trello.rxlifecycle.ActivityEvent
+import java.util.ArrayList
```
jgsamudio, line 15: Do we need this import?

Arkariang: yes, It's explicitly using the ArrayList from java, otherwise the compiler thinks it's the kotlin one, and that will force more code changes within this class or a full migration to kotlin.

app/src/main/java/com/kickstarter/KSApplication.java
Arkariang: - FirebaseApp.getApps(getApplicationContext()).isEmpty() check is for initializing just once. 
- `getApplicationContext()` could be substituted by `this` i'm just accessing it with the contextWrapper method -> https://developer.android.com/reference/kotlin/android/content/ContextWrapper#getapplicationcontext

jgsamudio: Ah I see. Was this a change with the new version of Kotlin? Why was this changed?

jgsamudio (approved):

PR #1057, opened 2020-12-07
NT-715: Update OkHTTP & Retrofit libraries

Arkariang:
# 📲 What

- Update Networking libraries

# 👀 See
- Log in in the app (V1) 
- load projects (V1)
logIn+explore

- select a project and back it with addOns (GraphQL)
- Check older backed project (GraphQL)
back-project

|  |  |

# 📋 QA
- Perform some of the basic features within the app

** Test V1 (OkHTTP + Retrofit) **
- Log in in the APP
- Lad projects
- Select and project

** Test GraphQL (OkHTTP) **
- Back a project with addons if you find one 
- Look at one previously backed project 

# Story 📖

[NT-715](https://kickstarter.atlassian.net/browse/NT-715)

### DIFF
--- a/app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
+++ b/app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
@@ -5,20 +5,25 @@ import com.kickstarter.libs.Build
 import com.kickstarter.libs.CurrentUserType
 import com.kickstarter.libs.utils.WebUtils
 import okhttp3.Interceptor
+import okhttp3.Interceptor.Chain
 import okhttp3.Response
 
 class GraphQLInterceptor(private val clientId: String,
                          private val currentUser: CurrentUserType,
                          private val build: Build) : Interceptor {
-    override fun intercept(chain: Interceptor.Chain?): Response {
-        val original = chain!!.request()
-        val builder = original.newBuilder().method(original.method(), original.body())
-        if (this.currentUser.exists()) {
-            builder.addHeader("Authorization", "token " + this.currentUser.accessToken)
-        }
+    override fun intercept(chain: Chain): Response {
+        val original = chain.request()
+        val builder = original.newBuilder().method(original.method, original.body)
+
+        this.currentUser.observable()
+                .subscribe {
+                    builder.addHeader("Authorization", "token " + this.currentUser.accessToken)
+                }
+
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
```diff
@@ -5,20 +5,25 @@ import com.kickstarter.libs.Build
...
-        if (this.currentUser.exists()) {
-            builder.addHeader("Authorization", "token " + this.currentUser.accessToken)
-        }
+    override fun intercept(chain: Chain): Response {
+        val original = chain.request()
+        val builder = original.newBuilder().method(original.method, original.body)
+
+        this.currentUser.observable()
+                .subscribe {
+                    builder.addHeader("Authorization", "token " + this.currentUser.accessToken)
```
leighdouglas, line 20: What is exactly happening here? I understand that when this `currentUser.observable()` emits, we're adding a header to the chain builder, but what if this emits after the this method is executed and returns the chain, what happens?

app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
```diff
@@ -5,20 +5,25 @@ import com.kickstarter.libs.Build
 import com.kickstarter.libs.CurrentUserType
 import com.kickstarter.libs.utils.WebUtils
 import okhttp3.Interceptor
+import okhttp3.Interceptor.Chain
 import okhttp3.Response
 
 class GraphQLInterceptor(private val clientId: String,
```
leighdouglas, line 11: Nit: Is it possible we could add documentation here since we're modifying this class already? Would help provide context as to what the purpose of this class is. Just an idea!

leighdouglas (approved):
Looks good to me!

app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
Arkariang: this.currentUser.exists() is deprecated, so changing it over .observable

app/src/main/java/com/kickstarter/services/interceptors/GraphQLInterceptor.kt
Arkariang: Done!

PR #1059, opened 2020-12-09
NT-1707: Update Activity crash

Arkariang:
# 📲 What

- Minimize the top most crash we are currently suffering 
- Migrate the UpdateActivity.java to kotlin, and protect any method susceptible of being called outside of the lifecycle context.
- Delete any code related with the` butterknife` dependency for this Activity.

# 🤔 Why

ActualTop first occurring crash, not able to reproduce locally, but [here](https://console.firebase.google.com/u/1/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/cb66d708261c387b165474ba82cba1fd?time=last-seven-days) you can see with more detail the stacktrace. 

The stacktrace suggest the crash happens when the method
`
.handleProjectUpdateUriRequest`

is called before the activity lifecycle `onCreate`  method or after `onDestroy` method, the only moments in which the ViewModel.inputs can be null, so we are tacking the change to migrate this class from Java to kotlin and protect against NullPointerExceptions using Kotlin optionals. Also deleting all references in this activity to the ` butterknife` dependency.

# 👀 See
migration_update_activity

|  |  |

# 📋 QA
- Open the Update activity screen, load some updates for projects from that activity, interact with some of the updates (like, comment, share ...).
- Open the Update activity screen, open the project page from some project

# Story 📖

[NT-1707](https://kickstarter.atlassian.net/browse/NT-1707)


leighdouglas (approved):
Great job!! 🎉

PR #1064, opened 2021-01-05
NT-1700 Create StringExt.kt and Test Suite

leighdouglas:
# 📲 What

Created the `StringExt` file as well as the test suite. These methods will replace the ones used from `StringUtils.java` A subsequent pull request will address updating the these methods throughout the codebase. 

# 🤔 Why

An ongoing effort to migrate our codebase to Kotlin and address technical debt.

# 📋 QA

N/A, these additions in the PR are isolated and not reflected in the app in the UI or functionality. 

# Story 📖

[Migrate StringUtils.java to kotlin extension](https://kickstarter.atlassian.net/browse/NT-1700)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
@@ -1,7 +1,8 @@
+@file:JvmName("StringExt")
 package com.kickstarter.libs.utils.extensions
 
 import android.util.Patterns
-import java.util.*
+import java.util.Locale
 
     const val MINIMUM_PASSWORD_LENGTH = 6
 
@@ -16,16 +17,14 @@ import java.util.*
     * Returns a boolean that reflects if the string is empty or the length is zero when space
     * characters are trimmed
     */
-    fun String?.isEmpty(): Boolean {
+    fun String.isEmpty(): Boolean {
         return this == null || this.trim().length == 0
     }
 
     /**
     * Returns a boolean of if the string is not empty
     */
-    fun String?.isPresent(): Boolean {
-        return !this.isEmpty()
-    }
+    fun String.isPresent() = !this.isEmpty()
 
     /**
     * Returns a boolean of if the string is not empty and has more than 5 characters
@@ -38,19 +37,18 @@ import java.util.*
      * Returns a string with only the first character capitalized.
      */
     fun String.sentenceCase(): String {
-        return if (this.length <= 1) this.toUpperCase(Locale.getDefault()) else this.substring(0, 1).toUpperCase(Locale.getDefault()) + this.substring(1).toLowerCase(Locale.getDefault())
+        return if (this.length <= 1) this.toUpperCase(Locale.getDefault())
+        else this.substring(0, 1).toUpperCase(Locale.getDefault()) + this.substring(1).toLowerCase(Locale.getDefault())
     }
 
     /**
      * Returns a string with no leading or trailing whitespace.
      */
     fun String.trim(): String {
-        return this.replace('\u00A0', ' ').trim { it <= ' ' }.replace(" +".toRegex(), " ")
+        return this.trim()
     }
 
     /**
      * Returns a string wrapped in parentheses.
      */
-    fun String.wrapInParentheses(): String {
-        return "($this)"
-    }
+    fun String.wrapInParentheses() = "($this)"
--- a/app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
@@ -14,23 +14,23 @@ import java.util.Locale
     }
 
     /**
-    * Returns a boolean that reflects if the string is empty or the length is zero when space
+    * Returns a boolean that reflects if the string is empty or the length is zero when white space
     * characters are trimmed
     */
-    fun String.isEmpty(): Boolean {
-        return this == null || this.trim().length == 0
+    fun String.isTrimmedEmpty(): Boolean {
+        return this.isNullOrEmpty() || this.trimAllWhitespace().length == 0
     }
 
     /**
     * Returns a boolean of if the string is not empty
     */
-    fun String.isPresent() = !this.isEmpty()
+    fun String.isPresent() = !this.isTrimmedEmpty()
 
     /**
     * Returns a boolean of if the string is not empty and has more than 5 characters
     */
     fun String.isValidPassword(): Boolean {
-        return !this.isEmpty() && this.length >= MINIMUM_PASSWORD_LENGTH
+        return !this.isTrimmedEmpty() && this.length >= MINIMUM_PASSWORD_LENGTH
     }
 
     /**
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun String.isEmail(): Boolean {
+        return Patterns.EMAIL_ADDRESS.matcher(this).matches()
+    }
+
+    /**
+    * Returns a boolean that reflects if the string is empty or the length is zero when space
+    * characters are trimmed
+    */
+    fun String?.isEmpty(): Boolean {
+        return this == null || this.trim().length == 0
```
Arkariang, line 20: this implementation already exists in kotlin it's called isNullOrEmpty -> https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-null-or-empty.html

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+    * Returns a boolean of if the string is not empty
+    */
+    fun String?.isPresent(): Boolean {
```
Arkariang, line 26: (suggestion here) could this by an inline function?, seems is gonna be less verbose that way

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -0,0 +1,56 @@
...
+    */
+    fun String.isValidPassword(): Boolean {
+        return !this.isEmpty() && this.length >= MINIMUM_PASSWORD_LENGTH
+    }
+
+    /**
+     * Returns a string with only the first character capitalized.
+     */
+    fun String.sentenceCase(): String {
+        return if (this.length <= 1) this.toUpperCase(Locale.getDefault()) else this.substring(0, 1).toUpperCase(Locale.getDefault()) + this.substring(1).toLowerCase(Locale.getDefault())
```
Arkariang, line 41: the else branch on the next line

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -0,0 +1,56 @@
...
+     * Returns a string with no leading or trailing whitespace.
+     */
+    fun String.trim(): String {
+        return this.replace('\u00A0', ' ').trim { it <= ' ' }.replace(" +".toRegex(), " ")
+    }
+
+    /**
+     * Returns a string wrapped in parentheses.
+     */
+    fun String.wrapInParentheses(): String {
```
Arkariang, line 54: could this be an inline function as well?

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -0,0 +1,56 @@
+package com.kickstarter.libs.utils.extensions
+
+import android.util.Patterns
+import java.util.*
```
Arkariang, line 4: why do you need here an import from Java?

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -0,0 +1,56 @@
+package com.kickstarter.libs.utils.extensions
```
Arkariang, line 1: You need to add on top of the package the annotation for being able to use this extension from the Java source code 
`@file:JvmName("StringExt")`
take a look into UriExt.kt or ConfigExtension.kt

- Also a heads up, adding that annotation will force you to not have optional methods in an extension class as the ones you have here String?.isEmpty or String?.isPresent

Arkariang (changes requested):

app/src/test/java/com/kickstarter/libs/utils/extensions/StringExtKtTest.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Test
+    fun isEmail_whenNotGivenEmail_shouldReturnFalse() {
+        assertFalse(INVALID_EMAIL.isEmail())
+    }
+
+    @Test
+    fun isEmpty_whenGivenEmptyString_shouldReturnTrue() {
+        assertTrue(EMPTY_STRING.isEmpty())
```
Arkariang, line 20: No need for this test, isEmpty is an String function both in Java and kotlin  java -> https://www.javatpoint.com/java-string-isempty, Kotlin -> https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-empty.html

Same with all tests for the functionallities already existing in the languages

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -0,0 +1,56 @@
...
+     * Returns a string with only the first character capitalized.
+     */
+    fun String.sentenceCase(): String {
+        return if (this.length <= 1) this.toUpperCase(Locale.getDefault()) else this.substring(0, 1).toUpperCase(Locale.getDefault()) + this.substring(1).toLowerCase(Locale.getDefault())
+    }
+
+    /**
+     * Returns a string with no leading or trailing whitespace.
+     */
+    fun String.trim(): String {
```
Arkariang, line 47: Trim does also existis in kotlin ->  https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/trim.html
And in java -> https://www.javatpoint.com/java-string-trim

Arkariang (changes requested):

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
leighdouglas: So question about this - the previous functionality is testing if a string is null. I don't want to change the functionality for fear of causing a crash so I'd like to leave the null checks there. Will removing these optionals cause any issues?

app/src/test/java/com/kickstarter/libs/utils/extensions/StringExtKtTest.kt
leighdouglas: This is testing the isEmpty extension function in the StringsExt file, not the isEmpty java/kotlin function.

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
Arkariang: the you'll have to check only on the java code before using that extension function if its null, and in the kotlin code is it's an optional string use the operator '?'

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
leighdouglas: So `isEmpty` and `trim` were methods that existed in the `StringUtils.java` class, so I just migrated all of the old functions to the ext. Should i remove them from the extension and update their usages throughout the app to use the kotlin/java methods? I was trying to avoid this as not to touch the logic too much throughout the application.

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
leighdouglas: We're using it for the `Locale` java file. Thanks for catching this, I am going to update this to import this specific class

leighdouglas:
@Arkariang So I ended up leaving the `trim()` and `isEmpty()` methods, but I updated their function names since they are doing additional functionality and not just reflecting what the respective java and kotlin String functions are doing. Trim was written this way in the `StringUtils.java` class because the java `trim()` method does not consider non-breaking spaces to be a whitespace, and we had to account for that. Now that kotlin's `trim()` does that, we don't need to convert the non-breaking characters to a blank space to remove them, and we can use this extension method in java classes to call the kotlin `trim()`.

Arkariang (approved):

PR #1067, opened 2021-01-07
Introduce a simple Screenshot Test, and make sure it runs on CircleCI

tdrhq:
# 📲 What

Hi @Arkariang!

As per my promise from yesterday. In this PR, I intentionally haven't done the full integration with Screenshotbot, I only wrote your first screenshot test, and added code to run your instrumentation tests on CircleCI. I wanted to make sure we can run these tests properly, then I'll come back and do the final step of integrating with Screenshotbot (which would be the easy part).

So quick summary of what's going on here, because there's a bit. It looks like you already had an old instrumentation test, but as far as I could tell it wasn't been actively run for a while. So I brought it up to speed, made sure I could run it properly locally. After that I added a basic screenshot tests. (I picked a layout where it seemed like you modified it lately)

Now, I wanted to make sure you can run this test on CircleCI. So I used my own CircleCI instance to modify `.circleci/config.yml`. I've set it up to run the tests with Firebase Test Lab. This is where I'm really going to need your help: you do have to do some minor set up on your end (or may be it's already set up? In which case the builds on this PR should just pass). Basically we'll need to create a service key with your gcloud account and set up some CircleCI environment variables. If you have access to the Google Cloud account, I'm happy to do something like AnyDesk and do that work for you.

At this point we don't run the `verify` step of screenshot tests (this is because the default screenshot test plugin doesn't work with Firebase Test Lab out of the box, but I'll be able to do that once I bring in Screenshotbot). I did record the screenshot though, which is attached to this PR!

Your build time might go up a bit, but there's room for optimization. Currently we're building both :app:externalRelease and :app:regressionDebug. In theory, you should be able to get away with building only one of them, but that's for another day.

Link to my passing CircleCI build: https://app.circleci.com/pipelines/github/tdrhq/android-oss/23/workflows/c8a8ca7c-bf42-49a3-bcc4-5994df0f7975 (I think it should be public)

cc @justinswart, @singhhari

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖


Arkariang:
This looks amazing! 😻   We do not have yet I'm afraid the CI environment ready, let me speak with the infrastructure team and I'll be back to you as soon as I can! 

Thank you so much @tdrhq !

Arkariang:
We will retake this conversation once we have some Screenshoot tests in place, thanks for your help!

tdrhq:
@Arkariang No worries, let me know when you're ready to start using Screenshot tests, I'd love to help out at that time.

PR #1069, opened 2021-01-13
NT-1580 : Forgot password email

hadia:
# 📲 What

fill email field in reset password screen with the email from the login screen  

# 🤔 Why

If the user enters an email address on the login screen and they navigate to the forgot password screen

# 🛠 How

- Pass email from Login Activity through Intent to Reset Password Activity
 

# 👀 See
device-2021-01-13-184426

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖

NT-1580/Android Forgot Password Email


Arkariang (approved):
Fantastic work here! 🎆

PR #1077, opened 2021-01-26
NT-1739: Remove all instances of koala tracking events

leighdouglas:
# 📲 What

- Remove Koala from environment object
- Remove Koala tracking events from view models and activities
- Remove Koala instances from test classes 

# 🤔 Why

Migrating to Segment, we no longer need these events tracked by Koala. 

# 🛠 How

The environment object contains data used across view models and activities. Its dependencies are created in the ApplicationModule. Removing this dependency from the environment object, we can no longer use koala in our view model and activity classes where it is being used. In subsequent PRs, we will remove Koala from the ApplicationModule completely. 

# 📋 QA

No visible changes to the application. 

# Story 📖

[NT-1739: Android Segment: Remove Koala Tracking Client](https://kickstarter.atlassian.net/browse/NT-1739)


Arkariang (approved):
lgtm! great work @leighdouglas 🎊

PR #1078, opened 2021-01-28
[NT-1745]&[NT-1740]: Refactor tracking clients & bring segment dependency

Arkariang:
# 📲 What

Several things on this PR: 
- Delete `KoalaTrackingClient`, `KoalaWorker`, `KoalaService`, and `KoalaRetrofit`
- Migrated to kotlin: `TrackingClient`, `TrackingClientType`
- Renamed and migrated `Koala.java` to `AnalyticsEvents.kt`
- Added Segment dependency and created `SegmentTrackingClient`
- AnalyticsEvents instance now holds a collection of trackingClients: LakeTrackingClient and SegmentTrackingClient
- Prepared test environment for future Segment specific use cases

# 🤔 Why

- Now the ViewModels speaks to the AnalyticsEvents instance, in that way we abstracted away the specification for which client should that event be send to.

# 👀 See
* On logcat console you can see the events for both ends : Lake and Segment, without code updates on any ViewModel
Screen Shot 2021-01-27 at 4 08 57 PM

* You can also check on the Segment Dashboard, on the Staging source the received events: 
Screen Shot 2021-01-27 at 4 11 38 PM

|  |  |

# 📋 QA

- Checkout this branch and load staging, make sure you can see on the logcat the events, and also in the segment dashboard(staging source).

# Story 📖

[NT-1745: Refactor Tracking Clients](https://kickstarter.atlassian.net/browse/NT-1745)
[NT-1740: Bring Segment Dependency](https://kickstarter.atlassian.net/browse/NT-1740)


hadia (approved):
great job👏

app/src/main/java/com/kickstarter/ApplicationModule.java
```diff
@@ -133,16 +134,15 @@ static Environment provideEnvironment(final @NonNull @ActivitySamplePreference I
     final @NonNull InternalToolsType internalToolsType,
     final @NonNull KSCurrency ksCurrency,
     final @NonNull KSString ksString,
-    final @NonNull @LakeTracker Koala lake,
+    final @NonNull @LakeTracker AnalyticEvents analytics,
```
leighdouglas, line 137: What is the `@LakeTracker` annotation doing here? Do we need to keep this if the AnalyticsEvents class is not supposed to be specific to one event tracker?

app/src/main/java/com/kickstarter/ApplicationModule.java
```diff
@@ -172,6 +172,32 @@ static Environment provideEnvironment(final @NonNull @ActivitySamplePreference I
...
+    Analytics segmentClient = null;
+
+    if (build.isRelease() && Build.isExternal()) {
+      apiKey = Secrets.Segment.PRODUCTION;
+    }
+    if (build.isDebug()) {
+      apiKey = Secrets.Segment.STAGING;
+    }
+
+    if (context instanceof KSApplication && !((KSApplication) context).isInUnitTests()) {
```
leighdouglas, line 189: Since we're already checking if context is an instance of KSApplication, do we need to cast? Or would that be a compiler error?

app/src/main/java/com/kickstarter/ApplicationModule.java
```diff
@@ -406,20 +415,20 @@ Application provideApplication() {
...
-    final @NonNull Build build, final @NonNull CurrentConfigType currentConfig, final @NonNull ExperimentsClientType experimentsClientType) {
-    return new Koala(new LakeTrackingClient(context, currentUser, build, currentConfig, experimentsClientType));
+  static AnalyticEvents provideAnalytics(
+          final @ApplicationContext @NonNull Context context,
+          final @NonNull CurrentUserType currentUser,
+          final @NonNull Build build,
+          final @NonNull CurrentConfigType currentConfig,
+          final @NonNull ExperimentsClientType experimentsClientType,
+          final @Nullable Analytics segment) {
+    final LakeTrackingClient lakeTrackingClient = new LakeTrackingClient(context, currentUser, build, currentConfig, experimentsClientType);
```
leighdouglas, line 428: Is there a reason that you are creating instances of these tracking clients inside of this provider method? I thought it would make more sense to have these dependencies handled by their own dagger provider methods in this ApplicationModule.

app/src/main/java/com/kickstarter/libs/FragmentViewModel.java
```diff
@@ -24,11 +24,11 @@
 
   private final PublishSubject<ActivityResult> activityResult = PublishSubject.create();
   private final PublishSubject<Bundle> arguments = PublishSubject.create();
-  protected final Koala lake;
+  protected final AnalyticEvents lake;
```
leighdouglas, line 27: nit: Maybe we should change the name of this variable from `lake` to `analyticsEvents`.

app/src/main/java/com/kickstarter/libs/SegmentTrackingClient.kt
```diff
@@ -0,0 +1,38 @@
+package com.kickstarter.libs
+
+import android.content.Context
+import com.segment.analytics.Analytics
+import com.segment.analytics.Properties
+
+class SegmentTrackingClient(
```
leighdouglas, line 7: nit: Could we alphabetize this list of dependencies?

app/src/main/java/com/kickstarter/libs/TrackingClientType.kt
```diff
@@ -0,0 +1,126 @@
...
+        LAKE("💧 Lake"), SEGMENT("\uD83C\uDF81 Segment");
+    }
+
+    protected abstract val isGooglePlayServicesAvailable: Boolean
+    protected abstract val isTalkBackOn: Boolean
+
+    abstract fun optimizely(): ExperimentsClientType?
+    abstract fun loggedInUser(): User?
+
+    protected abstract fun type(): Type
```
leighdouglas, line 20: nit: could we alphabetize type? Seems like the rest of this list is in alphabetical order.

leighdouglas (approved):
🎉 This is great!!!

app/src/main/java/com/kickstarter/ApplicationModule.java
Arkariang: Compilation error, it needs the cast

app/src/main/java/com/kickstarter/libs/FragmentViewModel.java
Arkariang: I'll rename it on a separate PR just with that change, it will affect a bunch of VM's

app/src/main/java/com/kickstarter/libs/TrackingClientType.kt
Arkariang: Done

app/src/main/java/com/kickstarter/libs/SegmentTrackingClient.kt
Arkariang: Done

PR #1084, opened 2021-02-05
NT-1751: Reset method called when user logs out

Arkariang:
# 📲 What

- `Reset` method from segment SDK needs to be called when a logged in user logs out 

# 🛠 How
- Added new method `reset` on `AnalyticsEvents` and the rest of the hierarchy

# 📋 QA
- Logout from the app
- Take a look into logcat, you'll see the debug logs for the reset method being called
Screen Shot 2021-02-04 at 4 26 29 PM

# Story 📖

[NT-1751: Call reset on logout](https://kickstarter.atlassian.net/browse/NT-1751)

### DIFF
--- a/app/src/test/java/com/kickstarter/viewmodels/ChangePasswordViewModelTest.kt
+++ b/app/src/test/java/com/kickstarter/viewmodels/ChangePasswordViewModelTest.kt
@@ -3,8 +3,12 @@ package com.kickstarter.viewmodels
 import UpdateUserPasswordMutation
 import com.kickstarter.KSRobolectricTestCase
 import com.kickstarter.R
-import com.kickstarter.libs.Environment
+import com.kickstarter.libs.*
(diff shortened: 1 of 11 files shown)


app/src/test/java/com/kickstarter/viewmodels/ChangePasswordViewModelTest.kt
```diff
@@ -3,8 +3,12 @@ package com.kickstarter.viewmodels
 import UpdateUserPasswordMutation
 import com.kickstarter.KSRobolectricTestCase
 import com.kickstarter.R
-import com.kickstarter.libs.Environment
+import com.kickstarter.libs.*
```
leighdouglas, line 6: nit: could we remove this wild card import statement?

leighdouglas (approved):

PR #1094, opened 2021-02-10
EP-216: Create event for Pledge Initiate CTA

leighdouglas:
# 📲 What

Added track event for "pledge initiate" button clicks that fires when the "Back this project" button is pressed on the projects screen.

# 🤔 Why

Segment integration.

# 🛠 How

- Created a new function to our AnalyticsEvents class that handles sending the event name and properties to the client.
- Added this function to the `ProjectViewModel` when the "Back this project" button is pressed.

# 👀 See
This is the button that fires this event. 
Screenshot_1612992209

# 📋 QA

- Tap on a live project
Both Segment (CTA Clicked) and DataLake (Project Page Pledge Button Clicked) should appear in the segment dashboard:
Screen Shot 2021-02-10 at 4 22 45 PM

- Tap on CTA Clicked, the list of properties should appear on the right side of the screen:
Screen Shot 2021-02-10 at 4 22 39 PM

# Story 📖

[EP-216: Android: CTA Clicked (pledge_initiate)](https://kickstarter.atlassian.net/browse/EP-216)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -2,7 +2,8 @@ package com.kickstarter.libs
 
 import com.kickstarter.libs.KoalaContext.*
 import com.kickstarter.libs.KoalaEvent.ProjectAction
-import com.kickstarter.libs.utils.EventContext
+import com.kickstarter.libs.utils.EventContext.CtaContextName.ADD_ONS_CONTINUE
+import com.kickstarter.libs.utils.EventContext.CtaContextName.PLEDGE_INITIATE
 import com.kickstarter.libs.utils.EventName
 import com.kickstarter.libs.utils.ExperimentData
 import com.kickstarter.libs.utils.KoalaUtils
@@ -13,8 +14,9 @@ import com.kickstarter.services.DiscoveryParams
 import com.kickstarter.services.apiresponses.PushNotificationEnvelope
 import com.kickstarter.ui.data.*
 import com.kickstarter.ui.data.Mailbox
-import java.util.*
 import kotlin.collections.HashMap
+
+private const val CONTEXT_CTA = "context_cta"
 
 class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -13,9 +14,10 @@ import com.kickstarter.services.DiscoveryParams
 import com.kickstarter.services.apiresponses.PushNotificationEnvelope
 import com.kickstarter.ui.data.*
 import com.kickstarter.ui.data.Mailbox
-import java.util.*
 import kotlin.collections.HashMap
 
+private const val CONTEXT_CTA = "context_cta"
```
Arkariang, line 19: Suggestion here, move the constant to `EventContext.kt` as companion object 

```
    companion object {
        const val CONTEXT_CTA = "context_cta"
        fun getPledgeInitiateCTAProps(): HashMap<String, Any> = hashMapOf(CONTEXT_CTA to CtaContextName.PLEDGE_INITIATE.contextName)
    }
```

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -625,6 +627,17 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track(PROJECT_PAGE_PLEDGE_BUTTON_CLICKED, props)
     }
 
+    /**
+     * Sends data associated with the initial pledge CTA click event to the client.
+     *
+     * @param projectData: The project data.
+     */
+    fun trackPledgeInitiateCTA(projectData: ProjectData) {
+        val props: HashMap<String, Any> = hashMapOf(CONTEXT_CTA to PLEDGE_INITIATE.contextName)
```
Arkariang, line 636: Following up the previous suggestion: 

```
val props: HashMap<String, Any> = getPledgeInitiateCTAProps()

```

Arkariang (approved):
Great work here! 
Just a small suggestion, feel free to ignore it if you don't like it :)

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
leighdouglas: So I purposely chose not to use a `companion object` because this isn't a global constant or method, and when you create a companion object, under the hood an unnecessary getter and setter are created to make these things accessible. Since these are not needed globally, it seems that this is unnecessary. 

https://stackoverflow.com/questions/44038721/constants-in-kotlin-whats-a-recommended-way-to-create-them

PR #1099, opened 2021-02-11
EP-219: Create event for Add-ons screen viewed

leighdouglas:
# 📲 What

- Added track event for add-ons screen that fires when the `BackingAddonsFragmentViewModel` is loaded. 
- Fixed instances of "CTA Clicked" in `ProjectViewModelTest` to use a constant instead of a string literal.

# 🤔 Why

Segment Integration

# 🛠 How

- Created a new function in the `AnalyticsEvents` class that handles sending the event name and properties to the client.
- Added this function to the `BackingAddonsFragmentViewModel` when the initialization method is called.

# 👀 See

This is the screen that fires this event:
Screenshot_1613083922

# 📋 QA

- Tap on a live project that has add-ons
- Select a reward which should then proceed to the add-ons screen 
- Both Segment (Page Viewed) and DataLake (Add-Ons Page Viewed) should appear in the segment dashboard.
- You should see `context_page = "add_ons" in the list of properties
Screen Shot 2021-02-11 at 6 21 04 PM
Screen Shot 2021-02-11 at 6 21 13 PM

# Story 📖

[EP-219: Android: Page Viewed (add_ons)](https://kickstarter.atlassian.net/browse/EP-219)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/EventContext.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/EventContext.kt
@@ -37,4 +37,10 @@ class EventContext {
         REWARDS("rewards"),
         THANKS("thanks")
     }
+
+    enum class ContextPropertyName(val contextName: String) {
+        CONTEXT_CTA("context_cta"),
+        CONTEXT_PAGE("context_page"),
+        CONTEXT_TYPE("context_type")
+    }
 }
--- a/app/src/main/java/com/kickstarter/libs/utils/EventContext.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/EventContext.kt
@@ -38,6 +38,16 @@ class EventContext {
         THANKS("thanks")
     }
 
+    /**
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/libs/utils/EventContext.kt
```diff
@@ -37,4 +37,20 @@ class EventContext {
...
+     * Represents the different context keys for properties sent to analytics
+     *
+     * @param contextName: The name of the context key.
+     *
+     * CONTEXT_CTA: Key that stores the button that was tapped.
+     * CONTEXT_PAGE: Key that stores the page that was viewed.
+     * CONTEXT_TYPE: Key that stores contextual details about an event that was
+     * fired that aren't captured in other context properties
+     */
+    enum class ContextPropertyName(val contextName: String) {
```
Arkariang, line 51: nice touch

Arkariang (approved):
lgtm!

PR #1106, opened 2021-02-18
EP-220: Create event for Checkout screen viewed

leighdouglas:
# 📲 What

- Added track event for add-ons screen that fires when the `PledgeViewFragmentViewModel` is loaded.

# 🤔 Why

Segement integration. 

# 🛠 How

- Created a new function in the AnalyticsEvents class that handles sending the event name and properties to the client.
- Added this function to the `PledgeViewFragmentViewModel` when the initialization method is called.
- This event only fires when the checkout is loaded from a new pledge, this does not fire when updating or managing a pledge. 
- 
# 👀 See

This is the screen that fires this event:
Screenshot_1613674651

# 📋 QA

- Tap on a live project 
- Select a reward or no reward
- If project has add-ons, press skip CTA or choose an addon and press continue
- Both Segment (Page Viewed) and DataLake (Checkout Payment Page Viewed) should appear in the segment dashboard.
You should see `context_page = "checkout" in the list of properties
Screen Shot 2021-02-18 at 1 55 01 PM
Screen Shot 2021-02-18 at 1 56 50 PM

# Story 📖

[EP-220: Android Page Viewed (checkout)](https://kickstarter.atlassian.net/browse/EP-220)


Arkariang (approved):
lgtm!

PR #1108, opened 2021-02-18
Reorganized key and value enums for segment.events

leighdouglas:
# 📲 What

I reorganized and renamed some of the enums we are using for segment to be more clear. 

# 🤔 Why

As I am adding more keys and values, it was becoming clear that I had put a few things in the wrong place. This will set us up to be better organized as we are moving forward. 

# 📋 QA

All segment work should be the same as before, just updated the imports. 

# Story 📖

N/A

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/EventKeys.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/EventKeys.kt
@@ -1,0 +1,21 @@
+package com.kickstarter.libs.utils
+
+class EventKeys {
+
+    /**
+     * Represents the different context keys for properties sent to analytics
+     *
+     * @param contextName: The name of the context key.
+     *
+     * CONTEXT_CTA: Key that stores the button that was tapped.
+     * CONTEXT_PAGE: Key that stores the page that was viewed.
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/libs/utils/EventKeys.kt
```diff
@@ -0,0 +1,21 @@
+package com.kickstarter.libs.utils
+
+class EventKeys {
```
Arkariang, line 3: You don't need this `EventKeys` class. This file can perfectly hold just the enum class.
The only difference that will make on `AnalyticsEvents.kt` is the imports instead of having 
`import com.kickstarter.libs.utils.EventKeys.ContextPropertyKeyName.CONTEXT_CTA` you'll get 
`import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_CTA`

leighdouglas: Got it! The only thing is, in the future, I planned to add additional enums to this class as there are more property keys. I could remove the `EventKeys` class and put them all under one generic enum `EventKeys`, or i can keep them separate and keep the class name. Up to you! I'm fine with either one.

PR #1115, opened 2021-02-21
NT-1785:Refactor CreatorDashboardActivity

hadia:
# 📲 What

Remove butter knife from CreatorDashboardActivity and replaced with view binding

# 🤔 Why

deprecated butterknife

# 🛠 How

1- change CreatorDashboardActivity from java to kotlin
2-remove butter-knife and replace it with viewbinding

# 👀 See
https://user-images.githubusercontent.com/1075310/108638723-40264400-7499-11eb-91ca-276847211d9a.mp4

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA
Open app choose Setting from side menu 
Open Dashboard

# Story 📖
https://kickstarter.atlassian.net/browse/NT-1785


Arkariang (approved):
lgtm! tested with theRealNative account 👍

app/src/main/java/com/kickstarter/ui/activities/CreatorDashboardActivity.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    private fun bottomSheetScrimClicked() {
+        viewModel.inputs.scrimClicked()
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        binding.creatorDashboardRecyclerView.adapter = null
+        binding.creatorDashboardBottomSheetRecyclerView.adapter = null
+        bottomSheetBehavior.setBottomSheetCallback(null)
```
leighdouglas, line 101: `setBottomSheetCallback()` is deprecated. Passing a null here just logs a statement to the console that this method is deprecated, and clears the callbacks but does nothing else, is that the intended purpose? If not i think we can remove this. If so, disregard!

leighdouglas: Ah now i'm seeing that this is in onDestroy, so it makes sense that we would want to clear the callbacks here. Nevermind!

leighdouglas (approved):
LGTM! Test on my local env and everything works great 🎉

PR #1123, opened 2021-02-23
NT-1792:Refactor FacebookConfirmationActivity

hadia:
# 📲 What

Remove butter knife from FacebookConfirmationActivity and replaced with view binding

# 🤔 Why

deprecated butterknife

# 🛠 How

1- change FacebookConfirmationActivity from java to kotlin
2-remove butter-knife and replace it with viewbinding

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

QA login in the app with a Facebook account for the production version

# Story 📖
https://kickstarter.atlassian.net/browse/NT-1792


Arkariang (approved):
lgtm! I tested it on my environment 👍

PR #1129, opened 2021-02-24
Sunday/nt 1058 reset password error dialog

sunday-okpoluaefe:
# 📲 What
- NT-1058: Reset password error dialogs are not shown
- 

# 🤔 Why

Show error dialogs for the failed operation.

# 🛠 How

- Changed observable type to boolean to separate errors related to invalid email and others ( Network timeout etc )
- Removed filter operator from the error observable so errors are passed to the view 

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |
Screenshot 2021-02-24 at 20 11 07

# 📋 QA

- Navigate to the Forgot Password screen (aka Reset Password) 

- Enter an invalid email (an unregistered email, for example)

- Tap the "Reset password" button
- An error is received from the backend (404) and an error dialog is displayed 

- Submit the "Reset Password" button again (with a registered or unregistered email) 
- An Error or success is shown

# Story 📖

https://kickstarter.atlassian.net/browse/NT-1058

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ResetPasswordActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ResetPasswordActivity.kt
@@ -22,7 +22,7 @@ import rx.android.schedulers.AndroidSchedulers
 class ResetPasswordActivity : BaseActivity<ResetPasswordViewModel.ViewModel>() {
 
     private var forgotPasswordString = R.string.forgot_password_title
-    private var errorMessageString = R.string.forgot_password_error
+    private var errorMessageString = R.string.Something_went_wrong_please_try_again
     private var errorTitleString = R.string.general_error_oops
 
     override fun onCreate(savedInstanceState: Bundle?) {
--- a/app/src/main/java/com/kickstarter/ui/activities/ResetPasswordActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ResetPasswordActivity.kt
@@ -51,7 +51,7 @@ class ResetPasswordActivity : BaseActivity<ResetPasswordViewModel.ViewModel>() {
                 .compose(bindToLifecycle())
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe {
-                    var message: Int = if( it == true ) this.errorNetworkMessageString
+                    val message: Int = if( it == true ) this.errorNetworkMessageString
                     else this.errorMessageString
                     ViewUtils.showDialog(this, getString(this.errorTitleString), getString(message)) }
 
--- a/app/src/main/java/com/kickstarter/ui/activities/ResetPasswordActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ResetPasswordActivity.kt
@@ -22,7 +22,8 @@ import rx.android.schedulers.AndroidSchedulers
 class ResetPasswordActivity : BaseActivity<ResetPasswordViewModel.ViewModel>() {
 
     private var forgotPasswordString = R.string.forgot_password_title
-    private var errorMessageString = R.string.Something_went_wrong_please_try_again
+    private var errorMessageString = R.string.forgot_password_error
+    private var errorNetworkMessageString = R.string.social_error_no_internet_connection
     private var errorTitleString = R.string.general_error_oops
 
     override fun onCreate(savedInstanceState: Bundle?) {
@@ -49,7 +50,10 @@ class ResetPasswordActivity : BaseActivity<ResetPasswordViewModel.ViewModel>() {
         this.viewModel.outputs.resetError()
                 .compose(bindToLifecycle())
                 .observeOn(AndroidSchedulers.mainThread())
-                .subscribe { ViewUtils.showDialog(this, getString(this.errorTitleString), getString(this.errorMessageString)) }
+                .subscribe {
+                    var message: Int = if( it == true ) this.errorNetworkMessageString
+                    else this.errorMessageString
+                    ViewUtils.showDialog(this, getString(this.errorTitleString), getString(message)) }
 
         reset_password_button.setOnClickListener { this.viewModel.inputs.resetPasswordClick() }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/viewmodels/ResetPasswordViewModel.kt
```diff
@@ -84,14 +87,20 @@ interface ResetPasswordViewModel {
 
             resetPasswordNotification
                     .compose(errors())
-                    .map { ErrorEnvelope.fromThrowable(it) }
-                    .filter { ObjectUtils.isNotNull(it) }
+                    .map {
+                        // host -> internet related error or otherwise -> invalid email address
+                        if( it !=  null) return@map it.localizedMessage!!.contains("host")
```
Arkariang, line 92: not forcing optionals, `!!` is not allowed, use the `?.let` unwrapping technique or any other technique

app/src/main/java/com/kickstarter/ui/activities/ResetPasswordActivity.kt
```diff
@@ -49,7 +50,10 @@ class ResetPasswordActivity : BaseActivity<ResetPasswordViewModel.ViewModel>() {
         this.viewModel.outputs.resetError()
                 .compose(bindToLifecycle())
                 .observeOn(AndroidSchedulers.mainThread())
-                .subscribe { ViewUtils.showDialog(this, getString(this.errorTitleString), getString(this.errorMessageString)) }
+                .subscribe {
+                    val message: Int = if( it == true ) this.errorNetworkMessageString
```
Arkariang, line 54: We need to avoid any logic in the Activities. Any bussines logic should live on the viewmodel in order to test it.
What I mean here is that you should emit from the ViewModel directly the string you wanna display on the dialog

Arkariang (changes requested):

app/src/main/java/com/kickstarter/ui/activities/ResetPasswordActivity.kt
leighdouglas: +1 yes this logic should be moved into the view model. We do not have tests for activities but our viewmodels are backed by tests.

app/src/main/java/com/kickstarter/viewmodels/ResetPasswordViewModel.kt
leighdouglas: I think we can simplify this line to be:
`.map { it?.localizedMessage?.contains("host") ?: false }`

If i'm not mistaken, this will return false if any of the variables are null or if the localized message does not contain "host". 

Also, nitpicky request, but could we move "host" into a constant instead of a string literal? Up to you, that is just a personal preference.

app/src/main/java/com/kickstarter/viewmodels/ResetPasswordViewModel.kt
```diff
@@ -84,14 +87,20 @@ interface ResetPasswordViewModel {
...
+                        else return@map false
+                    }
+                    //.filter { ObjectUtils.isNotNull(it) }
```
leighdouglas, line 101: nit: remove commented out code

leighdouglas:
Great first PR @sunday-okpoluaefe ! 🎉 As soon as you address the comments we can approve. Also, if you could cleanup the PR summary a bit, that would be great. For example, the title should be formatted as so: 
"NT-1058: Reset password error dialogs are not shown"

Likewise, it is really helpful to post screenshots if this change effects the UI, which in this case it does. 

Great work! 💯

app/src/main/java/com/kickstarter/viewmodels/ResetPasswordViewModel.kt
```diff
@@ -84,14 +87,20 @@ interface ResetPasswordViewModel {
...
+                    .takeUntil(this.resetSuccess)
                     .compose(bindToLifecycle())
                     .subscribe(this.resetError)
 
             this.lake.trackForgotPasswordPageViewed()
         }
 
+
```
leighdouglas, line 109: nit: remove extra whitespace

app/src/main/java/com/kickstarter/viewmodels/ResetPasswordViewModel.kt
```diff
@@ -12,6 +12,7 @@ import com.kickstarter.services.ApiClientType
 import com.kickstarter.services.apiresponses.ErrorEnvelope
 import com.kickstarter.ui.IntentKey
 import com.kickstarter.ui.activities.ResetPasswordActivity
+import com.noveogroup.android.log.Log
```
Arkariang, line 15: this import should be removed

app/src/main/java/com/kickstarter/viewmodels/ResetPasswordViewModel.kt
```diff
@@ -127,9 +129,10 @@ interface ResetPasswordViewModel {
         override fun resetError(): Observable<String> {
             return this.resetError
                     .takeUntil(this.resetSuccess)
-                    .map { it.errorMessage() }
+                    //.map { it.errorMessage() } /*  null pointer exception is thrown */
```
Arkariang, line 132: is probably returning an exception because the filter on line 90 was deleted, bring back the filter and do this
 `.map { it.errorMessage() ?: ""}`
If you do `.map{""}` you will always return an empty string instead of the error envelope message, you want to return the empty string just in case there is not message on the error envelope I guess

app/src/main/java/com/kickstarter/viewmodels/ResetPasswordViewModel.kt
Arkariang: Actually bring back the filter, we do want to avoid sending null objects to the `resetError` observable

Arkariang (changes requested):

app/src/main/java/com/kickstarter/viewmodels/ResetPasswordViewModel.kt
leighdouglas: Question: If we bring back the `.filter()` operator to filter out null emissions, wouldn't it not emit anything at all to the `resetError` stream?

leighdouglas: @Arkariang Question: If we bring back the .filter(ObjectUtils.isNotNull(it)) operator to filter out null emissions, wouldn't it not emit anything at all to the resetError stream?

PR #1139, opened 2021-02-27
EP-243: CTA clicked search

sunday-okpoluaefe:
# 📲 What

Added track event for "search" button clicks that fires when the "Search" button is pressed on the discovery screen.

# 🤔 Why

Segment integration.

# 🛠 How

- Added a new function to the AnalyticsEvents class that handles sending the event name and properties to the client.
- Called this function from the `SearchViewModel` when the "SEARCH" button is pressed.

# 👀 See
Screenshot 2021-02-27 at 20 23 09

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

 - Click the search icon/button on the discovery page/screen
 - (CTA Clicked) event should appear in the segment dashboard:

Screenshot 2021-02-27 at 20 33 21

# Story 📖

https://kickstarter.atlassian.net/browse/EP-243


app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -666,6 +667,16 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track(SEARCH_BUTTON_CLICKED)
     }
 
+    /**
+     * Sends data associated with the search CTA click event to the client.
+     *
+     * @param discoveryParams: The discovery data.
+     */
+    fun trackSearchCTAButtonClicked(discoveryParams: DiscoveryParams) {
+        val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams)
```
leighdouglas, line 676: We need to add the property `context_cta` with a value of `search`, as well as the `context_location` with a value of `global_nav`, to the `props` object

leighdouglas (changes requested):
Great job! Only requesting changes because we are missing the "context_cta" and the "context_location" properties which are required by the ticket. Also, please reformat the title of this PR to be "EP-243: CTA clicked search".

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -2,9 +2,7 @@ package com.kickstarter.libs
 
 import com.kickstarter.libs.KoalaContext.*
 import com.kickstarter.libs.KoalaEvent.ProjectAction
-import com.kickstarter.libs.utils.BooleanUtils
-import com.kickstarter.libs.utils.ExperimentData
-import com.kickstarter.libs.utils.AnalyticEventsUtils
+import com.kickstarter.libs.utils.*
```
leighdouglas, line 5: Sorry I didn't notice this before. We avoid using star-imports on new code. If you could revert these changes and add whatever specific utils classes you need, that would be great. Thanks!

leighdouglas (dismissed):
Looks great! As soon as you revert the star-imports, feel free to land 🚀 Thanks again!

leighdouglas (approved):
Left a small comment, otherwise looks great!

leighdouglas (approved):
Great job! 🎉

PR #1144, opened 2021-03-03
EP-237: CTA Clicked (discover)

sunday-okpoluaefe:
# 📲 What

Added track event for "discover projects" button clicks in the activity screen

# 🤔 Why

Segment Implementation 

# 🛠 How

  - Create a new method that submits CTA click event to segment
  - Called method from ActivityFeedViewModel class for every click on "Discover Projects" button

# 👀 See

Screenshot 2021-03-03 at 17 09 12

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

- Navigate to the Activity Screen. 
- Click "Discover Projects" button. You should the event registered on segment dashboard

Screenshot 2021-03-03 at 20 23 40

# Story 📖

https://kickstarter.atlassian.net/browse/EP-237

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -208,6 +208,16 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         }
     }
 
+    /**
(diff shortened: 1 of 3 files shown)


Arkariang (commented):
lgtm!

Arkariang (approved):

PR #1145, opened 2021-03-03
EP-163: Create properties checkout

Arkariang:
# 📲 What

Added properties: 
- checkout_add_ons_count_total
- checkout_add_ons_count_unique
- checkout_add_ons_minimum_usd
- checkout_bonus_amount_usd
- checkout_reward_minimum_usd
- checkout_shipping_amount_usd:

# 🛠 How
- All addOns related properties are collected on Checkout step, but storaged on pledgeData.

# 👀 See
Project_WithAddOns_Unique
Project_WAddOns
Project_WOAddOns_No_USD
|  |  |

# 📋 QA
- Pledge to a project with addOns, if the project is from US the values will be the same as you see on the `Manage your pledge ` screen
- Pledge to a project with addOns, but a project from other country different from US, on this new properties you'll se the conversion from the project currency to us dolars

# Story 📖

[EP-163](Thttps://kickstarter.atlassian.net/browse/EP-163)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/AnalyticEventsUtils.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/AnalyticEventsUtils.kt
@@ -4,13 +4,14 @@ import com.kickstarter.libs.RefTag
 import com.kickstarter.libs.utils.RewardUtils.isItemized
 import com.kickstarter.libs.utils.RewardUtils.isShippable
 import com.kickstarter.libs.utils.RewardUtils.isTimeLimitedEnd
-import com.kickstarter.libs.utils.extensions.totalAmount
+import com.kickstarter.libs.utils.extensions.*
 import com.kickstarter.models.*
 import com.kickstarter.services.DiscoveryParams
 import com.kickstarter.ui.data.CheckoutData
 import com.kickstarter.ui.data.PledgeData
 import java.util.*
 import kotlin.math.ceil
+import kotlin.math.round
 import kotlin.math.roundToInt
 
 object AnalyticEventsUtils {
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/kickstarter/libs/utils/extensions/PledgeDataExt.kt
```diff
@@ -0,0 +1,46 @@
+@file:JvmName("PledgeDataExt")
```
leighdouglas, line 1: nit: I don't think we need this annotation as the compiler is already generating this java class with this name under the hood. Might be good to keep it though for consistency/readability.

app/src/main/java/com/kickstarter/libs/utils/extensions/CheckoutDataExt.kt
```diff
@@ -9,7 +9,39 @@ import com.kickstarter.ui.data.CheckoutData
...
+ * @return Double
+ */
+fun CheckoutData.totalAmount(usdRate: Float) = this.totalAmount() * usdRate
+
+/**
+ * Returns the bonus amount added to the pledge
+ *
+ * @return Double
+ */
+fun CheckoutData.bonus() = this.bonusAmount()?.let { it } ?: 0.0
```
leighdouglas, line 33: nit: the `let` call can be removed

app/src/main/java/com/kickstarter/libs/utils/AnalyticEventsUtils.kt
```diff
@@ -4,13 +4,14 @@ import com.kickstarter.libs.RefTag
 import com.kickstarter.libs.utils.RewardUtils.isItemized
 import com.kickstarter.libs.utils.RewardUtils.isShippable
 import com.kickstarter.libs.utils.RewardUtils.isTimeLimitedEnd
-import com.kickstarter.libs.utils.extensions.totalAmount
+import com.kickstarter.libs.utils.extensions.*
```
leighdouglas, line 7: nit: remove wild card import

leighdouglas (approved):
LGTM! Left a few nit pick comments but otherwise awesome! Also great to have a pledge data extension class 🎉 🚀

PR #1152, opened 2021-03-10
EP-352: Add context page property to all CTA events and update enum classes

leighdouglas:
# 📲 What

- Added the `context_page` property to all CTA click events.
- Renamed and reorganized enum values by property
- Update enum documentation

# 🤔 Why

The `context_page` property is required on all CTA click events as a part of segment integration. Previously we were not sending it. 

# 🛠 How

- Updated CTA click events in the `AnalyticEvents` class with the property.
- Reorganized and renamed some enums to reflect their property keys
- Added more documentation.

# 📋 QA

All CTA click events should now have the property `context_page` with the corresponding value associated with that event. Please see previous CTA click event tickets for QA instructions.

# Story 📖

[EP-352: Add missing context_page property to CTA clicked events](https://kickstarter.atlassian.net/browse/EP-352)


Arkariang (approved):
lgtm, once the conflicts are resolved is ok to go :)

PR #1154, opened 2021-03-10
[NO-JIRA]: Remove deprecated method

Arkariang:
# 📲 What

`this.currentUser.exists()` is deprecated and should not be used. Instead in order to get the current user, observable method should be used and subscribe to it.

Screen Shot 2021-03-10 at 2 05 45 PM

|  |  |

# 📋 QA
- You should be able to log in and make calls to V1 as usual


sunday-okpoluaefe:
lgtm!

sunday-okpoluaefe (approved):

PR #1155, opened 2021-03-11
EP-284: Android: CTA Clicked (campaign_details)

sunday-okpoluaefe:
# 📲 What

Android: CTA Clicked (campaign_details) 

# 🤔 Why

Segment Implementation

# 🛠 How

- Created new method ``` trackCampaignDetailsCTAClicked``` in ``` AnalyticEvents ``` class
- Called method from ``` ProjectViewModel ``` on ``` campaign ``` button clicked

# 👀 See

Screenshot 2021-03-11 at 15 18 07

Screenshot 2021-03-11 at 15 26 47

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

- Goto ``` Project Details ``` page. 
- Select ``` campaign ```.
 You should see the event ``` CTA Clicked ``` registered

Screenshot 2021-03-11 at 15 27 12

# Story 📖

https://kickstarter.atlassian.net/browse/EP-284


app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -2,53 +2,26 @@ package com.kickstarter.libs
...
+import com.kickstarter.libs.utils.EventName.*
+import com.kickstarter.libs.utils.ExperimentData
 import com.kickstarter.models.Activity
 import com.kickstarter.models.Project
 import com.kickstarter.models.User
 import com.kickstarter.services.DiscoveryParams
 import com.kickstarter.services.apiresponses.PushNotificationEnvelope
 import com.kickstarter.ui.data.*
 import com.kickstarter.ui.data.Mailbox
-import java.util.Locale
```
Arkariang, line 51: nit: we try to not use wild card imports

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -2,53 +2,26 @@ package com.kickstarter.libs
...
-import com.kickstarter.libs.utils.EventName.CTA_CLICKED
-import com.kickstarter.libs.utils.EventName.PAGE_VIEWED
-import com.kickstarter.libs.utils.EventName.CARD_CLICKED
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_CTA
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_TYPE
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_PAGE
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_SECTION
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_LOCATION
+import com.kickstarter.libs.utils.BooleanUtils
+import com.kickstarter.libs.utils.ContextPropertyKeyName.*
```
leighdouglas, line 7: Can we revert these back to not use the wild card imports? Thanks!

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -2,53 +2,26 @@ package com.kickstarter.libs
...
 import com.kickstarter.libs.utils.EventContextValues.CtaContextName.DISCOVER
-import com.kickstarter.libs.utils.EventContextValues.LocationContextName.DISCOVER_ADVANCED
-import com.kickstarter.libs.utils.EventContextValues.LocationContextName.DISCOVER_OVERLAY
-import com.kickstarter.libs.utils.EventContextValues.LocationContextName.GLOBAL_NAV
-import com.kickstarter.libs.utils.EventName
+import com.kickstarter.libs.utils.EventContextValues.CtaContextName.SEARCH
+import com.kickstarter.libs.utils.EventContextValues.DiscoveryContextType.*
+import com.kickstarter.libs.utils.EventContextValues.LocationContextName.*
+import com.kickstarter.libs.utils.EventContextValues.PageViewedContextName.*
+import com.kickstarter.libs.utils.EventName.*
```
leighdouglas, line 15: Please revert these as well

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -956,6 +929,13 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track(CAMPAIGN_DETAILS_BUTTON_CLICKED, experimentProperties(projectData))
     }
 
+    fun trackCampaignDetailsCTAClicked(projectData: ProjectData) {
+        val props: HashMap<String, Any> = hashMapOf(CONTEXT_PAGE.contextName to CAMPAIGN_DETAILS.contextName)
```
leighdouglas, line 933: We should be using the `context_cta` key, not `context_page`

leighdouglas, line 933: I think this is the wrong value, since we're tapping the CTA on the project page. I think the key should be `project`

leighdouglas (changes requested):
Looks great! Requesting changes since Laurie confirmed that we should be sending `context_page = project` on this event. Also if you could revert the wildcard imports that would be great. Thank you!

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
sunday-okpoluaefe: sure. thanks

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
sunday-okpoluaefe: Yeah noticed that too. Thanks for pointing that out. Corrected

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -956,6 +956,12 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track(CAMPAIGN_DETAILS_BUTTON_CLICKED, experimentProperties(projectData))
     }
 
+    fun trackCampaignDetailsCTAClicked(projectData: ProjectData) {
```
leighdouglas, line 959: This is missing the `context_page = project` property.

leighdouglas (approved):
LGTM! 🚀 🎉

PR #1156, opened 2021-03-11
EP-317:Android: CTA Clicked (creator_details)

hadia:
# 📲 What

Added the CREATOR_DETAILS property to all CTA context name.

# 🤔 Why

Handling Creator Details Clicked from Project screen 

# 🛠 How

update fun trackCreatorDetailsClicked in AnalyticsEvents class that handles sending the event name and properties to the client.

image

# 👀 See

image

# 📋 QA

Open Project details  screen
Click on Creator  data
image
image

# Story 📖

https://kickstarter.atlassian.net/browse/EP-317


app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -2,53 +2,26 @@ package com.kickstarter.libs
...
-import com.kickstarter.libs.utils.EventContextValues.DiscoveryContextType.WATCHED
-import com.kickstarter.libs.utils.EventName.CTA_CLICKED
-import com.kickstarter.libs.utils.EventName.PAGE_VIEWED
-import com.kickstarter.libs.utils.EventName.CARD_CLICKED
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_CTA
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_TYPE
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_PAGE
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_SECTION
-import com.kickstarter.libs.utils.ContextPropertyKeyName.CONTEXT_LOCATION
+import com.kickstarter.libs.utils.BooleanUtils
```
leighdouglas, line 6: Can we please revert these back? We don't use wildcard imports. Thank you!

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -961,7 +934,9 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
     }
 
     fun trackCreatorDetailsClicked(projectData: ProjectData) {
-        client.track(CREATOR_DETAILS_CLICKED, experimentProperties(projectData))
+        val props: HashMap<String, Any> = hashMapOf(CONTEXT_CTA.contextName to CREATOR_DETAILS.contextName)
```
leighdouglas, line 937: We are leaving all koala events as is and creating new methods for segment events and sending them alongside the old koala events.

leighdouglas (changes requested):
Thanks for picking up this event ticket! Requesting changes for a few reasons:
- We should be creating new events for segment and leaving old koala events alone
- We need to send the new event where we are sending the old koala event, so essentially two events
- All events need a `context_page` property
- Wildcard imports should be reverted

Thanks again!

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
hadia: ok thanks

hadia: @leighdouglas 
image

leighdouglas: Great!

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -964,6 +965,13 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track(CREATOR_DETAILS_CLICKED, experimentProperties(projectData))
     }
 
+    fun trackCreatorDetails(projectData: ProjectData) {
+        val props: HashMap<String, Any> = hashMapOf(CONTEXT_CTA.contextName to CREATOR_DETAILS.contextName)
+        props[CONTEXT_PAGE.contextName] = PROJECT.contextName
+        props.putAll(experimentProperties(projectData))
```
leighdouglas, line 971: one small thing here, the `experimentProperties()` function here is calling sending an event to optimizely, which we don't need for this. I would change this to `props.putAll(AnalyticEventsUtils.projectProperties(project, client.loggedInUser()))`

leighdouglas (approved):
LGTM! Small comment about calling `experimentProperties` vs `projectProperties`, otherwise great!! 🚀 🎉

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
hadia: Resolved

PR #1162, opened 2021-03-13
EP-273: Android: CTA Clicked (log_in_or_sign_up)

sunday-okpoluaefe:
# 📲 What

Register SignIn or Sign Up events:
- Log In or Sign Up Button Clicked,
- Facebook Log In or Sign Up Button Clicked

# 🤔 Why

Segment Implementation

# 🛠 How

- Created method ```  trackLoginOrSignUpCtaClicked  ```
``` 
fun trackLoginOrSignUpCtaClicked(type: String?, page: String) {
        val props: HashMap<String, Any> = hashMapOf(CONTEXT_CTA.contextName to LOGIN_OR_SIGN_UP.contextName)
        props[CONTEXT_PAGE.contextName] = page
        type?.let { props[CONTEXT_TYPE.contextName] = it }
        client.track(CTA_CLICKED.eventName, props)
    }
```

- Called method from ``` DiscoveryFragmentViewModel ``` and ``` LoginToutViewModel ``` respectively
 ```
 this.discoveryOnboardingLoginToutClick
        .compose(bindToLifecycle())
        .subscribe(v -> {
          this.lake.trackLogInSignUpButtonClicked();
          this.lake.trackLoginOrSignUpCtaClicked(null, EventContextValues.ContextPageName.DISCOVER.getContextName());
        });
```

```
 this.facebookLoginClick
        .compose(ignoreValues())
        .compose(bindToLifecycle())
        .subscribe(__ -> {
          this.lake.trackFacebookLogInSignUpButtonClicked();
          this.lake.trackLoginOrSignUpCtaClicked(ContextTypeName.FACEBOOK.getContextName(), ContextPageName.LOGIN_SIGN_UP.getContextName());
        });
```

# 👀 See

Screenshot 2021-03-13 at 06 01 42

Screenshot 2021-03-13 at 04 40 35

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

- From the ``` Native Discovery ``` screen, Click the ``` Sign Up or Log in ``` button
- From the ``` Sign Up Or Sign In ``` screen, click the ``` Log in with Facebook ``` button

- You should the two events registered in segment console

Screenshot 2021-03-13 at 06 25 44

Screenshot 2021-03-13 at 06 26 15

# Story 📖

https://kickstarter.atlassian.net/browse/EP-273

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -276,7 +276,7 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
     fun trackLoginOrSignUpCtaClicked(type: String, page: String) {
         val props: HashMap<String, Any> = hashMapOf(CONTEXT_CTA.contextName to LOGIN_OR_SIGN_UP.contextName)
         props[CONTEXT_PAGE.contextName] = page
-        type.let { props[CONTEXT_TYPE.contextName] = type }
+        props[CONTEXT_TYPE.contextName] = type
         client.track(CTA_CLICKED.eventName, props)
     }
 
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -270,6 +270,12 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track("Application Login or Signup", properties)
     }
 
+    fun trackLoginOrSignUpCtaClicked(type: String) {
+        val props: HashMap<String, Any> = hashMapOf(CONTEXT_CTA.contextName to LOGIN_OR_SIGN_UP.contextName)
+        props[CONTEXT_PAGE.contextName] = LOGIN_SIGN_UP.contextName
+        client.track(CTA_CLICKED.eventName, props)
+    }
+
     fun trackLoginSuccess() {
         client.track(KoalaEvent.LOGIN)
     }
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt [hunk 1]
```diff
@@ -271,6 +272,19 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track("Application Login or Signup", properties)
     }
 
+    fun trackLoginOrSignUpCtaClicked(type: String, page: String) {
+        val props: HashMap<String, Any> = hashMapOf(CONTEXT_CTA.contextName to LOGIN_OR_SIGN_UP.contextName)
+        props[CONTEXT_PAGE.contextName] = page
+        type.let { props[CONTEXT_TYPE.contextName] = type }
```
Arkariang, line 278: why using .let in a non optional ?

sunday-okpoluaefe: The method is called from a java class. I had to check for null

Arkariang: that's ok, then declare on the kotlin side he parameter as optional, and unwrap it 👍  ` fun trackLoginOrSignUpCtaClicked(type: String?, page: String)`

sunday-okpoluaefe: thanks

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -272,6 +273,28 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
...
 
+    /**
+     * Tracks a login or sign up button clicked.
+     * @param type
+     * @param page
+     */
+    fun trackLoginOrSignUpCtaClicked(type: String?, page: String) {
+        val props: HashMap<String, Any> = hashMapOf(CONTEXT_CTA.contextName to LOGIN_OR_SIGN_UP.contextName)
+        props[CONTEXT_PAGE.contextName] = page
+        type?. let { props[CONTEXT_TYPE.contextName] = it }
```
leighdouglas, line 284: Do we need an overloaded method for this? We can use the same method and pass in null for `type` when it is not needed.

sunday-okpoluaefe: yeah, correct since I have made the parameter optional. thanks

Arkariang (approved):
lgtm!

PR #1176, opened 2021-03-19
NT-1811: Regression-bugs

Arkariang:
# 📲 What

- After running a manual regression throwout the app two minor issues have been found
- When looking at the updates for a project without any published update, there is an empty state view loaded, that was not being showed.

# 🛠 How
- Declare as optional and unwrap the URL on `HelpActivity.kt` -> here the crash it was causing, I could not reproduce it manually https://console.firebase.google.com/u/1/project/android-internal-release/crashlytics/app/android:com.kickstarter.kickstarter.internal/issues/0bbc43dd6be74c4f679fd471dd478d74
- Load the ViewBinding associated with `R.layout.empty_updates_layout` on the `EmptyViewHolder`
- make `EmptyViewHolder` constructor with the ViewBinding super class as we are not accessing any view attributes when using EmptyViewHolder

# 👀 See

| **BEFORE** 🐛| 

https://user-images.githubusercontent.com/4083656/111846456-17d90c00-88c4-11eb-99a9-e6431c07d968.mp4

| **NOW** 🦋 |

https://user-images.githubusercontent.com/4083656/111846469-1f001a00-88c4-11eb-8879-95b342b5a432.mp4

|  |  |

# 📋 QA
- Go to a project without any updates (- usually the ones under the newest tab do not have updates yet), load the project scree, tab on updates, you should see the screen with an icon an a text instead of a empty screen

# Story 📖

[NT-1811](https://kickstarter.atlassian.net/browse/NT-1811)


hadia (approved):
Thanks 🤩

PR #1180, opened 2021-03-24
[Android] EP-389: Phase 1 QA Feedback Priority 1 properties

leighdouglas:
# 📲 What

- Updated `session_current_variants -> session_variants_internal`
- Added `user_backed_projects_count` to user properties
- Added `user_launched_projects_count` to user properties
- Updated `currentVariants` to return an array instead of a string
- Updated the `SegmentTest` and `LakeTest` class

# 🤔 Why

Segment integration.

# 🛠 How

- Updated the property `session_current_variants` in the `sessionProperties()` function in the `TrackingClient` class, and updated the method that is populating this field to return an array instead of a string. Updated the tests for this new logic as well. 
- Created the property `user_backed_projects_count` in the `userProperties()` function in the `AnalyticEventsUtils` ext
- Created the property `user_launched_projects_count` in the `userProperties()` function in the `AnalyticEventsUtils` ext
- Updated all tests in the `SegmentTestClass` that are calling an event where user properties would be sent

# 📋 QA

- To test this, you should fire off an event in the android app and see that the correct property is in segment.
- The user properties will only fire when the user is logged in

Example:
- Open App and login if necessary, make sure the account you are using has **backed** AND **launched** projects
- Tap on a project
- In segment, you should see the "Page Viewed" event with `context_page = project`. You should also see the following properties:
Screen Shot 2021-03-24 at 4 02 15 PM

The raw JSON for this property should look like this: 
Screen Shot 2021-03-24 at 4 02 26 PM

# Story 📖

[EP-389: [Android] Create Properties: Phase 1 QA Feedback Priority 1](https://kickstarter.atlassian.net/browse/EP-389)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/TrackingClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/TrackingClient.kt
@@ -107,7 +107,7 @@ abstract class TrackingClient(
 
     override fun buildNumber(): Int = BuildConfig.VERSION_CODE
 
-    override fun currentVariants(): JSONArray? = this.config?.currentVariants()
+    override fun currentVariants(): Array<String>? = this.config?.currentVariants()
 
     override fun deviceDistinctId(): String = FirebaseInstanceId.getInstance().id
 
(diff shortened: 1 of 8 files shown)


Arkariang (approved):
lgtm!

hadia (approved):
🎉👌

PR #1181, opened 2021-03-24
[Android] EP-387: Phase 1 QA Create Property session_country 

leighdouglas:
# 📲 What

- Create `session_country` property
- Update `SegmentTest` suite

# 🤔 Why

Segment integration.

# 🛠 How

Added `session_country` to the session properties, which gets its value from the `country code` in the `config` object

# 📋 QA

- Open app
- Fire any event, for example, tap on a project card
- In segment, under that event, you should see the `session_country` property:
Screen Shot 2021-03-24 at 4 30 42 PM

# Story 📖

[EP-387: [Android] Phase 1 QA Create Property: session_country](https://kickstarter.atlassian.net/browse/EP-387)


Arkariang (approved):
lgtm

PR #1183, opened 2021-03-24
[Android] EP-390: Phase 1 QA Feedback Priority 2 properties

leighdouglas:
# 📲 What

Added 3 new props to user properties: 
- `user_created_projects_count`
- `user_facebook_connected`
- `user_watched_projects_count`

Updated the `SegmentTest` suite

# 🤔 Why

Segment integration.

# 🛠 How

- Added all three properties to the` userProperties()` function in the `AnalyticEventsUtils` ext
- Updated the `assertUserProperties()` function in the `SegmentTest` class as well as the test case that checks we are returning the correct value if the fields from v1 that are populating these properties are null

# 📋 QA

- Open app and fire any event. For example, tap on a project card
- In segment, you should see the three new properties:
Screen Shot 2021-03-24 at 7 07 10 PM

# Story 📖

[EP-390: [Android] Create Properties Phase 1 QA Feedback Priority 2](https://kickstarter.atlassian.net/browse/EP-390)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/AnalyticEventsUtils.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/AnalyticEventsUtils.kt
@@ -113,6 +113,9 @@ object AnalyticEventsUtils {
         val properties = HashMap<String, Any>()
         properties["backed_projects_count"] = user.backedProjectsCount() ?: 0
         properties["launched_projects_count"] = user.memberProjectsCount() ?: 0
+        properties["created_projects_count"] = user.createdProjectsCount() ?: 0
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/kickstarter/libs/MockTrackingClient.java
```diff
@@ -113,7 +113,7 @@ protected String deviceFormat() {
 
   @Override
   protected String deviceOrientation() {
-    return "Portrait";
+    return "portrait";
```
leighdouglas, line 116: This will be removed once the previous PR is landed and I rebase.

hadia (approved):
👌🎉

PR #1186, opened 2021-03-31
NT-1823: isExpanded functionality on BaseFragment

Arkariang:
# 📲 What

- `BaseFragmentViewModel` emits if the fragment is expanded or not
- `BaseFragment` has a new method to setUp the state.

# 🤔 Why
- Fuctionallity to know when a Fragment has suffer a transformation via Animation outside the fragment itself

# 👀 See
- For the PageViewed event on the RewardsCarousel you can see this working, every time we show in full screen the rewards carousel the page viewed event should be fired.
- Tests and code already added for the `pageViewed event` for rewards carousel @leighdouglas 
- PlaceHolder on `BackingFragmentViewModel` for the `Manage Pledge Screen PageViewed` event @sunday-okpoluaefe 

|  |  |

# 📋 QA
- Open the Rewards Carousel you'll see that the event is fired just when the carousel is expanded

# Story 📖

[NT-1823](https://kickstarter.atlassian.net/browse/NT-1823)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/fragments/BackingFragment.kt
+++ b/app/src/main/java/com/kickstarter/ui/fragments/BackingFragment.kt
@@ -26,6 +26,7 @@ import com.kickstarter.libs.utils.ObjectUtils
 import com.kickstarter.libs.utils.ViewUtils
 import com.kickstarter.libs.utils.extensions.setGone
 import com.kickstarter.models.Reward
+import com.kickstarter.ui.activities.BackingActivity
 import com.kickstarter.ui.adapters.RewardAndAddOnsAdapter
 import com.kickstarter.ui.data.PledgeStatusData
 import com.kickstarter.ui.data.ProjectData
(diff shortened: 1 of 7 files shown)


sunday-okpoluaefe (approved):
lgtm!

app/src/main/java/com/kickstarter/ui/fragments/BackingFragment.kt
```diff
@@ -254,6 +255,11 @@ class BackingFragment : BaseFragment<BackingFragmentViewModel.ViewModel>() {
         }
     }
 
+    override fun onStart() {
+        super.onStart()
+        this.setState(activity is BackingActivity)
```
leighdouglas, line 260: Just curious, how is `activity is BackingActivity` related to the fragments visibility?

Arkariang: when the fragment is loaded from the activity BackingActivity, there is not animation, so the state is set to always be expanded

PR #1188, opened 2021-04-01
EP-241: Fix properties for discover filter CTA clicks

leighdouglas:
# 📲 What

Specs on the discover sort CTA clicks have changed to be the following:
- `discover_sort` is the sort the screen was on before the CTA was tapped
- `context_type` is the sort that was selected

`discover_sort` = `magic, popular, newest, ending_soon`
`context_location` = `discover_advanced`
`context_type` = `magic, popular, newest, ending_soon`

# 🤔 Why

Segment integration.

# 🛠 How

- Updated the analytic event `trackDiscoverSortCTA` to take in a parameter "current sort" which represents the sort the user was on when they tapped a different sort
- Updated the discover view model to hold on to the previous and newly selected sort values
- Updated the discoveryParamsProperties method to take an optional currentSort value, which defaults to value of `params.sort()` unless explicitly passed in (this accounts for the use of this method in other analytic events)

# 👀 See

These are the CTAs that are tapped to trigger this event:
Screenshot_1614288807

# 📋 QA

- Open application
- Tap on "Popular" at the top of the screen
- In the segment dashboard, you should see

```
context_cta = discover_sort
context_location = discover_advanced
context_page = discover
context_type = popular
discover_sort = magic
```
Screen Shot 2021-04-06 at 11 40 22 AM

# Story 📖

[EP-241: Android - CTA Clicked (discover_sort)](https://kickstarter.atlassian.net/browse/EP-241)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -730,11 +730,13 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
      *
      * @param discoveryParams: The discovery parameters.
      */
-    fun trackDiscoverSortCTA(discoveryParams: DiscoveryParams) {
+    fun trackDiscoverSortCTA(currentSort: DiscoveryParams.Sort, discoverParams: DiscoveryParams) {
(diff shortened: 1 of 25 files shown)


Arkariang (approved):
lgtm! great work @leighdouglas

PR #1193, opened 2021-04-05
NT-1818/Android Feature Flag Toggle

hadia:
# 📲 What

Change the feature flag screen to enable change segment value 

# 🤔 Why
Toggle segment flag value from an internal tool 

# 🛠 How

1. Change Ui ti toggle and disable change for other feature flags
2. Update currentConfig object with the new flag value 
3.update the resume app method to keep value if the app was sent to the background 

# 👀 See
image
image

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

https://user-images.githubusercontent.com/1075310/113621378-d71b1a00-965b-11eb-96bb-11a4ab38b16b.mp4

# 📋 QA

1. open internal tools from side menus
2. click featureflag
3. change segment value 

# Story 📖

https://kickstarter.atlassian.net/browse/NT-1818


app/src/internal/java/com/kickstarter/ui/viewholders/FeatureFlagViewHolder.kt
```diff
@@ -1,35 +1,55 @@
...
 import com.kickstarter.viewmodels.FeatureFlagViewHolderViewModel
-import kotlinx.android.synthetic.internal.item_feature_flag.view.*
 
-class FeatureFlagViewHolder(val view: View) : KSViewHolder(view) {
+class FeatureFlagViewHolder(val binding: ItemFeatureFlagBinding, val delegate: Delegate) : KSViewHolder(binding.root) {
 
     private val vm: FeatureFlagViewHolderViewModel.ViewModel = FeatureFlagViewHolderViewModel.ViewModel(environment())
 
+    interface Delegate {
+        fun featureOptionToggle(featureName: String, isEnabled: Boolean)
```
jgsamudio, line 13: Can you document this function? Also should this Delegate be inside the class or on the outside?

app/src/internal/java/com/kickstarter/ui/adapters/FeatureFlagsAdapter.kt
```diff
@@ -4,22 +4,24 @@ import android.view.LayoutInflater
...
 import com.kickstarter.R
 import com.kickstarter.databinding.EmptyViewBinding
+import com.kickstarter.databinding.ItemFeatureFlagBinding
+import com.kickstarter.model.FeatureFlagsModel
 import com.kickstarter.ui.viewholders.EmptyViewHolder
 import com.kickstarter.ui.viewholders.FeatureFlagViewHolder
 import com.kickstarter.ui.viewholders.KSViewHolder
 
-class FeatureFlagsAdapter : KSAdapter() {
+class FeatureFlagsAdapter(private val delegate: FeatureFlagViewHolder.Delegate) : KSAdapter() {
```
jgsamudio, line 13: @Arkariang Do we normally use the namespace to identify an interface like this or is the standard `FeatureFlagViewHolderDelegate` instead?

app/src/internal/java/com/kickstarter/viewmodels/FeatureFlagsViewModel.kt
```diff
@@ -1,23 +1,24 @@
...
+import com.kickstarter.libs.utils.extensions.SEGMENT_ENABLED
+import com.kickstarter.model.FeatureFlagsModel
 import com.kickstarter.ui.activities.FeatureFlagsActivity
 import rx.Observable
 import rx.subjects.BehaviorSubject
 
 interface FeatureFlagsViewModel {
-    interface Inputs
+    interface Inputs {
+        fun updateSegmentFlag(flag: Boolean)
```
jgsamudio, line 13: Can this function be documented as well?

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
```diff
@@ -49,11 +53,23 @@ public void onActivityResumed(final @NonNull Activity activity) {
       // Facebook: logs 'install' and 'app activate' App Events.
       AppEventsLogger.activateApp(activity.getApplication());
 
+      //to get the current config for segment feature when app resume from background
+      this.config.observable().subscribe (c ->{
+              this.currentConfig = c;
```
jgsamudio, line 58:
```suggestion
         this.currentConfig = c;
```

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
```diff
@@ -49,11 +53,23 @@ public void onActivityResumed(final @NonNull Activity activity) {
...
+      });
+
       // Refresh the config file
       this.client.config()
         .compose(Transformers.pipeApiErrorsTo(this::handleConfigApiError))
         .compose(Transformers.neverError())
-        .subscribe(c -> this.config.config(c));
+        .subscribe(c -> {
+          if(c.features().containsKey(SEGMENT_ENABLED)){
+             boolean isChecked= currentConfig.features().get(SEGMENT_ENABLED);
```
jgsamudio, line 67:
```suggestion
             boolean isChecked = currentConfig.features().get(SEGMENT_ENABLED);
```

app/src/internal/java/com/kickstarter/ui/adapters/FeatureFlagsAdapter.kt
Arkariang: On android for this cases should probably be something like `FeatureFlagViewHolder.ViewListener`, but throughout  the app there is mostly called `Delegate` - the android app was starter by some iOS folks - so I see where it comes from see this one as example [ProjectCardViewHolder](https://github.com/kickstarter/android-oss/blob/fa7a573b7f5e0879ca903f394b99660ab85a7383/app/src/main/java/com/kickstarter/ui/viewholders/ProjectCardViewHolder.kt#L35) .
But yeah I agree, a naming more android-related would be nice

app/src/internal/java/com/kickstarter/ui/viewholders/FeatureFlagViewHolder.kt
Arkariang: it belongs internally to the ViewHolder

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
```diff
@@ -44,24 +49,34 @@ public void onActivityStarted(final @NonNull Activity activity) {
...
+
       // Refresh the config file
       this.client.config()
-        .compose(Transformers.pipeApiErrorsTo(this::handleConfigApiError))
-        .compose(Transformers.neverError())
-        .subscribe(c -> this.config.config(c));
+              .compose(Transformers.pipeApiErrorsTo(this::handleConfigApiError))
+              .compose(Transformers.neverError())
+              .subscribe(c -> {
+                if (Objects.requireNonNull(c.features()).containsKey(ConfigFeatureName.SEGMENT_ENABLED.getConfigFeatureName())) {
```
Arkariang, line 66: There is no way for us to override the feature flags without having to touch any actual production code? what would happen when we have more feature flags?

hadia: the issue that code called every time app resumed  so if the app in the background and come to foreground the config is ovverrided

hadia: Done ✅

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
hadia: Done

app/src/internal/java/com/kickstarter/viewmodels/FeatureFlagsViewModel.kt
hadia: Done

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
```diff
@@ -44,24 +51,36 @@ public void onActivityStarted(final @NonNull Activity activity) {
...
       this.client.config()
-        .compose(Transformers.pipeApiErrorsTo(this::handleConfigApiError))
-        .compose(Transformers.neverError())
-        .subscribe(c -> this.config.config(c));
+              .compose(Transformers.pipeApiErrorsTo(this::handleConfigApiError))
+              .compose(Transformers.neverError())
+              .subscribe(c -> {
+                if (this.build.isDebug() || Build.isInternal()) {
+                  if (Objects.requireNonNull(c.features()).containsKey(ConfigFeatureName.SEGMENT_ENABLED.getConfigFeatureName())) {
+                    final boolean isChecked = Objects.requireNonNull(this.currentConfig.features()).get(ConfigFeatureName.SEGMENT_ENABLED.getConfigFeatureName());
```
Arkariang, line 70: I'm gonna suggest a few things in here, but I'll be happy to Pair program with you if it's not clear :) 
1 - when a feature flag is checked storage a user preference with the name of the feature flag and the checked/uncheck value
2 - remove the lines 58 to 61
3 - for overriding the value instead of checking the previous config file in memory just check the user preference for that feature flag value

You can even create a Config extension method that will check for each feature flag if there is an associated user preference value.

Arkariang (changes requested):

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
hadia: Done

app/src/main/java/com/kickstarter/libs/utils/extensions/ConfigExtension.kt
```diff
@@ -54,19 +59,40 @@ fun Config.enabledFeatureFlags(): JSONArray? {
...
+}
 
 /**
- * Internal name for "android_segment"
- * @{link https://github.com/kickstarter/kickstarter/blob/e9c61ea9f1e4817bb64560db7c32f3b9704cdc60/config/features.yml#L219}
+ * set the saved feature flags in to config feature object
  */
-const val SEGMENT_ENABLED = "android_segment"
+
+fun Config.setUserFeatureFlagsPrefWithFeatureFlag(
```
Arkariang, line 79: Love this!

app/src/internal/java/com/kickstarter/model/FeatureFlagsModel.kt
```diff
@@ -0,0 +1,7 @@
+package com.kickstarter.model
+
+data class FeatureFlagsModel(
```
Arkariang, line 3: Nice!

Arkariang (approved):
Lgtm! great work @hadia.
I tested it locally, now when you set the Segment Feature flag off on the Internal tools screen, we don't send any event. If you want to send events again you need to set that feature flag On again. I tried on both environments Production and Staging 👍

PR #1194, opened 2021-04-05
NT-1827: Side-by-side integration Segment-Braze

Arkariang:
# 📲 What

- Added `appboy-segment` dependency in order to complete the Side-By-Side integration, with this dependency added and configured on the segment client, we should be able to flow data from Segment to Braze.
- Some configuration has been done on Segment Dashboard to add braze dependency
- Some configuration has been done on Braze Dashboard to hold the data regarding our Segment implementation
- Updated firebase related dependencies as is mandatory for the next braze SDK integration

# 🤔 Why

# 🛠 How

# 👀 See
[Side-by-side integration](https://segment.com/docs/connections/destinations/catalog/braze/#android)
Screen Shot 2021-04-05 at 3 07 16 PM

|  |  |

# Story 📖

[NT-1827](https://kickstarter.atlassian.net/browse/NT-1827)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -245,7 +245,7 @@ dependencies {
     implementation 'com.perimeterx.sdk:msdk:1.13.1'
 
     // Firebase
-    implementation platform('com.google.firebase:firebase-bom:26.0.0')
+    implementation platform('com.google.firebase:firebase-bom:26.7.0')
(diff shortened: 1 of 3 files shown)


leighdouglas (approved):
🎉 LGTM!

PR #1201, opened 2021-04-09
EP-424: Page Viewed (search)

sunday-okpoluaefe:
# 📲 What

Android: Page Viewed (search)

# 🤔 Why

Datalake Event to Convert for Segment

When the user types a search term, the app currently makes a network call to fetch the results. When the results return from the API, call the Page Viewed event with the search term in the discover_search_term property with the correct total count in the discover_search_results_count

# 🛠 How

- Created method ``` trackSearchResultPageViewed ``` in ``` AnalyticEvents ```
```
/**
     * Sends data associated with the search results page viewed to segment.
     * @param discoveryParams: DiscoveryParams
     * @param count: Int
     * @param sort: DiscoveryParams.Sort
     */
    fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams, count: Int, sort: DiscoveryParams.Sort) {
        val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams).toMutableMap()
        props[CONTEXT_PAGE.contextName] = SEARCH.contextName
        props[DISCOVER_SORT.contextName] = sort.name.toLowerCase(Locale.ROOT)
        discoveryParams.term()?.let { props["discover_search_term"] = it }
        props["discover_search_results_count"] = count
        client.track(PAGE_VIEWED.eventName, props)
    }

```

- Called method from ``` SearchViewModel ``` on search results loaded from server
``` 
    params
          .compose(takePairWhen(this.discoverEnvelope))
          .compose(combineLatestPair(pageCount))
          .compose(bindToLifecycle())
          .filter(it -> ObjectUtils.isNotNull(it.first.first.term())
                  && StringExt.isPresent(it.first.first.term())
                  && it.first.first.sort() != defaultSort
                  && IntegerUtils.intValueOrZero(it.second) == 1)
          .distinct()
          .subscribe(it -> {
            this.lake.trackSearchResultPageViewed(defaultParams, it.first.second.stats().count(), defaultSort);
          });
```

- Added unit tests to segment event and properties 
``` SegmentTest ```
```
 @Test
    fun testSearchResultPageViewed_Properties() {
        val user = user()
        val client = client(user)
        client.eventNames.subscribe(this.segmentTrack)
        client.eventProperties.subscribe(this.propertiesTest)
        client.identifiedId.subscribe(this.segmentIdentify)
        val segment = AnalyticEvents(listOf(client))

        val params = DiscoveryParams
            .builder()
            .term("test")
            .sort(DiscoveryParams.Sort.POPULAR)
            .staffPicks(true)
            .build()

        segment.trackSearchResultPageViewed(params, 200, DiscoveryParams.Sort.POPULAR)

        assertSessionProperties(user)
        assertContextProperties()
        assertUserProperties(false)

        val expectedProperties = propertiesTest.value
        assertEquals("test", expectedProperties["discover_search_term"])
        assertEquals(200, expectedProperties["discover_search_results_count"])
        assertEquals(false, expectedProperties["discover_everything"])
        assertEquals(true, expectedProperties["discover_pwl"])
        assertEquals(false, expectedProperties["discover_recommended"])
        assertEquals("recommended_popular", expectedProperties["discover_ref_tag"])
        assertEquals(false, expectedProperties["discover_social"])
        assertEquals("popular", expectedProperties["discover_sort"])
        assertNull(expectedProperties["discover_subcategory_id"])
        assertNull(expectedProperties["discover_subcategory_name"])
        assertEquals(null, expectedProperties["discover_tag"])
        assertEquals(false, expectedProperties["discover_watched"])

        this.segmentTrack.assertValue(EventName.PAGE_VIEWED.eventName)
    }

```

``` SearchViewModelTest ```
```
this.lakeTest.assertValues("Search Button Clicked", EventName.CTA_CLICKED.getEventName(), "Search Page Viewed",  EventName.PAGE_VIEWED.getEventName());
```

# 👀 See

Screenshot 2021-04-09 at 03 34 30

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

- Click on the search button from the global nav
- Type a search string into the search box, you should see the new event registered in segment console

Screenshot 2021-04-09 at 03 36 47

# Story 📖

https://kickstarter.atlassian.net/browse/EP-424

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -853,6 +853,12 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track(CTA_CLICKED.eventName, props)
     }
 
+    /**
+     * Sends data associated with the search results page viewed to segment.
+     * @param discoveryParams: DiscoveryParams
+     * @param count: Int
+     * @param sort: DiscoveryParams.Sort
+     */
     fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams, count: Int, sort: DiscoveryParams.Sort) {
         val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams).toMutableMap()
         props[CONTEXT_PAGE.contextName] = SEARCH.contextName
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -859,8 +859,8 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
      * @param count: Int
      * @param sort: DiscoveryParams.Sort
      */
-    fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams, count: Int) {
-        val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams).toMutableMap()
+    fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams, count: Int, sort: DiscoveryParams.Sort) {
+        val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams, sort).toMutableMap()
         props[CONTEXT_PAGE.contextName] = SEARCH.contextName
         discoveryParams.term()?.let { props["discover_search_term"] = it }
         props["discover_search_results_count"] = count
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -854,9 +854,13 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track(CTA_CLICKED.eventName, props)
     }
 
-    fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams) {
-        val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams)
-        client.track(SEARCH_PAGE_VIEWED, props)
+    fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams, count: Int, sort: DiscoveryParams.Sort) {
+        val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams).toMutableMap()
+        props[CONTEXT_PAGE.contextName] = SEARCH.contextName
+        props[DISCOVER_SORT.contextName] = sort.name.toLowerCase(Locale.ROOT)
+        discoveryParams.term()?.let { props["discover_search_term"] = it }
+        props["discover_search_results_count"] = count
+        client.track(PAGE_VIEWED.eventName, props)
     }
 
     fun trackSearchPageViewed(discoveryParams: DiscoveryParams) {
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -859,10 +859,9 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
      * @param count: Int
      * @param sort: DiscoveryParams.Sort
      */
-    fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams, count: Int, sort: DiscoveryParams.Sort) {
+    fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams, count: Int) {
         val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams).toMutableMap()
         props[CONTEXT_PAGE.contextName] = SEARCH.contextName
-        props[DISCOVER_SORT.contextName] = sort.name.toLowerCase(Locale.ROOT)
         discoveryParams.term()?.let { props["discover_search_term"] = it }
         props["discover_search_results_count"] = count
         client.track(PAGE_VIEWED.eventName, props)
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -854,6 +854,11 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         client.track(CTA_CLICKED.eventName, props)
     }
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt [hunk 1]
```diff
@@ -853,6 +853,21 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
...
+     * Sends data associated with the search results page viewed to segment.
+     * @param discoveryParams: DiscoveryParams
+     * @param count: Int
+     * @param sort: DiscoveryParams.Sort
+     */
+    fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams, count: Int, sort: DiscoveryParams.Sort) {
+        val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams).toMutableMap()
+        props[CONTEXT_PAGE.contextName] = SEARCH.contextName
+        props[DISCOVER_SORT.contextName] = sort.name.toLowerCase(Locale.ROOT)
+        discoveryParams.term()?.let { props["discover_search_term"] = it }
```
leighdouglas, line 866: Do the specs say that if the search term is null, we don't send the `discover_search_term` property at all? Otherwise we should send an empty string.

app/src/main/java/com/kickstarter/viewmodels/SearchViewModel.java
```diff
@@ -135,12 +143,22 @@ public ViewModel(final @NonNull Environment environment) {
...
 
       params
-        .compose(takePairWhen(pageCount))
-        .filter(paramsAndPageCount -> paramsAndPageCount.first.sort() != defaultSort && IntegerUtils.intValueOrZero(paramsAndPageCount.second) == 1)
-        .map(paramsAndPageCount -> paramsAndPageCount.first)
-        .observeOn(Schedulers.io())
-        .compose(bindToLifecycle())
-        .subscribe(this.lake::trackSearchResultsLoaded);
+          .compose(takePairWhen(pageCount))
+          .compose(takePairWhen(this.discoverEnvelope))
```
leighdouglas, line 147: Instead of using `.takePairWhen` twice and creating a pair within a pair, consider using `Observable.zip` for all three streams and utilizing kotlin's `Triple` object. For example:

```     
      Observable.zip(
              params,
              pageCount,
              this.discoverEnvelop,
              (params, pageCount, discoverEnvelop) -> new Triple<DiscoveryParams, Integer, DiscoverEnvelop>(params, pageCount, discoverEnvelop));
```

This is the verbose example, but the lambda can be replaced with a method reference as long as you explicitly define what the parameter types of the `Triple` are before you create the method reference, otherwise the compiler won't know what type of items are in the triple and you won't have access to any of their class methods. I think it looks a bit cleaner and more readable. Inevitably, it should look something like this:

```Observable.zip(params, pageCount, this.discoverEnvelop, Triple::new)```

File: app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt [hunk 3]
```diff
@@ -853,6 +853,21 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
...
+    /**
+     * Sends data associated with the search results page viewed to segment.
+     * @param discoveryParams: DiscoveryParams
+     * @param count: Int
+     * @param sort: DiscoveryParams.Sort
+     */
+    fun trackSearchResultPageViewed(discoveryParams: DiscoveryParams, count: Int, sort: DiscoveryParams.Sort) {
+        val props = AnalyticEventsUtils.discoveryParamsProperties(discoveryParams).toMutableMap()
+        props[CONTEXT_PAGE.contextName] = SEARCH.contextName
+        props[DISCOVER_SORT.contextName] = sort.name.toLowerCase(Locale.ROOT)
```
leighdouglas, line 865: So if you look at the discoveryParamsProperties method, it is handling adding the `discover_sort` property automatically. You can either pass it in explicitly, or it will default to whatever value is in `discoveryParams.sort`. If you include this property in this method, it will be sending `discover_sort` twice.

Arkariang:
is ok to go on my end, as soon as @leighdouglas validates the properties side of things. Great work @sunday-okpoluaefe 👍 !

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
sunday-okpoluaefe: This is because the getter ``` term() ``` returns an optional

leighdouglas: I understand that, I'm just trying to figure out if this should be moved into the `discoveryParamsProperties` method in the `analyticEventsUtils` class, or if it should be isolated to this method. Also wondering if we should send `discover_search_term = ""` if the `term()` is empty.

sunday-okpoluaefe:
```
 .filter(it -> ObjectUtils.isNotNull(it.first.first.term())
                  && StringExt.isPresent(it.first.first.term())
                  && it.first.first.sort() != defaultSort
                  && IntegerUtils.intValueOrZero(it.second) == 1)
```

this ensures ``` term() ``` is never empty

leighdouglas (approved):
LGTM! 🚀 🎉

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
leighdouglas: Offline chat: these two properties are unique to this event, and therefore shouldn't be moved to the `discoverProperties()` function in the `AnalyticEventsUtils` class.

app/src/main/java/com/kickstarter/viewmodels/SearchViewModel.java
leighdouglas: Closing this because we plan on creating a `combineLatestTriple()` transformer in the near future that would solve this.

app/src/test/java/com/kickstarter/viewmodels/SearchViewModelTest.java
```diff
@@ -42,6 +42,7 @@ private void setUpEnvironment(final @NonNull Environment environment) {
     this.vm.outputs.searchProjects().map(ps -> ps.size() > 0).subscribe(this.searchProjectsPresent);
   }
 
+
```
leighdouglas, line 45: nit: remove whitespace

PR #1203, opened 2021-04-12
EP-463 Android: QA 4/6 - User Properties Fixes

leighdouglas:
# 📲 What

- Update `user_created_projects_count` to be a value equal to the sum of `createdProjectsCount` and `draftProjectsCount` from the `User` payload
- Update `user_launched_projects_count` to be a value equal to the `createdProjectsCount` from the `User` payload

# 🤔 Why

Segment Integration

# 🛠 How

- Added the `draftProjectsCount` to the `User` struct, which is available from the api
- Changed the above user props accordingly. 

# 📋 QA

- Login to the app with an account that has a value for both `createdProjectsCount`, `draftProjectsCount`, and `starredProjectsCount`
- Fire any event
- `user_created_projects_count` prop should be equal to `createdProjectsCount` and `draftProjectsCount`
- `user_launched_projects_count` prop should only be equal to `createdProjectsCount`
Screen Shot 2021-04-12 at 12 33 27 PM

# Story 📖

[EP-463: Android QA 4/6 - User Properties Fixes](https://kickstarter.atlassian.net/browse/EP-463)


Arkariang (approved):
lgtm! great work here @leighdouglas

PR #1205, opened 2021-04-13
NT-1821: Switch prelaunch display logic to use display_prelaunch field from v1 API

leighdouglas:
# 📲 What

Switch logic tied to `prelaunchActivated()` from the `User` payload to rely on `displayPrelaunch()` instead. 

# 🤔 Why

Currently, native is using the field `prelaunch_activated` to determine the display logic of prelaunch deeplinks. We want to change this logic to check `display_prelaunch` instead. 

The backend will be updating `prelaunch_activated` to actually return the correct `project.prelaunch_activated` value. This is also keeping parity with web as they are using this field. Because of this, any instances using prelaunch_activated for display logic will need to be updated to use display_prelaunch instead.

Relevant backend PR: https://github.com/kickstarter/kickstarter/pull/21408

# 🛠 How

- Update the `User` struct to grab `displayPrelaunch` value available from the API
- Switch instances of `prelaunchActivated()` to `displayPrelaunch()` so logic relys on this boolean
- Update the `ProjectFactory` method `prelaunchProject` to be set the `displayPrelaunch` value 

# 📋 QA

This change requires smoke and regression testing. 
- Find a project in prelaunch
- Open the deeplink on your mobile device
- Make sure the deeplink opens the project in the web browser on your device

# Story 📖

[NT-1821: Switch prelaunch display logic to use display_prelaunch field from v1 API](https://kickstarter.atlassian.net/browse/NT-1821)

### DIFF
--- a/app/src/main/java/com/kickstarter/mock/factories/ProjectFactory.java
+++ b/app/src/main/java/com/kickstarter/mock/factories/ProjectFactory.java
@@ -483,7 +483,7 @@ public final class ProjectFactory {
 
     return project()
       .toBuilder()
-      .prelaunchActivated(true)
+      .displayPrelaunch(true)
       .urls(Project.Urls.builder().web(web).build())
       .build();
   }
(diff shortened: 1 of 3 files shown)


sunday-okpoluaefe (approved):
lgtm!

PR #1208, opened 2021-04-16
EP-483: Checkout Amount Total Property

sunday-okpoluaefe:
# 📲 What

Android QA: Checkout Amount Total Property

# 🤔 Why

The total amount of my checkout was $602 AUD. Web correctly converted that to 459.51, but Android is sending 555.3922764062881. I don’t know how this is being calculated on Android, but it should be the total pledge amount at checkout, converted to USD from the project currency. 

Should also be a decimal, not a float.

# 🛠 How

- Changed 
``` fun CheckoutData.totalAmount(usdRate: Float) = this.totalAmount() * usdRate ```
 to 
``` fun CheckoutData.totalAmount(usdRate: Float) = this.amount() * usdRate ```

This is because ``` this.amount() ``` is already ``` pledgeAmount + addOns + shippingAmount + bonus ```. 

- Rounded up to 2dp decimal 
```
 put("amount", checkoutData.amount().round()) 
 put("amount_total_usd", checkoutData.totalAmount(project.staticUsdRate()).round())
```

- Adjusted segment tests cases

# 👀 See

checkoutAmount

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖

https://kickstarter.atlassian.net/browse/EP-483

### DIFF
--- a/app/src/main/java/com/kickstarter/ApplicationModule.java
+++ b/app/src/main/java/com/kickstarter/ApplicationModule.java
@@ -351,7 +351,7 @@ public class ApplicationModule {
   @NonNull
   static HttpLoggingInterceptor provideHttpLoggingInterceptor() {
     final HttpLoggingInterceptor interceptor = new HttpLoggingInterceptor();
-    interceptor.setLevel(HttpLoggingInterceptor.Level.HEADERS);
+    interceptor.setLevel(HttpLoggingInterceptor.Level.BODY);
     return interceptor;
   }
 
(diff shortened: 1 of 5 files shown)


Arkariang (commented):
lgtm! as soon as @leighdouglas review it is good to go!

leighdouglas (approved):
LGTM! 🚀

leighdouglas:
Oh, one small thing, if you could attach a screenshot of the segment output that shows the correct value for `checkout_amount_total_usd`, that would be great :)

PR #1210, opened 2021-04-16
EP-462: Android QA 4/6 - Session Properties Fixes

leighdouglas:
# 📲 What

The format with which we were sending `session_variants_optimizely` was incorrect. We want to send an array of strings formatted with the name of the optimizely experiment and the experiment group the user is in:
```
"suggested_no_reward_amount[control]"
```
# 🤔 Why

Segment integration/QA feedback

# 📋 QA

- Open kickstarter and make sure you are logged in
- Fire any event that include session properties
- Verify on segment that the raw json is formatted correctly (you have to be whitelisted in an optimizely experiment): 
Screen Shot 2021-04-16 at 4 31 18 PM

# Story 📖

[EP-462: Android QA 4/6 - Session Properties Fixes](https://kickstarter.atlassian.net/browse/EP-462)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/ExperimentsClientType.kt
+++ b/app/src/main/java/com/kickstarter/libs/ExperimentsClientType.kt
@@ -57,7 +57,7 @@ interface ExperimentsClientType {
      * "session_variants_optimizely" : [ "Experiment1":"variant1",
      *                                   "Experiment2":"varian2"]
      */
-    fun getTrackingProperties(): Map<String, Array<Map<String, String>>>
+    fun getTrackingProperties(): Map<String, Array<String>>
 }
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/libs/ExperimentsClientType.kt
```diff
@@ -57,7 +57,7 @@ interface ExperimentsClientType {
      * "session_variants_optimizely" : [ "Experiment1":"variant1",
```
Arkariang, line 57: Update also the comments if the expected format has change to "Experiment[variant]" please :)

Arkariang (approved):
lgtm! just update the documentation on `getTrackingProperties()` with the new format specification and is good to go 👍

PR #1217, opened 2021-04-21
EP-488: Android QA Project Subcategory & Tags

leighdouglas:
# 📲 What

This ticket addresses the following properties:

- `project_subcatergory` was previously sending the localized name of the subcategory. This PR updates this to send the subcategory name in english
- `project_tags` were not previously being sent, so the project model and project properties have been updated to send this value

# 🤔 Why

Segment Integration/QA Feedback

# 🛠 How

For `project_subcategory`, the v1 api was updated to include a new field, analyticsName, with the category model, which includes the non-localized subcategory name (backend PR: https://github.com/kickstarter/kickstarter/pull/21281)
For `project_tags`, the backend exposed a field to return the public tags for a given project. We updated the project struct and now are sending the tags as a string delineated by the comma. 

# 📋 QA

`project_subcategory`:
- Change your device language to something other than English
- Open the app 
- Tap the hamburger menu
- Tap on a category, for example "Games"
- Tap on a subcategory, for example "Tabletop Games" 
- Tap on a project card
- In segment, you should see the subcategory name in english, not localized

`project_tags`:
- Open the app 
- Tap on a project card on the discover screen
- In segment, a list of project tags as a string delineated by a comma
📣 Note that project tags are not currently available for all events, as the payload for projects we receive from the backend differs depending on where we are in the app 📣 

Screen Shot 2021-04-21 at 2 36 24 PM

# Story 📖

[EP-488: Android QA Project Subcategory & Tags](https://kickstarter.atlassian.net/browse/EP-488)


sunday-okpoluaefe (approved):
LGTM! Good job

PR #1218, opened 2021-04-21
EP-500: Android QA Discover Category Name Localized

leighdouglas:
# 📲 What

The discover category and subcategory names were coming back localized to the device language. For analytics purposes, we need to send these values in English at all times. 

# 🤔 Why

Segment integration/QA Feedback

# 🛠 How

A new field was added to the category struct called `analyticsName` that will always return the english string of the category or subcategory. Simply updating the utility method to grab the value from `analyticsName` instead of the `name` field resolves this issue. 

# 📋 QA

Example of testing using the `discover_filter` `cta_click` event
- Change device language to anything other than English, for example, Spanish
- Open the app
- Tap the hamburger menu
- Select a category, such as "Arte"
- From this drop down menu, select a subcategory, such as "Cerámica"
- Segment should show a `cta_clicked` event with `context_cta = "discover_filter"`, and both the `discover_category_name` and `discover_subcategory_name` values should both be in english:
Screen Shot 2021-04-21 at 4 49 13 PM

# Story 📖

[EP-500: Android QA Discover Category Name Localized](https://kickstarter.atlassian.net/browse/EP-500)


sunday-okpoluaefe (approved):
LGTM!

PR #1219, opened 2021-04-21
NT-1852: Segment initialization moved to SegmentTrackingClient

Arkariang:
# 📲 What

- Segment initialization moved to SegmentTrackingClient.kt

# 🤔 Why
- Isolate the initialization of the dependency in the Client we created for that purpose
- Now the lifecycle events we send to segment are also behind the feature flag.

# 🛠 How

- The initialization for segment has been moved to `SegmentTrackingClient` in that way we've completely isolated the dependency on it's client.
- `SegmentTrackingClient` has the method `isEnabled()` to check if the feature flag is active

# 📋 QA
- Check that if you have the feature flag off, no lifecycle event is sent to Segment.

# Story 📖

[NT-1852](https://kickstarter.atlassian.net/browse/NT-1852)


leighdouglas (approved):
Looks great! 🎉

PR #1220, opened 2021-04-22
EP-509: Android QA: project_rewards_count

sunday-okpoluaefe:
# 📲 What

Android QA: project_rewards_count

- Number of rewards listed on the project at the time the event fired (this doesn't include add-ons)

# 🤔 Why
- Invalid project rewards count
- 
For the project Orbitkey ID Card Holder System, the value for project_rewards_count is 8, but it should be 7. We shouldn't count "no reward" as a reward.

# 🛠 How

- Changed 
```  
           project.rewards()?.let { rewards ->
                put("rewards_count", rewards.size)
            }
```
 to 
``` 
          project.rewards()?.let { a ->
                val rewards = a.filter { isReward(it) }
                put("rewards_count", rewards.size)
            }
```

- Adjusted segment tests cases

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

- On the search page. Search ```Orbitkey ID Card Holder System ```. Click on the project to view project details.
- A ``` Page Viewed ``` or ``` CTA Clicked ``` event should be registered on the segment with the ``` project_rewards_count ``` set to ``` 7 ``` as expected

# Story 📖

https://kickstarter.atlassian.net/browse/EP-509


Arkariang (approved):
lgtm!

PR #1227, opened 2021-04-28
NT-1867: eliminate unnecessary identify calls

Arkariang:
# 📲 What

- Removed from `TrackingClient.kt` internal variables `config` and `loggedInUser` those two properties have become class properties on `TrackingClientType.kt`
- `LakeTrackingClient.kt ` overrides  it's own `config` and `loggedInUser`
- `SegmentTrackingClient.kt ` overrides  it's own `config` and `loggedInUser`
- For `SegmentTrackingClient.kt` we update the current user only when some of the traits has change with method `updateUserAndCheckTraits`
- Avoid race condition for initializing Segment SDK, call initialize only when we have configuration available
- Added `isInitialized` state for Segment SDK
- Modified several tests

# 👀 See
- No visible changes on this PR. It's all around behaviour.

# 📋 QA
- Make sure when you come from background `identify` it's not called
- Make sure if you change some of your Opt-in preferences `identify` it's called
- Make sure if you change your password or email you are logged out of the app and no `identify` is called
- Make sure every time you log in `identify` is called 
- Race condition has been solved, all the events should get to Segment Dashboard

# Story 📖

[NT-1867](https://kickstarter.atlassian.net/browse/NT-1867)


hadia (approved):
prefect👌👌

PR #1232, opened 2021-05-02
EP-466:  context_page = other

sunday-okpoluaefe:
# 📲 What

Android: context_page = other

# 🤔 Why

Automatically send the value context_page = other for all events that occur on pages we haven’t named yet? This is particularly relevant for some Phase 2 events.

# 🛠 How

- Refactored ``` client.track ``` method to set ``` context_page ``` to ``` other ``` for properties without ``` context_page ``` set

```
fun track(eventName: String, additionalProperties: Map<String, Any>) {
            val properties = hashMapOf<String, Any>()
            properties.putAll(additionalProperties)
            if (!properties.containsKey(CONTEXT_PAGE.contextName))
                properties[CONTEXT_PAGE.contextName] = EventContextValues.ContextPageName.OTHER.contextName

            clients.forEach { client ->
                client?.track(eventName, properties)
            }
        }

```

- Added unit tests to test for properties  with and without ``` context_page ```

```

 @Test
    fun testProperties_WithContextPage() {
        val client = client(null)
        client.eventNames.subscribe(this.segmentTrack)
        client.eventProperties.subscribe(this.propertiesTest)
        val segment = AnalyticEvents(listOf(client))

        segment.trackSignUpPageViewed()

        val expectedProperties = propertiesTest.value
        assertEquals(SIGN_UP.contextName, expectedProperties[CONTEXT_PAGE.contextName])

        assertSessionProperties(null)
        assertContextProperties()
    }

```

```

 /*
     context_page should be set to other for properties without context page
     */
    @Test
    fun testProperties_WithoutContextPage() {
        val client = client(null)
        client.eventNames.subscribe(this.segmentTrack)
        client.eventProperties.subscribe(this.propertiesTest)
        val segment = AnalyticEvents(listOf(client))

        segment.trackSignUpPageViewedForTest()

        val expectedProperties = propertiesTest.value
        assertEquals(EventContextValues.ContextPageName.OTHER.contextName, expectedProperties[CONTEXT_PAGE.contextName])

        assertSessionProperties(null)
        assertContextProperties()
    }

```

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

- On the discovery page, click on any project.
- Events name ``` Project Page Viewed ``` with ``` context_page = other ```  and ``` CTA Clicked ``` with ``` context_page = discover  ``` should be registered on segment console.

other

context_page

# Story 📖

https://kickstarter.atlassian.net/browse/EP-466

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -1323,7 +1323,7 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         fun track(eventName: String, additionalProperties: Map<String, Any>) {
             val properties = hashMapOf<String, Any>()
             properties.putAll(additionalProperties)
-            if(!properties.containsKey(CONTEXT_PAGE.contextName))
+            if (!properties.containsKey(CONTEXT_PAGE.contextName))
                 properties[CONTEXT_PAGE.contextName] = EventContextValues.ContextPageName.OTHER.contextName
 
             clients.forEach { client ->
--- a/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
@@ -335,6 +335,14 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
     }
 
     /**
+     * Sends data to the client when the signup page is viewed.
+     */
+    fun trackSignUpPageViewedForTest() {
+        val props: HashMap<String, Any> = HashMap()
+        client.track(PAGE_VIEWED.eventName, props)
+    }
+
+    /**
      * Sends data to the client when the login/sign-up page is viewed.
      */
     fun trackLoginOrSignUpPagedViewed() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
```diff
@@ -1313,8 +1321,13 @@ class AnalyticEvents(trackingClients: List<TrackingClientType?>) {
         }
 
         fun track(eventName: String, additionalProperties: Map<String, Any>) {
+            val properties = hashMapOf<String, Any>()
```
Arkariang, line 1324: This method should just call track for each one of the clients, no properties added here please

leighdouglas: To add to this -> I would move this check into the `TrackingClientType.kt` class.

leighdouglas:
Looking through QA, it seems as though we are testing this by comparing datalake events to segment events. Is this the case because we currently don't have any events when an unknown/unnamed `context_page` name?

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
sunday-okpoluaefe: Noted. thanks

sunday-okpoluaefe: Noticed there are no events without context_page on android

PR #1235, opened 2021-05-03
NT-1872: Middleware + persist users traits

Arkariang:
# 📲 What
- Added a middleware for Braze destination
- [NT-1867 remaning point](https://kickstarter.atlassian.net/browse/NT-1867), kill the app and launch it again will not fire an `identify` call

# 🤔 Why
- We call `identify` for segment with all the user traits
- The middleware allow us to call `identify` and build an specific payload that will be sent to Braze
- The payload we build for braze contains ONLY the traits that has change from the previous identify call
- The traits are persisted at the end of the chain calls after calling identify.

# 👀 See
Screen Shot 2021-05-03 at 2 49 16 PM

|  |  |

# 📋 QA
- The first time you log in in the app all traits will be send to both destinations.
- On next log in events Segment will get all the traits while Braze will get traits just if some trait has changed.
- Go to the notification preferences screen change some opt-in option -> On segment you will see an identify call with all the traits, while in Braze you will see only the trait for the subscription changed.
- Go to background and come back, no identify call should be made.
- Kill the app and launch it again, no identify call should be made -> [NT-1867](https://kickstarter.atlassian.net/browse/NT-1867)
- Change email, log in again -> all traits on segment, none on braze
- Change password, log in again  -> all traits on segment, none on braze

# Story 📖

[NT-1872](https://kickstarter.atlassian.net/browse/NT-1872)
[NT-1867](https://kickstarter.atlassian.net/browse/NT-1867)


leighdouglas (approved):
LGTM! 🚀

hadia:
image
test great work 👌🎉

PR #1237, opened 2021-05-07
[no-jira]: Update segment version + flush queue with 1 event on staging

Arkariang:
# 📲 What

- Update segment version + flush queue with 1 event on staging

# 🤔 Why

- We are trying to solve some timestamp issues going on. See -> https://github.com/segmentio/analytics-android/issues/757 and it seems it might be solve on the latest Segment SDK version.

# 🛠 How

- Updated version from 4.9.0 to 4.9.4
- Added `.flushQueueSize(1)` when instanciating Segment client on testing builds.

# 👀 See
- no user facing changes
- 
|  |  |

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -313,7 +313,7 @@ dependencies {
     implementation "org.jsoup:jsoup:1.12.1"
 
     // Analytics Segment-Braze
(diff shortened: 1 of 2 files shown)


hadia (approved):
👌🎉

leighdouglas (approved):
LGTM! 🚀

PR #1240, opened 2021-05-10
NT-1735: Target API 30 🚀 

leighdouglas:
# 📲 What

Upgrading to Android API 30 requires a few updates for our codebase to be compatible. I have added a queries element to the Android manifest file for the new package visibility requirements in API 30. Also updated tests to reflect new formatting changes.

- Added a new queries element to android manifest.
- Updated the compileSdkVersion and targetSdkVersion to 30
- Upgraded robolectric dependency to 4.4
- Retrieving a project object from an intent now checks for nullability
- Set KSCurrency with a mock current config to the mock environment object used for testing view models to pass tests failing on the external release variant
- Disabled pausing and unpausing the shadow looper in the `ObserveForUITransformerTest` suite (more on that below)
- Amended the jacoco config struct to exclude coverage for internal jdk classes -> this is necessary since we are using JDK 1.8 (more on that below)
- Updated the codecov config struct to be informational and not pass/fail PRs
- Updated failing tests with new formatting to Date and Number strings:
```
Dec 17, 2015 6:35 PM -> Dec 17, 2015, 6:35 PM
17 déc. 2015 18:35:05 -> 17 déc. 2015 à 18:35:05
100 $ -> now has a non-breaking space
```

# 🛠 How

More context on `ObserveForUITransformerTest` -> Robolectric 4.4 was a massive update that included changes to their shadow looper. These changes are causing mysterious test failures during the `tearDown()` that happens after the test is run, specifically it is throwing an `java.lang.UnsupportedOperationException: main looper cannot be unpaused` exception. After exhausting a few solutions from their documentation (they have a whole article about it here-> http://robolectric.org/blog/2019/06/04/paused-looper/), and since all of this code is about 5 years old, @jgsamudio and I decided to remove these lines since the tests are passing without them. Perhaps robolectric is not performing these operations for us under the hood. 

For the jacoco config to exclude generating coverage for jvm internal classes, this is a known issue and related to our target jvm being 1.8. See -> https://github.com/gradle/gradle/issues/5184

# 👀 See

No UI changes

# 📋 QA

Since these are all internal changes and not user facing, I would test this by opening the app and making sure nothing is crashing. 

# Story 📖

[[SPIKE] Fix Test Environment for Android 11 Migration](https://kickstarter.atlassian.net/browse/NT-1735)
[[Spike]Android 11: Target API 30](https://kickstarter.atlassian.net/browse/NT-1626)

### DIFF
--- a/app/src/test/java/com/kickstarter/KSRobolectrictTestCase.kt
+++ b/app/src/test/java/com/kickstarter/KSRobolectrictTestCase.kt
@@ -3,7 +3,13 @@ package com.kickstarter
 import android.app.Application
 import android.content.Context
 import androidx.test.core.app.ApplicationProvider
-import com.kickstarter.libs.*
+import com.kickstarter.libs.AnalyticEvents
+import com.kickstarter.libs.Environment
+import com.kickstarter.libs.KSCurrency
+import com.kickstarter.libs.KSString
+import com.kickstarter.libs.MockCurrentUser
+import com.kickstarter.libs.MockTrackingClient
+import com.kickstarter.libs.TrackingClientType
 import com.kickstarter.libs.utils.Secrets
 import com.kickstarter.mock.MockCurrentConfig
 import com.kickstarter.mock.MockExperimentsClientType
@@ -46,7 +52,7 @@ abstract class KSRobolectricTestCase : TestCase() {
             .build()
 
         val config = ConfigFactory.config().toBuilder()
-                .build()
+            .build()
 
         mockCurrentConfig.config(config)
 
--- a/app/src/test/java/com/kickstarter/KSRobolectrictTestCase.kt
+++ b/app/src/test/java/com/kickstarter/KSRobolectrictTestCase.kt
@@ -3,15 +3,11 @@ package com.kickstarter
 import android.app.Application
 import android.content.Context
 import androidx.test.core.app.ApplicationProvider
-import com.kickstarter.libs.AnalyticEvents
-import com.kickstarter.libs.Environment
-import com.kickstarter.libs.KSString
-import com.kickstarter.libs.MockCurrentUser
-import com.kickstarter.libs.MockTrackingClient
-import com.kickstarter.libs.TrackingClientType
+import com.kickstarter.libs.*
 import com.kickstarter.libs.utils.Secrets
 import com.kickstarter.mock.MockCurrentConfig
 import com.kickstarter.mock.MockExperimentsClientType
+import com.kickstarter.mock.factories.ConfigFactory
 import com.kickstarter.mock.services.MockApiClient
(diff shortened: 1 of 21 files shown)


app/src/test/java/com/kickstarter/KSRobolectrictTestCase.kt
```diff
@@ -49,7 +51,13 @@ abstract class KSRobolectricTestCase : TestCase() {
             .applicationModule(TestApplicationModule(application()))
             .build()
 
+        val config = ConfigFactory.config().toBuilder()
```
Arkariang, line 54: nice!

app/src/test/java/com/kickstarter/libs/rx/transformers/ObserveForUITransformerTest.java
```diff
@@ -23,13 +23,13 @@
   @Before
   public void setUp() {
     RxAndroidPlugins.getInstance().reset();
-    ShadowLooper.pauseMainLooper();
+    //ShadowLooper.pauseLooper(Looper.myLooper());
```
Arkariang, line 26: delete this two lines please

Arkariang (approved):
Great work here @leighdouglas!

app/src/test/java/com/kickstarter/viewmodels/BackingFragmentViewModelTest.kt
```diff
@@ -94,6 +96,7 @@ class BackingFragmentViewModelTest : KSRobolectricTestCase() {
         this.vm.outputs.projectDataAndAddOns().subscribe(this.listAddOns)
         this.vm.outputs.bonusSupport().map { it.toString() }.subscribe(this.bonusAmount)
         this.vm.outputs.deliveryDisclaimerSectionIsGone().subscribe(this.disclaimerSectionIsGone)
+        Shadows.shadowOf(Looper.getMainLooper()).idle()
```
Arkariang, line 99: I think these `Shadows.shadowOf(Looper.getMainLooper()).idle()` are no longer necessary as you found the root cause if blocking problem

app/src/test/java/com/kickstarter/libs/rx/transformers/ObserveForUITransformerTest.java
leighdouglas: Done 👍🏽

app/src/test/java/com/kickstarter/viewmodels/BackingFragmentViewModelTest.kt
leighdouglas: Done, and everything passed CI! 🎉 🚀

PR #1241, opened 2021-05-11
NT-1871: InAppMessage listener + DeepLink

Arkariang:
# 📲 What
- Added a new Deeplink with the format `[ksr|https]://staging.kickstarter.com/settings/notify_mobile_of_marketing_update/true`
- When that deeplink is recognized we update the user property `notifyMobileOfMarketingUpdate`, we perform a networking call to the self endpoint and refresh the current user with the updated property.
- `DeepLinkActivity.java` styles have been updated to be a transparent activity invisible to the user.
- Added custom handler for inAppMessage `InAppCustomListener`, when a message is sent to the app we first check if the user is logged in and the feature flag is active before displaying the message on the screen

# 🤔 Why
- The InAppMessages on Braze dashboard will be configured to use the new Deeplink 
Screen Shot 2021-05-11 at 3 51 38 PM
- We are using InAppMessages to opt-in the user for the new Push Notifications for marketing

# 👀 See
- settings -> notifications for marketing push notifications off
opt-in-off
- Get InAppMessage on the App -> Press positive button
inappmessage
- The deep link has been executed go back to settings-> notifications check that is now active the opt-in for marketing updates
opt-in-on

- See the Segment identify call 
identifyCall

- See Braze custom attribute for the user 
brazeUser

|  |  |

# 📋 QA
- Log in in the app, go to Settings -> Notifications, make sure you have disabled the option for marking push notifications.
- Go to Braze Dashboard and send an InAppMessage to your user.
- Go back to the App (in case you don't see the message put the app in background and open it again) the message should be displayed.
- Press the positive button from the message.
- Go back to Settings -> Notifications, you'll see this time active the option for marketing push notifications.
- You will also see an `identify` call on Segment dashboard as the opt-in preference for the user has changed.

# Story 📖

[NT-1871](https://kickstarter.atlassian.net/browse/NT-1871)


leighdouglas (approved):
LGTM! 🎉

app/src/main/java/com/kickstarter/services/KSUri.java
```diff
@@ -14,6 +14,10 @@
 public final class KSUri {
```
leighdouglas, line 14: We should migrate this to kotlin at some point.

app/src/main/java/com/kickstarter/viewmodels/DeepLinkViewModel.java
```diff
@@ -33,11 +40,16 @@
 
     /** Emits when we should start the {@link com.kickstarter.ui.activities.ProjectActivity} with pledge sheet expanded. */
     Observable<Uri> startProjectActivityForCheckout();
+
+    /** Emits when we should finish the current activity */
+    Observable<Void> finisDeeplinkActivity();
```
leighdouglas, line 45: nit: typo in method name, missing an "h", should be `finishDeeplinkActivity`

app/src/main/java/com/kickstarter/services/KSUri.java
Arkariang: nw, I have another branch with the kotlin migration ;) it will be another separeted PR

leighdouglas: Awesome! 😃

app/src/main/java/com/kickstarter/viewmodels/DeepLinkViewModel.java
Arkariang: Fixed!

PR #1243, opened 2021-05-12
NT-1916: Feature flag comment threading

Arkariang:
# 📲 What

- Added feature Flag on Optimizely
- Added new attributes to optimizaly calls
 `"session_app_release_version_number" = 2131` version (2.13.1) in numbers,
 `"session_app_release_version_code" = BuildConfig.VERSION_CODE`

# 🤔 Why
- Feature Flag for the new Treaded Coment Screen
- Adding the new attributes will provide more flexibility around versioning for feature flags

# 👀 See
Screen Shot 2021-05-12 at 2 51 36 PM
- Feature Flag On
FeatureFlagOn
- Feature Flag Off
FeatureFlagOff

|  |  |

# 📋 QA
- Enable the feature flag, when you press the comments row in the project page no new screen will be presented, that screen is work in progress
- Disable the feature flag, the old comments screen will be presented 

# Story 📖

[NT-1916](https://kickstarter.atlassian.net/browse/NT-1916)

PR #1247, opened 2021-05-13
NT-1920: Setup Skeleton Activity and ViewModel for Root Comments

leighdouglas:
# 📲 What

This is the start of feature work on threaded comments. In this PR, we are:
- Deprecating legacy comment classes, tests, v1 api models and layouts
- Creating skeleton classes for `CommentsActivity` and `CommentsViewModel` for root comments. 

# 🤔 Why

We will not be using the old UI and legacy code for root comments as we are migrating to a new UI flow, and therefore these classes are no longer needed. Likewise, we will be moving on from the v1 api calls for comments and shifting to graphQL, and thus creating new models for these queries. 

# 👀 See

These are skeleton classes and have no UI as of yet. 

# 📋 QA

Test that fetching, scrolling, posting and updating comments is working as it should. 

# Story 📖

[NT-1920: Setup Skeleton Activity and ViewModel for Root Comments](https://kickstarter.atlassian.net/browse/NT-1920)


app/src/main/java/com/kickstarter/ui/activities/UpdateActivity.kt
```diff
@@ -141,7 +141,7 @@ class UpdateActivity : BaseActivity<UpdateViewModel.ViewModel?>(), KSWebView.Del
     }
 
     private fun startCommentsActivity(update: Update) {
-        val intent = Intent(this, CommentsActivity::class.java)
```
Arkariang, line 144: Well done! 
I'll add the feature flag check on the `UpdatesActivity` as well, I wasn't aware you can also open comments from the "UpdatesActivity" 👍

leighdouglas: Thanks! Great callout!

app/src/main/AndroidManifest.xml
```diff
@@ -50,7 +50,7 @@
             android:name=".ui.activities.EditProfileActivity"
             android:theme="@style/SettingsActivity" />
         <activity
```
Arkariang, line 52: I do not see the new Activity declared here, is gonna be done in next Pr's @leighdouglas ?

leighdouglas: Oh good callout, forgot about the manifest 😅 i'll update now 👍🏽

app/src/main/AndroidManifest.xml
```diff
@@ -46,11 +46,14 @@
...
             android:name=".ui.activities.ChangePasswordActivity"
             android:theme="@style/SettingsActivity" />
+        <activity
```
Arkariang, line 49: According to this comment you left on previous PR https://github.com/kickstarter/android-oss/pull/1243#discussion_r632628450 the activity should be called `RootCommentsActivity`

Arkariang (approved):
Small comment about the Activity name, apart form that all good 👍

app/src/main/AndroidManifest.xml
leighdouglas: Oh, iOS hasnt changed their naming yet but i think the idea is that they will do it in the future? I can update the classes to be prefixed with "Root" now if we want to get it over with sooner than later.

PR #1249, opened 2021-05-17
NT-1916: Feature flag on `Activity` screen

Arkariang:
# 📲 What
- Added feature flag check on `UpdateActivity`

# 📋 QA
- if you get any update with the format : "https://kck.str/projects/param/param/posts/id/comments" in case the feature flag is active it will open the new `CommentsActivity`

# Story 📖

[NT-1916](https://kickstarter.atlassian.net/browse/NT-1916)


hadia (approved):
👌

PR #1253, opened 2021-05-20
NT-1906: GraphQL API to post a comment

sunday-okpoluaefe:
# 📲 What

GraphQL API to post a comment

# 🤔 Why

Add a GraphQL API that allows a new comment to be posted

# 🛠 How

- Create graphql  ```CreateCommentMutation ```
```
mutation CreateComment($body: String!, $commentableId: ID!, $parentId: ID, $clientMutationId: String) {
  createComment(input: {
    body: $body,
    commentableId: $commentableId,
    parentId: $parentId,
    clientMutationId: $clientMutationId
  }) {
    comment {
      ...comment
    }
  }
}

```

- Added ``` override fun createComment(comment: PostCommentInput): Observable<Comment> ``` to apolloClient
```
 override fun createComment(comment: PostCommentInput): Observable<Comment> {
        return Observable.defer {
            val ps = PublishSubject.create<Comment>()
            this.service.mutate(
                CreateCommentMutation.builder()
                    .parentId(comment.parentId)
                    .commentableId(comment.commentableId() ?: "")
                    .clientMutationId(comment.clientMutationId)
                    .body(comment.body ?: "")
                    .build()
            )
                .enqueue(object : ApolloCall.Callback<CreateCommentMutation.Data>() {
                    override fun onFailure(exception: ApolloException) {
                        ps.onError(exception)
                    }

                    override fun onResponse(response: Response<CreateCommentMutation.Data>) {
                        if (response.hasErrors()) {
                            ps.onError(java.lang.Exception(response.errors?.first()?.message))
                        }

                        ps.onNext(
                            createCommentObject(response.data?.createComment()?.comment()?.fragments()?.comment())
                        )
                        ps.onCompleted()
                    }
                })
            return@defer ps
        }
    }

```

- Created Mock data 
```
override fun createComment(comment: PostCommentInput): Observable<Comment> {
        return Observable.just(CommentFactory.comment())
    }

```

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

- Paste this snippet in the ``` ProjectViewModel ``` init block
```
   val comment = backedProject
                .switchMap {
                    this.apolloClient.createComment(
                        PostCommentData.builder()
                        .body("Hello!! Just a comment")
                        .project(it)
                        .build()
                    )
                }
                .filter {
                    ObjectUtils.isNotNull(it)
                }
                .share()

            backedProject
                .compose(combineLatestPair(comment))
                .compose(bindToLifecycle())
                .subscribe{
                    val com = it
                }

```

- Add break points as shown

Screen Shot 2021-05-21 at 5 20 50 PM

- Run in debug mode
- Go to your list of backed projects
- Select any project from the list, a comment is automatically posted and you should get a break as shown

Screen Shot 2021-05-21 at 5 19 52 PM

# Story 📖

https://kickstarter.atlassian.net/browse/NT-1906


Arkariang:
@sunday-okpoluaefe  will be nice if the QA section on the PR's specified how to test/QA the work you've done

app/src/main/java/com/kickstarter/models/PostCommentInput.kt
```diff
@@ -0,0 +1,44 @@
+package com.kickstarter.models
```
Arkariang, line 1: Move this model into `com.kickstarter.services.mutations` you can see in there most of the data models used on graphql mutations.

-also a small note,  no need for the model that goes in a mutation to be parcelable. We send information, so need to share that model on an Intent.

app/src/main/java/com/kickstarter/models/PostCommentInput.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import android.os.Parcelable
+import com.kickstarter.libs.qualifiers.AutoGson
+import com.kickstarter.libs.utils.encodeRelayId
+import kotlinx.android.parcel.Parcelize
+
+@AutoGson
+@Parcelize
+class PostCommentInput(
```
Arkariang, line 10: - Rename it to `PostCommentData`

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
```diff
@@ -271,6 +268,37 @@ class KSApolloClient(val service: ApolloClient) : ApolloClientType {
...
+                        ps.onError(exception)
+                    }
+
+                    override fun onResponse(response: Response<CreateCommentMutation.Data>) {
+                        if (response.hasErrors()) {
+                            ps.onError(java.lang.Exception(response.errors?.first()?.message))
+                        }
+
+                        ps.onNext(
+                            createCommentObject(response.data?.createComment()?.comment()?.fragments()?.comment())
```
Arkariang, line 293: Why do we need to create a comment from the mutation response? 
- Unless specified otherwise,  when using the post comment mutation we just need to check no error is returned from in that operation.
Take a look to other mutations as `com.kickstarter.services.KSApolloClient#updateUserEmail` or `com.kickstarter.services.KSApolloClient#updateUserPassword`

app/src/main/java/com/kickstarter/models/PostCommentInput.kt
sunday-okpoluaefe: done

app/src/main/java/com/kickstarter/models/PostCommentInput.kt
sunday-okpoluaefe: done

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
sunday-okpoluaefe: I felt it will make sense to have a copy of what you posted. just in case we want to update the list without doing a full refresh

app/src/main/java/com/kickstarter/libs/utils/SharedFunctions.kt
```diff
@@ -61,3 +65,20 @@ fun checkoutProperties(
         .shippingAmount(shippingAmount)
         .build()
 }
+
+fun <T : Relay> encodeRelayId(relay: T): String {
```
Arkariang, line 69: encodeRelay is used on `Long` types, if you wanna extract it please do so in an Extension function for the `Long` class not into this class. 

This  `SharedFunctions` has been accepted as an exception for a concrete use case we had where the type using the functions was either `Backing` or `CheckoutData`, but should not be extended further that that.

app/src/main/java/com/kickstarter/libs/utils/SharedFunctions.kt
```diff
@@ -61,3 +65,20 @@ fun checkoutProperties(
...
+    val classSimpleName = relay.javaClass.simpleName.replaceFirst("AutoParcel_", "")
+    val id = relay.id()
+    return Base64Utils.encodeUrlSafe(("$classSimpleName-$id").toByteArray(Charset.defaultCharset()))
+}
+
+fun decodeRelayId(encodedRelayId: String?): Long? {
```
Arkariang, line 75: Same here, `decodeRelayId` can be an extension to the `String` class in case you need to extract this method

Arkariang (changes requested):
Hey @sunday-okpoluaefe a few comments left,  but looking good! Please ping me on slack in case you have any doubt

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
Arkariang: Perfect, it makes sense for now! then please add some documentation on the method `createCommentObject` explaining your reasoning, something like you just explained.
`make a copy of what you posted. just in case we want to update the list without doing a full refresh` so we can know in the future the reasoning behind it, to keep it or deleted in the future in case no necessary

app/src/main/java/com/kickstarter/services/mutations/PostCommentData.kt
```diff
@@ -0,0 +1,10 @@
+package com.kickstarter.services.mutations
```
Arkariang, line 1: 👍

Arkariang (approved):
As soon as the documentation is updated, feel free to merge, great work @sunday-okpoluaefe

PR #1254, opened 2021-05-21
NT-1853: Removed Feature Flag for Braze

Arkariang:
# 📲 What

- Remove Feature Flag for Braze

# 🛠 How
- Config object removed as dependency on `BrazeClient`,  `InAppCustomListener`, `InAppCustomListenerHandler`
- Any check related with the Feature Flag now is always true or directly removed
- Updated/removed unit tests around feature flag being enabled or disabled

# 📋 QA
- Make sure Braze client is always initialized when loading the app, you can take a look in logcat a search for `AppBoy` logs

# Story 📖

[NT-1853](https://kickstarter.atlassian.net/browse/NT-1853)

### DIFF
--- a/app/src/main/java/com/kickstarter/ApplicationModule.java
+++ b/app/src/main/java/com/kickstarter/ApplicationModule.java
@@ -173,8 +173,8 @@ public class ApplicationModule {
   @Provides
   @Nonnull
   @Singleton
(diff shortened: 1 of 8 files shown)


hadia (approved):
👌👌

PR #1259, opened 2021-05-25
[NT-1808, NT1837] Blank Screens for Activity and Backed Projects

jgsamudio:
# 📲 What

When the applications resumes, it refreshes the config file and checks if there is a 401 (Unauthorized). A 401 is given when the auth token passed is no longer valid. The application then forces a logout and displays the discovery activity. We've noticed however that when the user is logged in however the access token is null, the app will not be able to load authenticated endpoints such as Activity and Backed projects. 

In order to prevent this case, an additional check is made when the user is refreshed. When a null access token is passed to the config endpoint, it does not return a 401. This is because the endpoint is does not require the user to be logged in. Logged out users can refresh their config file. If the user is logged in and the access token is null, the app will also force a logout similar to the 401 error. 

We also added firebase events when there is a forced logout to see how often this is occurring within the application. 

# 🤔 Why

User's have reported through CS and app store reviews that they see blank screens in certain areas of the app. More information can be found in the tickets below.

# 📋 QA

To check the original issue, from master: 
1. Log into the app 
2. Go to `CurrentUser.java` and change the function `getAccessToken()` to return null
3. Rerun the app
4. Notice that the app does not sign out. Accessing activity, messages or profile will result in a blank screen

To check the fixed implementation, from this branch:
1. 1. Log into the app 
2. Go to `CurrentUser.java` and change the function `getAccessToken()` to return null
3. Rerun the app
4. Notice that the app signs the user out

To verify the Firebase events follow the guide [here](https://firebase.google.com/docs/analytics/debugview)
The event name is `force_logout` with params: location as either `access_token_null` or `config_api_error`

# Story 📖

[NT-1808](https://kickstarter.atlassian.net/browse/NT-1808)
[NT-1837](https://kickstarter.atlassian.net/browse/NT-1837)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
@@ -96,6 +96,7 @@ public final class ApplicationLifecycleUtil implements Application.ActivityLifec
     ApplicationUtils.startNewDiscoveryActivity(this.application);
     Bundle params = new Bundle();
     params.putString("location", context);
+
     FirebaseAnalytics.getInstance(this.application).logEvent("force_logout", params);
   }
 
--- a/app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
@@ -34,10 +34,15 @@ public final class ApplicationLifecycleUtil implements Application.ActivityLifec
 
   private final KSApplication application;
   private boolean isInBackground = true;
+  private boolean isLoggedIn = false;
 
   public ApplicationLifecycleUtil(final @NonNull KSApplication application) {
     this.application = application;
     application.component().inject(this);
+
+    this.currentUser.isLoggedIn().subscribe(isLoggedIn -> {
+      this.isLoggedIn = isLoggedIn;
+    });
   }
 
   @Override
@@ -94,7 +99,7 @@ public final class ApplicationLifecycleUtil implements Application.ActivityLifec
   private void forceLogout(final @NonNull String context) {
     this.logout.execute();
     ApplicationUtils.startNewDiscoveryActivity(this.application);
-    Bundle params = new Bundle();
+    final Bundle params = new Bundle();
     params.putString("location", context);
 
     FirebaseAnalytics.getInstance(this.application).logEvent("force_logout", params);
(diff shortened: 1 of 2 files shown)


Arkariang:
for running locally the Java checkstyle you can execute in the terminal `./gradlew :app:checkstyle`

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
```diff
@@ -53,40 +54,68 @@ public void onActivityResumed(final @NonNull Activity activity) {
...
+  }
+
+  /**
+   * Forces the current user session to be logged out.
+   */
+  private void forceLogout(final @NonNull String context) {
+    this.logout.execute();
+    ApplicationUtils.startNewDiscoveryActivity(this.application);
+    Bundle params = new Bundle();
+    params.putString("location", context);
```
Arkariang, line 98: Are we adding this param because of an Analytic event?

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
```diff
@@ -53,40 +54,68 @@ public void onActivityResumed(final @NonNull Activity activity) {
...
+
+    FirebaseAnalytics.getInstance(this.application).logEvent("force_logout", params);
+  }
+
+  /**
+   * Refreshes the user object if there is not a user logged in with a non-null access token.
+   */
+  private void refreshUser() {
+    final String accessToken = this.currentUser.getAccessToken();
+    final Boolean isloggedIn = ObjectUtils.isNotNull(this.currentUser.getUser());
```
Arkariang, line 108: `getUser` is deprecated in case you wanna check if there is a logged user you should subscribe to the currentUser observer

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
jgsamudio: Yes! I added the param to know when we are forcing the logout

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.java
jgsamudio: Updated to use subscribe instead of getUser

Arkariang (approved):
Lgtm! thanks for all the investigation work here @jgsamudio much appreciated :heart:

PR #1264, opened 2021-05-27
NT-1909: View comments with infinite scrolling

hadia:
# 📲 What

1. Users can view all root comments
2. More comments are loaded when a user is close to the bottom of the view

# 🤔 Why

Load all comments thread

# 🛠 How

Create  Pagination handler for list view  checking if we got to page end or we got to the next page empty

# 👀 See
https://www.loom.com/share/1b95bbe7bf294e52bb8fdb6d7a927f96

# 📋 QA
open comments and scroll to load more

# Story 📖

https://kickstarter.atlassian.net/browse/NT-1909


app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
```diff
@@ -18,7 +20,24 @@ class CommentsActivity :
     BaseActivity<CommentsViewModel.ViewModel>(),
     CommentsAdapter.Delegate {
     private lateinit var binding: ActivityCommentsLayoutBinding
-    private val adapter = CommentsAdapter(this)
+    private val adapter = CommentsAdapter(
+        this,
+        object : DiffUtil.ItemCallback<Any>() {
```
Arkariang, line 25: Awesome!

Arkariang: Could we extract it to a public class re-usable in other recyclerview adapters ?

app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
```diff
@@ -27,10 +46,48 @@ class CommentsActivity :
         setContentView(view)
         binding.commentsRecyclerView.adapter = adapter
 
+        val loadMoreListView = PaginationHandler(
```
Arkariang, line 49: From line 46 to line 61 could it be extracted to a function 
`fun setUpPagination` or any other name.

app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
```diff
@@ -27,10 +46,48 @@ class CommentsActivity :
...
+            .subscribe {
+                loadMoreListView.refreshing(it)
+            }
+
         viewModel.outputs.commentsList()
             .compose(bindToLifecycle())
             .observeOn(AndroidSchedulers.mainThread())
-            .subscribe { it?.let { comments -> adapter.takeData(comments) } }
+            .subscribe {
+                it?.let { comments -> adapter.takeData(comments) }
```
Arkariang, line 89: instead of `it?.let {...}` could we filter on the ViewModel the output?, so we can unit test that logic.
On the `Oncreate` methods for actvities/fragments if posible only have the subscription and no more logic unless related with UI

app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
```diff
@@ -18,7 +20,24 @@ class CommentsActivity :
...
+            }
+
+            override fun areContentsTheSame(oldItem: Any, newItem: Any): Boolean {
+                return threadsAreTheSame(oldItem, newItem)
+            }
+
+            private fun threadsAreTheSame(oldItem: Any, newItem: Any): Boolean {
+                val oldThread = oldItem as Comment
+                val newThread = newItem as Comment
+                return oldThread.id() == newThread.id()
```
Arkariang, line 37: if the equals operator if properly implemented by doing `oldObj == newObj ` should be enough, so we can make the DiffUtil generic

app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
hadia: done

app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
hadia: Done

app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
hadia: Done

app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
hadia: Done

Arkariang (approved):
LGTM! great work @hadia ! I tested it on my environment and looks great

PR #1267, opened 2021-05-30
NT-1955: UI – Comment Card with threads

sunday-okpoluaefe:
# 📲 What

UI – Comment Card with threads

# 🤔 Why

- In order to support threads we need to add the missing controls to our Comment
- Also, we need to support yet another variation of the Comment Card design, to display replies

# 🛠 How

- Added padding end and margin guidelines to threadactivitylayout
```
<androidx.constraintlayout.widget.Guideline
            android:id="@+id/left_guideline"
            app:layout_constraintGuide_begin="@dimen/grid_5"
            android:orientation="vertical"
            android:layout_width="wrap_content"
            android:layout_height="wrap_content"/>


        <androidx.recyclerview.widget.RecyclerView
            android:id="@+id/comment_replies_recycler_view"
            app:layout_constraintStart_toEndOf="@id/left_guideline"
            android:layout_width="0dp"
            android:paddingEnd="@dimen/grid_3"
            android:background="@color/kds_white"
            app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
            app:layout_constraintTop_toBottomOf="@id/comments_card_view"
            tools:layout_editor_absoluteX="0dp"
            tools:listitem="@layout/comment_card" />
```

# 👀 See

Screen Shot 2021-05-30 at 12 37 55 PM

Screen Shot 2021-05-30 at 2 19 06 PM

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

- Select any project. Go to comments.  You should see the updated comment card
- Click ``` View replies ``` button. You should be taken to the threads screen

# Story 📖

https://kickstarter.atlassian.net/browse/NT-1955


app/src/main/java/com/kickstarter/ui/viewholders/CommentCardViewHolder.kt
```diff
@@ -7,6 +7,7 @@ import com.kickstarter.libs.utils.DateTimeUtils
 import com.kickstarter.models.Comment
 import com.kickstarter.ui.views.OnCommentCardClickedListener
 import com.kickstarter.viewmodels.CommentsViewHolderViewModel
+import kotlinx.android.synthetic.main.comment_card.view.*
```
Arkariang, line 10: Wildcard imports are forbidden by the ktlinter, I'll check why the linter has not marked this as an error, but please change the wildcard import

app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
```diff
@@ -110,10 +119,22 @@ class CommentCard @JvmOverloads constructor(
         binding.commentBody.setTextColor(ContextCompat.getColor(context, commentBodyTextColor))
     }
 
+    fun setCommentReplies(replies: Int) {
+        /*
```
Arkariang, line 123: nit: move the documentation on top of the function name

app/src/main/java/com/kickstarter/ui/views/CommentReplyCard.kt
```diff
@@ -0,0 +1,99 @@
...
+import android.view.LayoutInflater
+import androidx.constraintlayout.widget.ConstraintLayout
+import androidx.core.content.ContextCompat
+import androidx.core.content.withStyledAttributes
+import androidx.core.view.isVisible
+import com.kickstarter.R
+import com.kickstarter.databinding.CommentReplyCardBinding
+import com.kickstarter.ui.extensions.loadCircleImage
+
+class CommentReplyCard @JvmOverloads constructor(
```
Arkariang, line 14: Do we really need to create another Component for the reply? it seems a lot of duplicated code just in order to add a margin on the left. 
We could re-use the already existing `CommentCard` and use these states or add a new one:
Screen Shot 2021-06-01 at 12 09 39 PM

  to add a (guideline and left constraint to it in case is a reply) or simply a marginleft, not really strong opinions in how to build that part of the UI

Arkariang (changes requested):
- The reply button works like a charm 
ViewReplies
- But for the Replies UI I do strongly think no need to have a new component please @sunday-okpoluaefe let me know if there is any concern around it or if I'm missing some context here

app/src/main/java/com/kickstarter/ui/views/CommentReplyCard.kt
sunday-okpoluaefe: thought about that. There are several logics around the card state. Did not want to complicate it. I will try reusing the existing commentCard.

Arkariang: For sure that simpler than adding the margin on the card UI component, great suggestion 👍

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
```diff
@@ -22,6 +22,9 @@ interface CommentsViewHolderViewModel {
         /** Call when the user clicks reply to comment */
         fun onReplyButtonClicked()
 
+        /** Call when the user clicks view replies to comment */
+        fun onViewRepliesButtonClicked()
```
Arkariang, line 26: There should be tests covering the new inputs/outputs

app/src/main/res/layout/comment_card.xml
```diff
@@ -37,27 +38,58 @@
         android:id="@+id/comment_body"
         style="@style/CommentCardBody"
         android:text="@string/A_successfully_funded_project_will_collect_your_pledge_in_its_native_currency"
-        android:visibility="gone"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="@id/avatar"
         app:layout_constraintTop_toBottomOf="@id/avatar" />
 
+    <androidx.appcompat.widget.LinearLayoutCompat
```
Arkariang, line 45: We decided a while a ago any new UI should use Constraint layout, we are trying to keep flat hierarchies as much as possible.
You can keep the 
`<androidx.appcompat.widget.AppCompatTextView
                android:id="@+id/replies"
                android:layout_weight="1"
                android:text="@string/view_replies"
                android:layout_width="wrap_content"
                android:layout_height="wrap_content"/>

            <androidx.appcompat.widget.AppCompatImageView
                android:src="@drawable/ic_mask"
                android:layout_width="wrap_content"
                android:layout_height="wrap_content"/>`

and use constaints to position it please

Arkariang (changes requested):

app/src/main/res/layout/comment_card.xml
leighdouglas: +1 to using constraint layout and flattening the view hierarcy

sunday-okpoluaefe: done

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
sunday-okpoluaefe: done

app/src/main/res/layout/comment_card.xml
```diff
@@ -101,6 +119,13 @@
         android:visibility="gone"
         app:constraint_referenced_ids="reply_button,flag_button" />
 
+    <androidx.constraintlayout.widget.Group
```
Arkariang, line 122: Nice touch!

Arkariang (approved):
Great work @sunday-okpoluaefe !

PR #1269, opened 2021-06-01
NT-1977: UX – Not logged in / not backer for Comment Card

leighdouglas:
# 📲 What

We want to disable the reply/comment card action buttons in different user states i.e. logged in or out, backing or not backing, etc.
I also renamed a few variables to be clearer and extracted some functionality into individual methods, for readability and all that good stuff 😄 

# 🤔 Why

- The reply option should only show if the user is logged in and has backed the project, or if they are the project creator. 
- A user should not be able to reply to projects if they are not logged in, or if they are logged in but have not backed the project or they are not the creator. 

# 🛠 How

- Created a new `CommentCardData` model that holds the `comment`, the `project` the comment is from, and the state of the card (for future use when replying, errored, etc).
- We then transform the list of comments returned in the network call to a list of `CommentCardData`, which is passed to the adapter and eventually the `CommentsViewHolderViewModel`
- The `CommentsViewHolderViewModel` view model then performs a check when initialized that returns true or false depending on if the user should see the reply button. This boolean is passed to the `CommentViewModel`, which toggles the visibility of the action group on the comment card, which holds the reply button as well as the flag comment button.

# 👀 See

No reply/action buttons: Logged out and logged in and NOT backing, respectively:
Screenshot_1622590680 Screenshot_1622591284 

Reply/Action buttons visible: Logged in and backing
Screenshot_1622591300

# 📋 QA
Make sure you have a project that you are not backing, a project that you are backing, and a project that you created, and all projects must have comments. Then test the different states listed below.

⛔ = no reply or action buttons
✅  = reply and action buttons visible

These are the different states to test:
Logged out = ⛔
Logged in and not backing = ⛔ 
Logged in and backing = ✅ 
Logged in and creator = ✅ 

# Story 📖

[NT-1977: UX – Not logged in / not backer for Comment Card](https://kickstarter.atlassian.net/browse/NT-1977)

### DIFF
--- a/app/src/main/java/com/kickstarter/mock/factories/CommentFactory.kt
+++ b/app/src/main/java/com/kickstarter/mock/factories/CommentFactory.kt
@@ -73,7 +73,7 @@ class CommentFactory {
                     )
                     .build(),
                 0,
-                    ProjectFactory.project().toBuilder().creator(UserFactory.creator().toBuilder().id(278438049).build()).build()
+                ProjectFactory.project().toBuilder().creator(UserFactory.creator().toBuilder().id(278438049).build()).build()
 
             )
         }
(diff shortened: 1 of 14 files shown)


Arkariang (approved):
Lgtm!

PR #1271, opened 2021-06-03
NT-1910: UX – Not logged in / Not backer for Comment Composer

leighdouglas:
# 📲 What

This change handles the different states of the comment composer in relation to the user state i.e. logged out, not backing, etc.
Also had to update the UI to make the separator disappear as it was still appearing when the composer was gone. 

# 🤔 Why

There are three states -> `ENABLED, DISABLED, GONE`
- `GONE` -> The comment composer should be completely gone if the user is not logged in.
- `ENABLED` -> If the user is logged in and a backer, or is the creator, the comment composer is enabled.
- `DISABLED` -> If the user is logged in but not a backer, then the comment composer should be in the disabled state

# 🛠 How

Created an enum that holds these three states, and when the comment view model is initiated, it emits this state to the activity and the view responds accordingly. 

# 👀 See

Logged in: Backing and not backing
Screenshot_1622673686 Screenshot_1622673709

Logged out:
Screenshot_1622673778

# 📋 QA

Make sure you have a project that you are not backing, a project that you are backing, and a project that you created, and all projects must have comments. Test going to the comments directly from the project page, as well as from the updates page. Then test the different states listed below. 

🔴  = no comment composer
🟡  = comment composer disabled
🟢  = comment composer enabled

These are the different states to test:
Logged out = 🔴 
Logged in and not backing = 🟡 
Logged in and backing = 🟢 
Logged in and creator = 🟢 

# Story 📖

[NT-1910: UX – Not logged in / Not backer for Comment Composer](https://kickstarter.atlassian.net/browse/NT-1910)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
@@ -49,18 +49,18 @@ class CommentsActivity :
                 binding.commentComposer.setAvatarUrl(it)
             }
 
-        viewModel.outputs.enableCommentComposer()
+        viewModel.outputs.commentComposerStatus()
             .compose(bindToLifecycle())
             .observeOn(AndroidSchedulers.mainThread())
             .subscribe {
-                binding.commentComposer.showCommentComposerDisabledView(!it)
+                binding.commentComposer.setCommentComposerStatus(it)
             }
 
         viewModel.outputs.showCommentComposer()
             .compose(bindToLifecycle())
             .observeOn(AndroidSchedulers.mainThread())
             .subscribe {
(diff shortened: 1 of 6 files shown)


Arkariang (approved):
It looks great @leighdouglas ! 
- Small request that can be handle in another ticket as it was not already there, but waht you think about adding tests for the composer component? we can use as example the `CommentCardTest` https://github.com/kickstarter/android-oss/blob/master/app/src/test/java/com/kickstarter/ui/view/CommentCardTest.kt

leighdouglas:
@Arkariang I noticed that too and figured we could update in a subsequent ticket. I'll make a ticket for it :)

PR #1272, opened 2021-06-03
NT-1984: Feature flags for threads

Arkariang:
# 📲 What

- New feature flag added that `hides View replies/Reply buttons`
- The Flag button has been removed
- Feature flag added to optimizely

# 🛠 How
- `CommentCard.kt` will be aware of the feature flag state via a new output from the viewmodel `isCommentEnableThreads`
- All the visibility updates will happend internally on `CommentCard`

# 👀 See
- New optimizely Feature Flag
Screen Shot 2021-06-03 at 2 35 26 PM

- Backer with feature flag off: reply and view replies buttons are gone
https://user-images.githubusercontent.com/4083656/120713432-bc57fc80-c476-11eb-8e24-6ceb533c805b.mp4

- Backer with feature flag on: reply and view replies buttons are visible and active
https://user-images.githubusercontent.com/4083656/120713769-36888100-c477-11eb-8978-bee459714c44.mp4

- No Backer with feature flag off: view replies buttons is gone
https://user-images.githubusercontent.com/4083656/120714369-f5dd3780-c477-11eb-81d8-e289f1146a90.mp4

- No Backer with feature flag on: view replies buttons is visible and active
https://user-images.githubusercontent.com/4083656/120714568-39d03c80-c478-11eb-8589-1a89b3fe3cdf.mp4

|  |  |

# 📋 QA
- With the feature flag on, make sure as baker you can reply and see replies
- With the feature flag off, make sure as baker you cannot see the reply button nor the view replies button
- With the feature flag on, make sure as NO baker you can see the view replies button and it's active
- With the feature flag off, make sure as NO baker you cannot see the view replies button

# Story 📖

[NT-1984](https://kickstarter.atlassian.net/browse/NT-1984)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
+++ b/app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
@@ -135,8 +135,8 @@ class CommentCard @JvmOverloads constructor(
 
     private fun shouldShowReplyButton(cardCommentStatus: CommentCardStatus) =
         cardCommentStatus == CommentCardStatus.COMMENT_WITH_REPLIES ||
-        cardCommentStatus == CommentCardStatus.COMMENT_WITHOUT_REPLIES ||
-        cardCommentStatus == CommentCardStatus.FAILED_TO_SEND_COMMENT
+            cardCommentStatus == CommentCardStatus.COMMENT_WITHOUT_REPLIES ||
+            cardCommentStatus == CommentCardStatus.FAILED_TO_SEND_COMMENT
 
     /*
      * To display replies count
@@ -146,7 +146,7 @@ class CommentCard @JvmOverloads constructor(
         setViewRepliesVisibility(replies > 0)
     }
 
-    fun setViewRepliesVisibility(isViewRepliesVisible: Boolean){
+    fun setViewRepliesVisibility(isViewRepliesVisible: Boolean) {
         binding.replies.isVisible = isViewRepliesVisible && isCommentEnabledThreads
     }
 
--- a/app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
+++ b/app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
@@ -85,7 +85,7 @@ class CommentCard @JvmOverloads constructor(
                 setCommentUserName(it)
             }
 
-            getBoolean(R.styleable.CommentCardView_is_comment_action_group_visible, true)?.also {
+            getBoolean(R.styleable.CommentCardView_is_comment_reply_button_visible, true)?.also {
                 setReplyButtonVisibility(it)
             }
 
--- a/app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
+++ b/app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
@@ -104,7 +104,7 @@ class CommentCard @JvmOverloads constructor(
     }
 
     fun setReplyButtonVisibility(isGroupVisible: Boolean) {
-        binding.replyButton.isVisible = isGroupVisible && isCommentEnabledThreads
+        binding.replyButton.isVisible = isGroupVisible && this.isCommentEnabledThreads
     }
 
     fun setCommentCardStatus(cardCommentStatus: CommentCardStatus) {
@@ -122,7 +122,7 @@ class CommentCard @JvmOverloads constructor(
         }
 
         binding.retryButton.isVisible =
-            cardCommentStatus == CommentCardStatus.FAILED_TO_SEND_COMMENT && isCommentEnabledThreads
+            cardCommentStatus == CommentCardStatus.FAILED_TO_SEND_COMMENT
 
         val commentBodyTextColor = if (cardCommentStatus == CommentCardStatus.FAILED_TO_SEND_COMMENT) {
             R.color.soft_grey_disable
--- a/app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
+++ b/app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
@@ -24,6 +24,8 @@ class CommentCard @JvmOverloads constructor(
 
     private var onCommentCardClickedListener: OnCommentCardClickedListener? = null
 
+    private var isCommentEnabledThreads: Boolean = false
+
     init {
         obtainStyledAttributes(context, attrs, defStyleAttr)
 
@@ -31,10 +33,6 @@ class CommentCard @JvmOverloads constructor(
 
         binding.retryButton.setOnClickListener {
             onCommentCardClickedListener?.onRetryViewClicked(it)
-        }
-
-        binding.flagButton.setOnClickListener {
-            onCommentCardClickedListener?.onFlagButtonClicked(it)
         }
 
         binding.replies.setOnClickListener {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
```diff
@@ -129,20 +133,29 @@ class CommentCard @JvmOverloads constructor(
         binding.commentBody.setTextColor(ContextCompat.getColor(context, commentBodyTextColor))
     }
 
+    private fun shouldShowReplyButton(cardCommentStatus: CommentCardStatus) =
```
Arkariang, line 136: @hadia for the new states when dealing with errors, just add the new ones in this function. It should make solving conflicts easier

app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
```diff
@@ -129,20 +133,29 @@ class CommentCard @JvmOverloads constructor(
...
      * To display replies count
      * binding.replies.text = String.format("%s (%d)",resources.getString(R.string.view_replies), replies)
      */
     fun setCommentReplies(replies: Int) {
-        binding.replies.isVisible = replies > 0
+        setViewRepliesVisibility(replies > 0)
+    }
+
+    fun setViewRepliesVisibility(isViewRepliesVisible: Boolean) {
+        binding.replies.isVisible = isViewRepliesVisible && isCommentEnabledThreads
```
Arkariang, line 150: @hadia @leighdouglas @sunday-okpoluaefe if you need to update visibility of the View replies button of the reply button please call `setReplyButtonVisibility` do not set directly `binding.replies.isVisible` values

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
```diff
@@ -171,15 +181,32 @@ interface CommentsViewHolderViewModel {
                 .subscribe(this.flagComment)
         }
 
-        private fun isActionGroupVisible(commentCardData: CommentCardData, user: User?) =
-            commentCardData.project?.let {
-                it.isBacking || ProjectUtils.userIsCreator(it, user)
```
Arkariang, line 176: isCreator was affecting the flagging button that has been deleted, so `ProjectUtils.userIsCreator(it, user)` needed to be removed as well

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
```diff
@@ -171,15 +181,32 @@ interface CommentsViewHolderViewModel {
...
+         * Updates the status of the current comment card.
+         * Also updates the current state of the [OptimizelyFeature.Key.COMMENT_ENABLE_THREADS]
+         * everytime the state changes.
+         */
         private fun cardStatus(comment: Comment?) = when {
             comment?.deleted() ?: false -> CommentCardStatus.DELETED_COMMENT
             (comment?.repliesCount() ?: false != 0) -> CommentCardStatus.COMMENT_WITH_REPLIES
             else -> CommentCardStatus.COMMENT_WITHOUT_REPLIES
+        }.also {
+            this.isCommentEnableThreads.onNext(optimizely.isFeatureEnabled(OptimizelyFeature.Key.COMMENT_ENABLE_THREADS))
```
Arkariang, line 209: Please @hadia @sunday-okpoluaefe @leighdouglas note here the `also` block, as from this moment any update on the state of the comment card will also update the value for the feature flag

app/src/main/res/layout/comment_card.xml
```diff
@@ -67,16 +66,6 @@
         app:layout_constraintStart_toStartOf="@id/avatar"
         app:layout_constraintTop_toBottomOf="@id/replies" />
 
-    <androidx.appcompat.widget.AppCompatImageButton
-        android:id="@+id/flag_button"
```
Arkariang, line 71: flag button removed as we agreed upon this morning on stand up

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
Arkariang: nevermind, `ProjectUtils.userIsCreator(it, user)` needs to be back as creators must see the reply button

app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
```diff
@@ -113,9 +115,11 @@ class CommentCard @JvmOverloads constructor(
             cardCommentStatus == CommentCardStatus.COMMENT_WITHOUT_REPLIES ||
             cardCommentStatus != CommentCardStatus.DELETED_COMMENT
 
-        binding.commentActionGroup.isVisible = cardCommentStatus == CommentCardStatus.COMMENT_WITH_REPLIES ||
-            cardCommentStatus == CommentCardStatus.COMMENT_WITHOUT_REPLIES ||
-            cardCommentStatus == CommentCardStatus.FAILED_TO_SEND_COMMENT
+        if (shouldShowReplyButton(cardCommentStatus)) {
```
leighdouglas, line 118: Question here, why do we need this functionality if we are setting the reply button visibility directly from the view model? I think the  card status should be decoupled from the reply button. This status is representative of showing the actual replies, not the reply button. We are essentially performing this logic twice, when the card status is emitted and when the isReplyButtonVisible stream emits in the view holder.

app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
leighdouglas: Ties back into my previous comment, the card status is for the replies, not the reply button. Unless I'm missing something here 😅

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
```diff
@@ -95,21 +100,30 @@ interface CommentsViewHolderViewModel {
...
+                    this.commentCardStatus.onNext(cardStatus(it))
+                }
 
             this.commentInput
                 .compose(Transformers.combineLatestPair(environment.currentUser().observable()))
                 .compose(bindToLifecycle())
-                .subscribe { this.isCommentActionGroupVisible.onNext(isActionGroupVisible(it.first, it.second)) }
+                .subscribe {
+                    this.isReplyButtonVisible.onNext(
+                        shouldReplyButtonBeVisible(it.first, it.second, optimizely.isFeatureEnabled(OptimizelyFeature.Key.COMMENT_ENABLE_THREADS))
```
leighdouglas, line 124: I think this is the only place that we should have logic about showing and hiding the reply button.

app/src/main/java/com/kickstarter/ui/views/CommentCard.kt
Arkariang: You are completely right! as we spoke on the PR review session earlier we will handle this in a separate PR once @hadia 's work for the retry is completed to avoid her any further conflict 👍

Arkariang: About my previous comment, the reply button is tied to the delete status. So the visibility should be updated when changing status
Screen Shot 2021-06-04 at 11 04 20 AM

leighdouglas: Makes sense to me! If that's the case then let's leave it as is 👍🏽

leighdouglas (approved):
lgtm! 🎉

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
leighdouglas: Updated to add that the reply button is actually tied to the card status -> deleted and failed, so we need to update when the status is updated 👍🏽

PR #1274, opened 2021-06-04
NT-1914: UX – Post new comment

sunday-okpoluaefe:
# 📲 What

UX – Post new comment

# 🤔 Why

- User is scrolled to the top if not there already

- After posting a new comment we want to guarantee it’s visible, therefore scrolling to the top.

# 🛠 How

- Added a delay after inserting comment to ensure the scroll takes effect
```
 /*
         * A little delay after new item is inserted
         * This is necessary for the scroll to take effect
         */
        viewModel.outputs.insertComment()
            .compose(bindToLifecycle())
            .delay(200, TimeUnit.MILLISECONDS)
            .subscribe {
                binding.commentsRecyclerView.scrollToPosition(0)
            }
```

# 👀 See

| Before 🐛 | After 🦋 |

https://user-images.githubusercontent.com/63934292/120843103-9c533680-c565-11eb-9272-2362759c3905.mp4

# 📋 QA

- Create a comment on any backed project. The new comment should be shown and scrolled to the top

# Story 📖

https://kickstarter.atlassian.net/browse/NT-1914


app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.kt
```diff
@@ -154,9 +154,11 @@ interface CommentsViewModel {
                 .compose(Transformers.takeWhen(this.refresh))
                 .doOnNext {
                     this.isRefreshing.onNext(true)
-                    // reset cursor
+                }
+                .doOnNext {
```
Arkariang, line 158: You have two `doOnNext`, look at line 155, you can combine them

sunday-okpoluaefe: done

Arkariang (approved):

PR #1277, opened 2021-06-09
NT-1962:UX – Show removed comments

hadia:
# 📲 What
Let comment body invisible to keep constrain for view replays in case message was deleted 

# 🤔 Why

there was an issue when a comment deleted and has replies

# 🛠 How

 comment body invisible in case message was deleted

# 👀 See
#Before
image
#After
image

# 📋 QA
Check comments in that project "shenmue-3" scroll till found comments deleted has replies 

# Story 📖

https://kickstarter.atlassian.net/browse/NT-1962


Arkariang (approved):
👍

PR #1281, opened 2021-06-10
NT-2004: Fix comments and updates screen on Thank You page

leighdouglas:
# 📲 What

Previously, the if a user opened the comments page before backing a project, exited the comments page and then successful completed checkout, the comments page would pop up on top of the Thank you page. This PR fixes that bug so that the comments only appear when the text view is tapped on. 

The same thing was happening with the updates page, so I fixed that here the same way. 

# 🤔 Why

When the user successfully completes checkout, an update is sent to an observable in the project view model, which makes a network request to pull the latest project data. This was pushing new data to the observable controlling the comments button clicks, and because an event had already been triggered from the previous button click, this observable was not completed and still open listening for new data to come down the pipeline. This was causing the comments page to pop up on top of the thank you screen. 

# 🛠 How

Rearranged the architecture of this observable to only react when the comments/updates text view is tapped, and to then grab the latest data project/project data available instead of that data triggering the event itself. 

# 👀 See

Comments screen:
Jun-09-2021 20-34-37

Updates screen:
Jun-09-2021 20-35-18

# 📋 QA

Follow these steps for both updates and comments screen on two separate projects. These are two different flows:

1. Go to a project you are not backing
2. Tap on either comments or updates
3. On comments, you should see the disabled comments UI
4. Exit the page and return to the project page
5. Successfully complete checkout/backing
- There should only be the thank you screen, updates or comments pages should not open
6. Exit the thank you screen
7. On the project page, tap updates or comments
- The comments page should now be showing the enabled state because you are a backer
- Updates should open as normal. 

# Story 📖

[NT-2004: Comments showing on top of "Thank you" screen](https://kickstarter.atlassian.net/browse/NT-2004)


Arkariang (approved):
👍

PR #1283, opened 2021-06-10
NT-1908:UI for Comment Composer

hadia:
# 📲 What
 Fix UI bug on comment composer fixed.

# 🤔 Why

# 🛠 How

- set constrain to composer view and add gone margin 

# 👀 See

https://user-images.githubusercontent.com/1075310/121536361-4702a380-ca03-11eb-9230-e99fe8d53b85.mp4


Arkariang (approved):
👍

PR #1294, opened 2021-06-16
Fix "Reply" copy

leighdouglas:
# 📲 What

We were displaying the wrong copy for the "Reply" button on the comment card. This updates to use the correct string.

# 👀 See

English and Spanish translations for reply, as examples:

Screenshot_1623862463 Screenshot_1623862500

# 📋 QA

Go to the comments page on a project you have backed and check to make sure the Reply button is spelled correctly. Change device language to another language (Spanish or german for example) and recheck the comments page to make sure the correct translation is being used.

### DIFF
--- a/app/src/main/res/layout/comment_card.xml
+++ b/app/src/main/res/layout/comment_card.xml
@@ -59,9 +59,9 @@
 
     <androidx.appcompat.widget.AppCompatButton
         android:id="@+id/reply_button"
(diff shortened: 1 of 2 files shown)


Arkariang (commented):
👍  lgtm!

Arkariang (approved):
lgtm 👍

PR #1296, opened 2021-06-16
NT-1962:UX – Show removed comments

hadia:
# 📲 What

Remove the underline in the flagged message

# 🤔 Why
Change the flagged message and remove the underline
image

# 👀 See

image

# 📋 QA

Check comments in that project "shenmue-3" scroll till found comments deleted has replies

# Story 📖

https://kickstarter.atlassian.net/browse/NT-1962


Arkariang (approved):
lgtm!

PR #1308, opened 2021-06-29
NT-2069:Refactor Comment Pagination code 

hadia:
# 📲 What

Fix issue as on master Pagination change scrolling position and crash at the last page

# 🤔 Why

the compare item wasn't working as it was null for project value 

# 👀 See
Before issue
https://user-images.githubusercontent.com/1075310/123808314-ecb57e00-d8f0-11eb-9ac1-8f234d2885cb.mp4

the Fix
https://user-images.githubusercontent.com/1075310/123807680-5d0fcf80-d8f0-11eb-9cdd-fc75a668e52d.mp4

# 📋 QA

Open any project with comments and has more than one page and scroll  and check pagination  

# Story 📖

https://kickstarter.atlassian.net/browse/NT-2069

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.kt
@@ -101,6 +101,7 @@ interface CommentsViewModel {
         private var commentableId: String? = null
 
         private val isFetchingComments = BehaviorSubject.create<Boolean>()
+        private lateinit var project: Project
 
         init {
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.kt
```diff
@@ -301,7 +305,7 @@ interface CommentsViewModel {
                     .distinctUntilChanged(true)
                     .startOverWith(startOverWith)
                     .envelopeToListOfData {
-                        mapToCommentCardDataList(Pair(it, null))
+                        mapToCommentCardDataList(Pair(it, this.project))
```
leighdouglas, line 308: Since project can technically be null, I think it would be best if we put a null check here to guarantee that no crash happens.

hadia: Changed map method to not allow null

leighdouglas (approved):
lgtm! 🎉

PR #1314, opened 2021-07-06
NT-2099:Crash when we create new comment and Press reply button

hadia:
# 📲 What

Crash when we create a new comment and Press the reply button

# 🤔 Why

when we create a new comment and press reply we got crash as the comment wasn't posted  and doesn't have id

# 🛠 How

The replay button is visible after a comment posted successfully 
image

# 👀 See

https://user-images.githubusercontent.com/1075310/124662572-544b6a80-dea9-11eb-884a-e8a610fa4c33.mp4
https://user-images.githubusercontent.com/1075310/124662856-a2f90480-dea9-11eb-8ba1-7984c27029d4.mp4

# 📋 QA
1- Post a new comment  and click reply
2- Post a new comment  in offline mode and  then retry after posting click reply
3- add the first comment 

# Story 📖
https://kickstarter.atlassian.net/browse/NT-2099


leighdouglas (commented):
lgtm! 🎉 🤖

leighdouglas (approved):

PR #1322, opened 2021-07-15
NT-2000: Retrying posting a failed reply

leighdouglas:
# 📲 What

- Handle retrying the call to post a reply when the replying fails the first time. 
- Also handle bug that posts the reply as a root comment
- Hide the reply button on a reply comment card
- Linter cleanup

# 🛠 How

- Update the view comment replies and reply to comment streams with the new reply comment that has a parent id - we were previously sending the old comment without the updated id.

Screen Shot 2021-07-15 at 10 16 37 AM

# 👀 See

https://user-images.githubusercontent.com/19390326/125805579-6a21094f-fbe4-4496-8d56-2bdd8cfc3456.mp4

# 📋 QA

Reply on old root comment
- Open a project you have backed
- Reply to a comment on a backed project while in airplane mode
- Turn off airplane mode and retry posting the reply
- Reply should show success state "posted"
- Navigate back to root comments screen and pull to refresh a few times
- Confirm there are no duplicated comments, or that your reply was posted as a root comment
- If no replies we posted on the root comment before yours, pull to refresh should now show "view replies"
- Tap view replies and check that your reply is posted

Reply on new root comment
- Open a project you have backed
- Post a root comment in airplane mode
- Turn off airplane mode and retry posting the root comment
- Confirm success "posted" and tap reply
- Turn on airplane mode again and reply to root comment
- Turn off airplane mode and retry posting the reply
- Reply should show success state "posted"
- Navigate back to root comments screen and pull to refresh a few times
- Confirm there are no duplicated comments, or that your reply was posted as a root comment
- If no replies we posted on the root comment before yours, pull to refresh should now show "view replies"
- Tap view replies and check that your reply is posted

# Story 📖

[NT-2000: Retrying posting a failed reply](https://kickstarter.atlassian.net/browse/NT-2000)


app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
```diff
@@ -339,7 +349,10 @@ interface CommentsViewHolderViewModel {
          * // TODO: we will need the entire comment plus very important the [parentId]
          * @return Observable<Comment>
          */
-        private fun executePostCommentMutation(postCommentData: PostCommentData, errorObservable: BehaviorSubject<Throwable>) =
+        private fun executePostCommentMutation(
```
Arkariang, line 352: the TODO can be removed.

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
```diff
@@ -232,6 +234,7 @@ interface CommentsViewHolderViewModel {
 
             comment
                 .compose(takeWhen(this.onViewCommentRepliesButtonClicked))
+                .map { postedSuccessfully?.value ?: it }
```
Arkariang, line 237: the value for comment and postedSuccessFully should be the same. Why do we need to map here those two values?

leighdouglas: From my understanding, the comment that was emitted from the comment thread still had the default parentId of -1, which is why the reply was being treated as a root comment. Updating to use the comment item emitted from the postedSuccessfully stream gives us the updated parentId.

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
leighdouglas: 👍🏽

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
Arkariang: I would suggest using `Observable.merge` on the declaration for comment observable, something like: 
Screen Shot 2021-07-15 at 4 09 00 PM
 in that way you don't need to update several RXChains.

I'll approve though, my comment up there is a suggested improvement 👍

Arkariang (approved):

app/src/main/java/com/kickstarter/viewmodels/CommentsViewHolderViewModel.kt
leighdouglas: Sounds good, thanks for catching this. Just updated 👍🏽

PR #1323, opened 2021-07-15
NT-2093: Skip first emitted value from paginated data to avoid initially showing empty state showing on root comments screen

leighdouglas:
# 📲 What

Upon navigating to the root comments screen, an empty state was briefly shown before loading the comments. We want to avoid this.

# 🤔 Why

Paginated data is initiated with a value that emits by default no matter what. We want to ignore this initial value as it does not represent the actual list returned from the network call. 

# 🛠 How

Simply adding a skip operator to the rx chain to skip the initial value:
Screen Shot 2021-07-15 at 10 58 10 AM

# 👀 See
Before, you can briefly see the empty state shown before the comments load:
https://user-images.githubusercontent.com/19390326/125811602-8b356ffc-33c3-4980-a796-99467cebf7b7.mp4

After, empty state does not show: 

https://user-images.githubusercontent.com/19390326/125811735-f1fa016e-0c2d-46e4-88f5-667d5aef58c0.mp4

# 📋 QA

- Navigate to any project
- Make sure it is a project with comments
- Tap on comments
- Root comments screen should always show loading indicator first, and then the comments or empty state if there are no comments
- No empty state should be shown

# Story 📖

[NT-2093: UX – Empty state showing up when loading comments](https://kickstarter.atlassian.net/browse/NT-2093)


app/src/main/java/com/kickstarter/viewmodels/CommentsViewModel.kt
```diff
@@ -291,6 +291,7 @@ interface CommentsViewModel {
                 .subscribe(this.isFetchingComments)
 
             apolloPaginate.paginatedData()?.share()
+                ?.skip(1)
```
Arkariang, line 294: This will always jump the emision 1, are we sure that's the behaviour we are expecting?

PR #1327, opened 2021-07-16
NT-2000: Retrying posting a failed reply Scrolling bug

hadia:
# 📲 What

When we scroll up and down the error state disappear

# 🤔 Why

We need to bind failure and success state for reply

# 🛠 How
Binding different state 
image
image

# 👀 See

https://user-images.githubusercontent.com/1075310/126015239-bd49dedc-5f03-43d1-af47-a64ba4a6ed28.mp4

# 📋 QA

1. Test failed to post reply scroll up and back down  
2. Posted successfully  and scroll ups and down

# Story 📖

NT-2000: Retrying posting a failed reply


app/src/main/java/com/kickstarter/models/extensions/CommentExt.kt
```diff
@@ -0,0 +1,60 @@
...
+
+import com.kickstarter.models.Comment
+import com.kickstarter.ui.data.CommentCardData
+import com.kickstarter.ui.views.CommentCardStatus
+
+/**
+ * Update the internal persisted list of comments with the successful response
+ * from calling the Post Mutation
+ */
+fun Comment.updateCommentAfterSuccessfulPost(
```
leighdouglas, line 11: nice 😎

leighdouglas (approved):
Lgtm! 🚀 🎉

PR #1328, opened 2021-07-19
NT-1996 :UI/UX – Error loading thread

hadia:
# 📲 What

Handle empty replies list after initial load error refresh 
# 🤔 Why
https://console.firebase.google.com/u/2/project/android-internal-debug/crashlytics/app/android:com.kickstarter.kickstarter.internal.debug/issues/cc3a1a32b2e70ccf79f0b4c84e0d3d21?time=last-hour&sessionEventKey=60F58E77019A000169622BCBD6974674_1565093027825415149
image

# 🛠 How
https://www.loom.com/share/097bfe6d290d4ae5b70247419a2b35bd
- Create an empty state as setting section list to empty crash in 
Screen Shot 2021-07-19 at 5 07 09 PM

- set value to total count to null to over come distinctUntilChanged
image

# 👀 See

https://user-images.githubusercontent.com/1075310/126183486-90fd684a-cf8d-4772-901d-9874f8e00922.mp4

# 📋 QA

Open different comment Replies  in offline mode and the refresh
1- no Replies
2- Replies less than 7
3- pagnatianted Replies

# Story 📖

https://kickstarter.atlassian.net/browse/NT-1996


app/src/main/java/com/kickstarter/ui/activities/ThreadActivity.kt
```diff
@@ -78,6 +78,16 @@ class ThreadActivity :
                 this.repliesAdapter.takeData(it.first.reversed())
             }
 
+        this.viewModel.outputs
+            .onCommentReplies()
+            .compose(bindToLifecycle())
+            .observeOn(AndroidSchedulers.mainThread())
+            .filter { it.first.isNullOrEmpty() }
```
Arkariang, line 85: Can the filter be applied on the ViewModel?

hadia: @Arkariang  
I changed 
image

hadia (commented):
image

app/src/main/java/com/kickstarter/ui/activities/ThreadActivity.kt
```diff
@@ -69,11 +69,14 @@ class ThreadActivity :
             .onCommentReplies()
             .compose(bindToLifecycle())
             .observeOn(AndroidSchedulers.mainThread())
+            .doOnNext {
+                linearLayoutManager.stackFromEnd = false
+                /** bind View more cell if the replies more than 7 or update after refresh initial error state **/
+                this.repliesStatusAdapter.addViewMoreCell(it.second)
+            }
             .filter { it.first.isNotEmpty() }
             .doOnNext { linearLayoutManager.stackFromEnd = false }
```
Arkariang, line 78: Remove the second `.doOnNext` on the chain

Arkariang (approved):

PR #1330, opened 2021-07-19
NT-2001:UX – Show pending reply dialog

hadia:
# 📲 What
Show pending comment dialog in replies thread activity

# 🤔 Why

Add confirmation dialog when users have a failed post or something written in the comment composer in reply thread and try to Go back  to root comments

# 🛠 How
Checking if there are pending replies handling back  action with dialog 

# 👀 See

https://user-images.githubusercontent.com/1075310/126225178-6182e054-cad9-4c99-a4ba-849a9689ab94.mp4

# 📋 QA
add pending reply try back action submit reply and then try back again 

# Story 📖
https://kickstarter.atlassian.net/browse/NT-2001


app/src/main/java/com/kickstarter/ui/activities/ThreadActivity.kt
```diff
@@ -167,6 +169,48 @@ class ThreadActivity :
                     )
                 )
             }
+
+        viewModel.outputs.hasPendingComments()
+            .compose(bindToLifecycle())
+            .observeOn(AndroidSchedulers.mainThread())
+            .subscribe {
+                if (it) {
```
leighdouglas, line 177: instead of an if/else block, can we have use the `filter` operator to only emit if true call the handleBackAction? or do we have to explicitly call `viewmodel.input.backPressed()` in the else block? either way, this can be shorted to  one line `if (it) handleBackAction() else viewmodel.inputs.backpressed()`

hadia: we want to user press back and had pending comment to show dialog to confirm back
so we have to intercept back action and check if there is pending comments then show dialog  if not back

hadia: changed to if (it) handleBackAction() else viewmodel.inputs.backpressed()

Arkariang (approved):

PR #1332, opened 2021-07-22
NT-2001 :UX – Show pending dialog

hadia:
# 📲 What

show the confirmation dialog when the user tries to leave the screen with something written in the comment composer.
image

# 🤔 Why

So we show a dialog in case there is a text in comment composer in comment or replies activity

# 🛠 How
image

image

# 👀 See

https://user-images.githubusercontent.com/1075310/126643807-38b9ec09-f285-45d2-9a2b-ef683725dfe7.mp4

# 📋 QA

1. Open Comment and thread Activity
2. Write a comment in the composer 
3. Press back 

# Story 📖
https://kickstarter.atlassian.net/browse/NT-2001

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/CommentsActivity.kt
@@ -173,7 +173,11 @@ class CommentsActivity :
     }
 
     override fun back() {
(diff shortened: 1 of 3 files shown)


leighdouglas (approved):
Lgtm! 🎉

PR #1336, opened 2021-07-27
NT-2003:UX – Show pending comment dialog

hadia:
# 📲 What
Solve issue when pressing back icon  with text in the composer to show a dialog in comment activity

# 🛠 How
image

# 👀 See

https://user-images.githubusercontent.com/1075310/127163775-5562c2a9-c084-4bcf-a1bf-d73117bea659.mp4

# 📋 QA

Press back icon and hardware back key when there is unposted text in the composer 

# Story 📖

https://kickstarter.atlassian.net/browse/NT-2003


Arkariang (approved):
👍

PR #1337, opened 2021-07-27
NT-1948: UI User badges

leighdouglas:
# 📲 What

We want to show users badges on comment cards if a badge exists for a user. 

# 🛠 How

We take priority in this order: `You, Creator/Collaborator, Superbacker`

To find out if a user has a `You` badge, we compare the user id of the current user with the id of the user that attached to the comment.

If not, then we check the other cases in order, and since technically the list retrieved from the backend can have up to 2 badges in it, we want to make sure we are checking for the badges in priority order and assigning them accordingly. 

Since the badges are not tied to by the state of the card as we currently have it implemented (logged in, not logged in, backer or not backer), I created a separate enum to handle the state of the badges that changes their visibility. 

# 👀 See

Superbacker and you badges: 
superbacker and you

signed out:
signed out

Creator:
creator

# 📋 QA

Test both signed in and signed out states, "You" should not show if you are signed out

# Story 📖

[NT-1948: UI User badges](https://kickstarter.atlassian.net/browse/NT-1948)


app/src/main/java/com/kickstarter/viewmodels/RootCommentViewHolderViewModel.kt
```diff
@@ -42,6 +50,17 @@ interface RootCommentViewHolderViewModel {
                 .compose(Transformers.takeWhen(this.onShowCanceledPledgeRootCommentClicked))
                 .compose(bindToLifecycle())
                 .subscribe { this.showCanceledPledgeRootComment.onNext(CommentCardStatus.CANCELED_PLEDGE_COMMENT) }
+
+            commentCardData
+                .withLatestFrom(
+                    Observable.merge(
```
Arkariang, line 56: I think you can use here instead of 
`Observable.merge(
                        currentUser.loggedInUser(), ---> emmits `user` when there is one, do not emmit if no user
                        currentUser.loggedOutUser() ---> emmits `null` if no user, do not emmit in case there is an user
                    )`
directly 
`currentUser` observable, it will emmit `null` in case the user is logged out and `user` in case there is one logged inin.

leighdouglas: Thank you for the tip, updated 😄

Arkariang (approved):

PR #1344, opened 2021-08-04
NT-2141: Stepper component

Arkariang:
# 📲 What
- Migrated to kotlin the Playground activity in order to test the new components
- Create a reusable component as the stepper on the addOns card

# 🛠 How
- The stepper can configure as input  Max, Min, initial value, and the increment value.
- The stepper has as output the displayed number 
- Once reached the Max limit or the Min limit the increase/decrease buttons on the stepper are disabled.

# 👀 See

https://user-images.githubusercontent.com/4083656/128101261-9eacb157-7ae1-4cf8-a1d2-8bf332448dbd.mp4

|  |  |

# 📋 QA
- Open the playground activity on the internal tools screen, and play around with the stepper

# Story 📖


hadia (approved):
great job🎉🎉

PR #1350, opened 2021-08-06
[No-jira]: Ready for Ruby 3.0

Arkariang:
# 📲 What

- The necessary changes on our gems to be able to use Ruby 3

# 🤔 Why
- Integration with Circle CI Android Machine executor requires Ruby 3

# 🛠 How
- 'bundle update'

|  |  |

# 📋 QA

- On your environment now you need to have Ruby 3 (as tip you can use rvm such like this)
```
rvm install ruby-X.X.X           // will install the specified ruby version
rvm --default use ruby-X.X.X     // will set as default the specified ruby version 
```

- Now execute `bundle install` it most likely will promt you to update using `bundle update`.
- Now you can execute all the scripts like the translations one : 
* ` ./bin/milkrun update-strings `
or the one for creating a new internal build and release it to Firebase: 
* `bundle exec fastlane internal`


hadia (approved):
Great. nice to fix it 👌

PR #1354, opened 2021-08-10
Imartin/nt 2171 send position to view holder

Arkariang:
# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖


hadia (approved):
Approved

app/src/main/java/com/kickstarter/ui/viewholders/CommentCardViewHolder.kt
```diff
@@ -111,12 +111,12 @@ class CommentCardViewHolder(
         this.vm.outputs.isSuccessfullyPosted()
             .compose(bindToLifecycle())
             .compose(Transformers.observeForUI())
-            .subscribe { this.delegate.onCommentPostedSuccessFully(it) }
+            .subscribe { this.delegate.onCommentPostedSuccessFully(it, absoluteAdapterPosition) }
```
Arkariang, line 114: [documentation around `absoluteAdapterPosition` method ](https://developer.android.com/reference/androidx/recyclerview/widget/RecyclerView.ViewHolder#getAbsoluteAdapterPosition)

PR #1358, opened 2021-08-11
NT-2143: Create add-on reusable component and migrate current add-ons screen to use new component

leighdouglas:
# 📲 What

- Created the add-on reusable component
- Integrated the add-on stepper
- Migrated to use new component in the `BackingAddOnViewHolder`
- Updated `BackingAddOnViewHolderViewModel` logic
- Updated and added new tests to `BackingAddOnViewHolderViewModelTest`

# 🤔 Why

With the integration of screenshot testing, we need to move forward with creating reusable components.

# 🛠 How

- Add-on component has inputs in the form of setters, for things like text and visibility
- The component outputs display changes from the stepper as well as the add button visibility
- When display changes emit from the stepper, the add button disappears

# 👀 See

https://user-images.githubusercontent.com/19390326/129092758-ce4e0943-500f-488c-af36-953ce40d15ce.mp4

https://user-images.githubusercontent.com/19390326/129092767-3d49c7a5-aa16-4912-9d15-300520f63329.mp4

https://user-images.githubusercontent.com/19390326/129092772-e3907ac4-13b9-4095-9b36-b50e2dddb666.mp4

# 📋 QA

There are many different edge cases to test here:
- Description visibility
- Itemized list visibility
- Shipping visibility
- Estimated cost visibility if user is using a different currency than the project
- "Add" button should disappear when tapped, but reappear when stepper min (-) button reaches 0
- Stepper max (+) button should disable at whichever number is lowest between the limit or remaining quantity (for example, if limit is 6 but remaining is 2, stepper max button should disable once quantity reaches 2)
- When managing a reward, the selected add-on stepper max button should be disabled by default and only decrease, stepper quantity should be set to the quantity that the user originally checked out with by default, and the "Add" button should be gone upon visiting the screen

# Story 📖

[NT-2143: Refactor the AddOns screen to use the re-usable components](https://kickstarter.atlassian.net/browse/NT-2143)


app/src/main/java/com/kickstarter/ui/adapters/BackingAddOnsAdapter.kt
```diff
@@ -21,15 +21,15 @@ class BackingAddOnsAdapter(private val viewListener: BackingAddOnViewHolder.View
     }
 
     override fun layout(sectionRow: SectionRow): Int = when (sectionRow.section()) {
-        SECTION_BACKING_ADD_ONS_CARD -> R.layout.item_add_on_pledge
+        SECTION_BACKING_ADD_ONS_CARD -> R.layout.add_on_card
```
Arkariang, line 24: Should this not continue as `R.layout.item_add_on_pledge`?

Arkariang (approved):
- Apart from the comment around the `R.layout.item_add_on_pledge` everyting looks good. 👍  amazing work here @leighdouglas.

Once the layout thing it's fixed feel free to merge! I tested it as much as I could :)

PR #1360, opened 2021-08-12
NT-2182: Allow identify calls every time the app opens

Arkariang:
# 📲 What

- Allow identify calls every time the app opens

# 🤔 Why

- Long time logged users are not calling identify ever

# 👀 See

https://user-images.githubusercontent.com/4083656/129238210-4775f82e-2338-4bc8-a9ce-2a2218a76432.mp4

|  |  |

# 📋 QA

- Launch the app being logged in, go to background, come back from background, go again ... every time you come back you should see on the Segment dashboard an identify call.

# Story 📖

[NT-2182](https://kickstarter.atlassian.net/browse/NT-2182)


hadia (approved):
🎉🎉🎉

PR #1363, opened 2021-08-12
Your comment wasn't posted dialog showing always after error

hadia:
# 📲 What

Fix that the leave dialog  show even if the replies submitted successfully in thread activity 

# 🤔 Why

We have 2 different issues 

1.  adding extra space to text before submit
2- reversing the replies thread in the view

# 🛠 How
Using the position to compare thanks @Arkariang 
 Reverse the list from the server and add the replies to the top

# 👀 See

https://user-images.githubusercontent.com/1075310/129254520-ea0fc888-fb1a-4efc-a674-07eaa3ed7ed2.mp4

# 📋 QA

Add multi-line replies with extra space in offline mode try to submit 
press back

# Story 📖
https://kickstarter.atlassian.net/browse/NT-2181


Arkariang (approved):
👍  tested a few times.
- posting comments on replies works as usual
- the dialog only appears when there is pending comments to send or content on the composer view

PR #1365, opened 2021-08-12
[no-jira]: prepare for arctic fox

Arkariang:
# 📲 What

- Prepare for new Android Studio Arctic Fox 
- Working with Gradle 7
- Removed deprecated 
- Updated Kotlin version
- Compiling now with Java 11

# 👀 See
Screen Shot 2021-08-12 at 4 16 22 PM

|  |  |

# 📋 QA
- ⚠️  The only change you need on your environment is having Java 11 ready, as it is required with Gradle 7
see here how to specify the JDK location for the gradle plugin in case 11 is not your JAVa_HOME version: 
Screen Shot 2021-08-12 at 2 45 59 PM
- ⚠️ as tip I suggest using sdk man to handle your Java versions, if you have it installed you can active the version you need as default like so: 

`sdk install java 11.0.8.j9-adpt` // install specified version 

or install it like so: 

`sdk default java 11.0.8.j9-adpt`  // stablish as current the JDK 11

- [quick cheatsheet on sdk man tool here](https://gist.github.com/Arkariang/c2d130ad9bf361778326e0c9bc1c7334)


hadia (approved):
👏 great job

PR #1367, opened 2021-08-17
NT-2162: Remove comments feature flags

Arkariang:
# 📲 What
- Remove all the feature flags related to comments from the source code
- Remove all deprecated code related to the old comments screen
- Remove all the deprecated code related to comments and the V1 API client

# 👀 See
- No user facing changes
|  |  |

# 📋 QA
- Test you can see root comments
- Test you can see thread comments
- Test you can see the reply button and view replies button only on the root comments screen.
- Test same scenarios as before in updates

# Story 📖

[NT-2162](https://kickstarter.atlassian.net/browse/NT-2162)


hadia (approved):
🎉

PR #1368, opened 2021-08-18
[no-jira]: migrate KSUri.java to URI kotlin extension

Arkariang:
# 📲 What
- Currently there is the class `KSUri.java`, I created a kotlin `Uri` extension with the same logic as `KSUri.java` has.
- Unit test all the methods on `UriExt.kt`

# 🤔 Why
- Up coming work is gonna be focus on deeplinks, meaning we will add several new URI's to be handled and recognized by the app, so preparing the ground for it.

# 👀 See
- No user facing changes
|  |  |

# 📋 QA
- Not QA to do yet, but you can run the tests, `KSUri.java` has not yet been replaced, that would come on next PR


hadia (approved):
like it 😍

PR #1370, opened 2021-08-19
NTV-85: Target Android 12

hadia:
# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖


jacoco.gradle
```diff
@@ -0,0 +1,65 @@
+apply plugin: 'jacoco'
```
Arkariang, line 1: 💖

app/build.gradle
```diff
@@ -253,7 +189,6 @@ apollo {
 repositories {
     // This is the only reference we cannot remove yet as it has not been migrated to mavenCentral() or google()
```
Arkariang, line 190: This comments could also be removed as it was an explanation around why we were keeping jcenter() repository

PR #1377, opened 2021-08-23
Nt-2068: Fix update payment fails on pledges with addons, and no rewards

leighdouglas:
# 📲 What

- Users were unable to update payment when a reward had addons
- Also found that updating the payment was crashing on pledges with no rewards
- Added functionality that fixes both of these use cases 

# 🤔 Why

- The "Confirm" button being enabled is tied to a few different streams, one of which is `totalIsValid`. Up the line, this stream is tied to the shippingRule. For some reason, when a project had addons, we were not emitting a shipping rule, and therefore we were not emitting a total. Down the line, this meant that totalIsValid was never evaluated, and therefore the button never enabled. 
- This functionality is specific to backings with addons, but it seems that this use case is for rewards before the user backs the project. 
- This functionality was also causing a crash when updating the payment of a pledge with no reward, because this functionality was requiring the location id to not be null. If there is a pledge with no reward, there is no need for a shipping id 😅 

# 🛠 How

Screen Shot 2021-08-23 at 12 14 21 PM

# 👀 See
Before: User would not see total and the confirm button would be disabled

After: User can see their total, and confirm button is enabled:
Screenshot_1629735822 Screenshot_1629735879

# 📋 QA

This should be very thoroughly QAed, as there are many many many use-cases:
- Pledge with addons
- Pledge without addons
- Pledge with no reward
- Digital pledge with addons
- Digital pledge without addons
- There might be more 😅 

# Story 📖

[NT-2068: Manage Pledge – Unable to change payment method](https://kickstarter.atlassian.net/browse/NT-2068)
[NT-2195: Manage Pledge - Update payment on pledge with no reward crashes](https://kickstarter.atlassian.net/browse/NT-2195)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
@@ -2,7 +2,6 @@ package com.kickstarter.viewmodels
 
 import android.content.SharedPreferences
 import android.text.SpannableString
-import android.util.Log
 import android.util.Pair
 import androidx.annotation.NonNull
 import com.kickstarter.R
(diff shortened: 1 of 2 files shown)


hadia (approved):
Nice work 👌

PR #1387, opened 2021-09-09
NTV-41: Add deeplink for discovery with collections

leighdouglas:
# 📲 What

We can now deeplink into the discovery activity with a sort parameter.

# 🤔 Why

Updated the DiscoveryParams fromUri() method to check for the collections parameter being set to true or false. 

# 🛠 How

It seems that in the past we had urls where the query param would be true instead of 1, so  I updated the existing functionality to account for both instances. 

# 👀 See
(I'm not following anyone so the social deeplink comes up blank 😅 )

https://user-images.githubusercontent.com/19390326/132756051-de59ceca-ef6f-476a-af45-b2d1502f263e.mp4

# 📋 QA

Thoroughly test all discovery deeplinks

# Story 📖

[NTV-41: Handle collections](https://kickstarter.atlassian.net/browse/NTV-24)


Arkariang (approved):
lgtm!

PR #1392, opened 2021-09-16
Ntv:166 confirm button functionality

hadia:
# 📲 What
Handle the understand button in the bottom sheet 

# 🤔 Why
Check that pledge is done only in case the dialog is confirmed

# 🛠 How
Create delegate to handle button action 
image

# 👀 See

Uploading device-2021-09-15-185841.mp4…

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA
Pledge project 
Update Pledge or payment 

# Story 📖
https://kickstarter.atlassian.net/browse/NTV-166


Arkariang (approved):
Tested on my device, looks great! 👍

PR #1394, opened 2021-09-16
NTV-183: Fix discovery crash when changing sort after choosing category

leighdouglas:
# 📲 What

After introducing the new deeplinking logic to the discovery screen, we ran into a crash that was happening when choosing a category from the navigation drawer, and then selecting a sort different sort. This crash was happening if you followed these steps right after app bootstrap. 

# 🤔 Why

**TLDR; We updated the discovery params sent with navigation drawer clicks to take the current sort if the sort is null. This sometimes null sort was causing a crash when trying to be sent to our analytics client. Below is an in-depth and possibly unnecessary explanation, feel free to skip 😅** 

A discovery params object holds all of the configurations for the discovery screen (which sort the user is on, category, collections, etc). Everytime the user interacts with the screen, we send a new discovery params object to the view model. The streams take this data and update the screen accordingly. However, because this is updated in several places throughout the activity and viewmodel, maintaining all of the different edge cases while introducing new logic can be tricky. 

This specific crash was being caused when the user taps a category from the navigation drawer, and the sort passed down with it was null. Tapping a category sends a new discovery params object to the view model that sometimes does not contain a sort type. This null sort was accounted for previously by ALWAYS setting the current discovery params (no matter what their source was) with the current sort. However, we needed to tweak this logic when deeplinking to a specific sort was introduced, as the deeplinked sort was being overridden with the current sort, and hence would not change sort tabs. 

At first, we removed this logic completely, but when trying to pass the current sort to the segment analytic event after tapping a drawer item, it was null and caused this crash. So instead, since we still needed this logic sometimes, but it no longer applied to ALL discovery params, we extrapolated it to the specific use case of the discovery params sent with the navigation drawer clicks.

# 🛠 How

Update discovery params from drawer clicks with the current sort if sort passed down is empty:
Screen Shot 2021-09-16 at 1 55 08 PM

# 📋 QA

- Switching tabs while on a category should not cause a crash
- Switching categories while on a sort should not revert it to magic (so if you're on popular and switch categories to art, you should still be on popular)
- Make sure all deeplinks work properly

# Story 📖

[NTV-183: 💥 – Explore with category/subcategory + sorting](https://kickstarter.atlassian.net/browse/NTV-183)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.kt
@@ -268,9 +268,17 @@ interface DiscoveryViewModel {
             val paramsFromIntent = intent()
                 .flatMap { DiscoveryIntentMapper.params(it, apiClient) }
 
+            val pagerSelectedPage = pagerSetPrimaryPage.distinctUntilChanged()
+
             val drawerParamsClicked = childFilterRowClick
                 .mergeWith(topFilterRowClick)
-                .map { it.params() }
+                .withLatestFrom(
+                    pagerSelectedPage.map { DiscoveryUtils.sortFromPosition(it) }
+                ) { drawerClickParams, currentParams ->
+                    if (drawerClickParams.params().sort() == null)
+                        drawerClickParams.params().toBuilder().sort(currentParams).build()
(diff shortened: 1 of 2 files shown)


Arkariang (approved):
lgtm!

PR #1400, opened 2021-09-18
[NTV-174 & NTV-175]: fetch faq &  environmentCommitment from GraphQL

Arkariang:
# 📲 What

- Query Faq and add it to the project data model for future work
- Query EnvironmentalCommitment and add it to the project data model for future work

# 🛠 How
- 2 new fragments
- added the two new fragments to `fullProject` fragment 
- `ProjectFaq` new model created
- `EnvironmentalCommitment` new model created
- `Project.java` model has two new fields :`projectFaqs: List<ProjectFaq>,  .envCommitments: List<EnvironmentalCommitment>`
- 2 new package level functions to convert GraphQL data in our DataModels
- Tests for the new models
- Tests for the transformers

# 👀 See
- You can see here how the project retrieved from GrapQL now includes the faq's and environmental commitment fields
Screen Shot 2021-09-17 at 5 15 34 PM

|  |  |

# 📋 QA

ℹ️  We should start moving all the transformer private methods inside `KSApolloClient` into the new file`GraphQLTransformers.kt ` to be tested 😄  let me know if any concern/doubts on it
- No user facing changes yet, but if you want to do some QA here put a breakpoint on `ProjectViewModel` on the currentproject observable and map the two new fields :) 

# Story 📖

[NTV-174](https://kickstarter.atlassian.net/browse/NTV-174)
[NTV-175](https://kickstarter.atlassian.net/browse/NTV-175)


app/src/main/java/com/kickstarter/services/KSApolloClient.kt
```diff
@@ -1169,11 +1158,13 @@ private fun projectTransformer(projectFragment: FullProject?): Project {
         .staticUsdRate(staticUSDRate)
         .usdExchangeRate(usdExchangeRate)
         .updatedAt(updatedAt)
-        // .unreadMessagesCount() TODO: unread messages can be fetched form the User Object
-        // .unseenActivityCount() TODO: unseen activity can be fetched form the User Object
+        // .unreadMessagesCount() unread messages can be fetched form the User Object
```
leighdouglas, line 1161: Nit: Do we need this commented out code? If not we should remove it, unless it's providing context.

leighdouglas (approved):
Awesome! Lgtm! 🚀

app/src/main/java/com/kickstarter/services/KSApolloClient.kt
Arkariang: It's for providing context, as those fields have value when fetched from V1, but  on GraphQL they do not exist under Project

PR #1402, opened 2021-09-20
[No-Jira]: Remove unnecessary network call

Arkariang:
# 📲 What

- Before adding the project query, we were getting project information from V1, and backing information from GraphQL
- Now that the entire project is fetched from GraphQL we added to the project query the backing query as well, making redundant the call to `apolloClient.getProjectBacking` as the information is already on the Project 
- Added initialization for two fields for project model 
- Exchanged `flatmap` over `switchmap` operator

# 🛠 How
- Remove network call to fetch backing information as it's already included in the project query.

# 👀 See
- No visible user facing changes, improvements on the reduction of networking calls
|  |  |

# 📋 QA
- Back a project.
- Load several previously backed project, and view your pledge
- View your pledge from a message


leighdouglas (approved):
lgtm! 🚀 🤖

PR #1403, opened 2021-09-21
NTV-163: Optimizely experiment flag for risk message modal

leighdouglas:
# 📲 What

An experiment is setup in optimizely for the risk messaging modal. When the user is in the `variant_1` group, it will display the risk message. If in they are in the `control` group, the normal flow will be presented (no risk modal).

# 🛠 How

Check if user is in the `control` group or in one of the variants, and default to normal flow if the user is in `control`. 

# 📋 QA

- Add your user ID to the whitelist here: [Native Risk Messaging Whitelist](https://app.optimizely.com/v2/projects/17442040660/experiments/20585857324/whitelist)
- Test both the `control` flow and the `variant_1` flow by changing your group on the whitelist. `control` should not show dialog, `variant_1` should show dialog

# Story 📖

[NTV-163: Optimizely experiment setup](https://app.optimizely.com/v2/projects/17442040660/experiments/20585857324/whitelist)


Arkariang (approved):
Tested on my environment 👍  looks good to me

PR #1404, opened 2021-09-21
NTV-204: Add new ProjectPageActivity and ProjectPageViewModel, and remove references to overview recyclerview

leighdouglas:
# 📲 What

- Create new `ProjectPageActivity`
- Remove all references to the overview recyclerview
- Added a button that launches what will be the new project activity from the playground

# 🤔 Why

- We will eventually migrate to the new ProjectPageActivity, and we need to differentiate between the two. 
- Adding a button that launches the project activity makes testing easier.
- Redesigning the overview, so the recyclerview is no longer needed. 

# 👀 See

https://user-images.githubusercontent.com/19390326/134404511-18684a18-aa76-4450-b854-5791fe775224.mp4

# 📋 QA

No user facing changes, but you can test that the project activity opens using the button from the playground. 
Also test current project pages to make sure they're working as usual.

# Story 📖

[NTV-204: ProjectPageActivity Refactor](https://kickstarter.atlassian.net/browse/NTV-204)


app/src/main/java/com/kickstarter/ui/activities/ProjectPageActivity.kt
```diff
@@ -0,0 +1,792 @@
...
+    BackingFragment.BackingDelegate {
+    private lateinit var adapter: ProjectAdapter
+    private lateinit var ksString: KSString
+
+    private val projectShareLabelString = R.string.project_accessibility_button_share_label
+    private val projectShareCopyString = R.string.project_share_twitter_message
+    private val projectStarConfirmationString = R.string.project_star_confirmation
+
+    private val animDuration = 200L
+    private lateinit var binding: ActivityProjectBinding
```
Arkariang, line 71: You should be loading anothe XML file here, therefor another Binding, without the recyclerView

app/src/main/java/com/kickstarter/ui/activities/ProjectPageActivity.kt
```diff
@@ -0,0 +1,792 @@
...
+            for (fragment in fragments) {
+                if (fragment == lastFragmentWithView) {
+                    fragment.view?.importantForAccessibility = View.IMPORTANT_FOR_ACCESSIBILITY_YES
+                } else {
+                    fragment.view?.importantForAccessibility = View.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS
+                }
+            }
+        }
+
+        this.adapter = ProjectAdapter(this.viewModel)
```
Arkariang, line 103: The adapter is no longer needed

Arkariang (commented):
@leighdouglas when opening the activity from the playgraound it crashes. The ViewModel and ViewModelTest needs to be duplicated/refactored as well

leighdouglas:
@Arkariang duplicated the view model and tests, also confirmed that crash is only happening when you are not logged in. This will be fixed later when it is actually loaded with the correct configuration.

Arkariang (approved):
@leighdouglas 👍  cool, once https://kickstarter.atlassian.net/browse/NTV-203 is finished we might need to come back and review any logic/issue once the new screen is plugged in with the rest of the flow

PR #1407, opened 2021-09-23
NTV-189: Wrong bonus update pledge

Arkariang:
# 📲 What

- Some backers were reporting invalid bonus amount when updating pledge or fixing pledge

# 🤔 Why
- Bug unraveled finally, the only transformation we do to the bonus amount when moving from `manage your pledge` screen to `update your pledge` screen is a parsing from String to Double like so
Screen Shot 2021-09-23 at 3 10 57 PM
`NumberUtils.parse(String, Locale)`, the issue lays here, every language uses different characters for decimal symbols and the character for thousands. 
- Some countries will do `0,99` and others `0.99` 
Reason why we have in mind the Locale when parsing. Well now, not all the languages are listed as Default Locale like is the case for Spanish or Polish, take a look here to see the available list for Java -> 

[Locale.java](https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/classes/java/util/Locale.java#L490)

- Basically the bonus amount, if the device Locale is in Spanish, was transformed from "X.0"  to "X0.0"
Screen Shot 2021-09-22 at 3 03 26 PM

- Said so I have not been able to reproduce in all my devices when changing the language to Spanish, on the Pixel device I was completely unable to reproduce it. But could be reproduced on my Nokia device.

# 🛠 How
- Implemented a String extension that will use a few regex to identify the format for the number and transform it correctly into a double.

# 👀 See
| Before 🐛 | 

https://user-images.githubusercontent.com/4083656/134591770-699fbc56-da9e-4149-8037-5958ed4b07c4.mp4

|After 🦋 |

https://user-images.githubusercontent.com/4083656/134591789-f0f7c940-1b63-4f3a-98ff-6699b6c9c4c1.mp4

|  |  |

# 📋 QA
- Change the language of your device to Spanish or Polish, pledge to a project with bonus amount and the try to update the pledge.

# Story 📖

[NTV-189](https://kickstarter.atlassian.net/browse/NTV-189)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/NumberUtils.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/NumberUtils.java
@@ -113,6 +113,10 @@ public final class NumberUtils {
     return parse(input, Locale.getDefault());
   }
 
+  @Deprecated
+  /**
+   * Use instead @link com.kickstarter.libs.utils.extensions.StringExt#toDouble(java.lang.String)
+   */
   public static double parse(final @NonNull String input, final @NonNull Locale locale) {
     try {
       return NumberFormat.getNumberInstance(locale).parse(input).doubleValue();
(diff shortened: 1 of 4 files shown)


hadia (approved):
😍

PR #1417, opened 2021-10-05
NTV-238: projectData emitted several times with different values on "EnvironmentalCommitment"

Arkariang:
# 📲 What

- Reduce number of emissions for the initial project, to emit ONLY after the network call is successful.
- Fixed snapshot tests 
- Modified Unit test the reflect the reduced number of emissions

# 🤔 Why
- The tab/fragment for EnvironmentalCommitment on the navigation selector needs to be hidden when empty, we were getting false negative emissions
- Snapshot tests were having issues

# 👀 See
- No user facing changes
|  |  |

# 📋 QA
- Load the project page as usual
- All test passing again

# Story 📖

[NTV-238](https://kickstarter.atlassian.net/browse/NTV-238)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ jobs:
             at: ~/project
       - android/create-avd:
           avd-name: Emulator
-          system-image: system-images;android-31;google_apis;x86
+          system-image: system-images;android-30;google_apis;x86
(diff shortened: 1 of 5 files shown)


hadia (approved):
nice 👌👌

PR #1452, opened 2021-11-10
NTV-270: Extend image parsing to handle captions, links

Arkariang:
# 📲 What

- The parser now is able to detect Images with captions and links

# 🛠 How
When the parser detects an image/gif, or an image/gif with caption, or an image/gif with a link, or an image/gif with caption and Link it will return a `ImageViewElement` with the populated fields according to each case: 
`data class ImageViewElement(
    val src: String,
    val href: String? = null,
    val caption: String? = null
) : ViewElement`

# 👀 See

ImageParser
| --- | --- |
|  |  |

# 📋 QA
No user facing changes
- make sure the list of elements the parser returns does not contain any duplicate.
- make sure if the project html contains an image or gif with or without caption and with or without link the ImageViewElement on each case has the correct fields populated.

# Story 📖

[NTV-270](https://kickstarter.atlassian.net/browse/NTV-270)


app/src/main/java/com/kickstarter/libs/htmlparser/ElementExt.kt
```diff
@@ -0,0 +1,55 @@
...
+    var src = ""
+    var caption: String? = null
+    var href: String? = null
+
+    if (this.parent()?.tag()?.name == "a") {
+        href = this.parent()?.attr("href")
+    }
+
+    caption = this.attr("data-caption")
+    src = this.children().getOrNull(0)?.children()?.getOrNull(0)?.attr("src").toString()
```
hadia, line 47: For gif images the URL is in that `data-src=$data_src ` we need to get the gif  url from data-src
```
if (element.hasAttr("data-src")) {
    element.attributes()["data-src"].apply {
        viewElements.add(ImageViewElement(this))
    }
} else {
    element.attributes()["src"].apply {
        viewElements.add(ImageViewElement(this))
    }
}
```
image

app/src/test/java/com/kickstarter/libs/htmlparser/HTMLParserTest.kt
```diff
@@ -42,4 +42,78 @@ class HTMLParserTest {
...
+        val imageView: ImageViewElement = listOfElements.first() as ImageViewElement
+        assert(listOfElements.size == 1)
+        assert(imageView.caption == "This is an Android with a caption and a link")
+        assert(imageView.href == href)
+        assert(imageView.src == src)
+    }
+
+    @Test
+    fun parseGifWithCaptionAndLink() {
+        val src = "https://ksr-qa-ugc.imgix.net/assets/035/272/962/ad1848184f8254f017730e6978565521_original.gif?ixlib=rb-4.0.2&w=700&fit=max&v=1635378954&auto=format&frame=1&q=92&s=fae855ae1e9f3919c1631c074419cd43"
```
hadia, line 101: For gif images the src and data-src are different

app/src/main/java/com/kickstarter/libs/htmlparser/ElementExt.kt
Arkariang: Done! line 50

hadia (approved):

PR #1459, opened 2021-11-16
no-jira: flaky tests around datetime

Arkariang:
# 📲 What

- For some tests two DateTime Objects were created, making the test flaky, failing randomly


leighdouglas (approved):
Thanks for catching this!

PR #1470, opened 2021-11-29
NTV-271:Parser – Fix video element

hadia:
# 📲 What
modify the parser to return a VideoViewElement(url:String) instead of VideoViewElement(list:ArrayList)

# 🛠 How
we use the high URL else use the first URL 

Screen Shot 2021-11-29 at 1 28 01 PM

# 👀 See
Screen Shot 2021-11-29 at 1 28 39 PM

# 📋 QA
Check parser and test class

https://kickstarter.atlassian.net/browse/NTV-271

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/htmlparser/HTMLParser.kt
+++ b/app/src/main/java/com/kickstarter/libs/htmlparser/HTMLParser.kt
@@ -33,7 +33,10 @@ class HTMLParser {
                 }
                 ViewElementType.VIDEO -> {
                     val sourceUrls = element.children().mapNotNull { it.attr("src") }
-                    val videoViewElement = VideoViewElement(ArrayList(sourceUrls))
+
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/libs/htmlparser/HTMLParser.kt
```diff
@@ -33,7 +33,10 @@ class HTMLParser {
                 }
                 ViewElementType.VIDEO -> {
                     val sourceUrls = element.children().mapNotNull { it.attr("src") }
-                    val videoViewElement = VideoViewElement(ArrayList(sourceUrls))
+
+                    val videoViewElement = VideoViewElement(
```
Arkariang, line 37: Can you move it to a VideoViewElement extension? 🙏

Arkariang (approved):

PR #1489, opened 2021-12-16
NTV-320: migrate to kotlin preferences package

Arkariang:
# 📲 What

- Migrated to kotlin  the entire `preferences` package
Screen Shot 2021-12-16 at 4 45 42 PM
- Added new class `BooleanDataStore` as experiment, it uses jetpack `Preferences DataStore` internally.
- In next tickest if the investigation goes well `BooleanPreferences` class will be substituted by `BooleanDataStore`

# 👀 See
- No user facing changes

|  |  |

# 📋 QA
- Delete cache and all storaged data from the app, log again, all should keep working as usual.

# Story 📖

[NTV-320](https://kickstarter.atlassian.net/browse/NTV-320)


hadia (approved):
nice work

PR #1491, opened 2021-12-20
NTV-329 :Picasso IllegalArgumentException error on 3.0.0

hadia:
# 📲 What
image

# 🛠 How
Check if the small image is empty replace it with the medium if both are empty hide the image 

# 👀 See

image

# 📋 QA
Fellow "https://www.kickstarter.com/profile/1840vienna"
Open one of his backed projects "https://www.kickstarter.com/projects/joshuastarr/shikoku-1889?ref=profile_backed"  
image

# Story 📖
(https://kickstarter.atlassian.net/browse/NTV-329)


app/src/main/java/com/kickstarter/ui/viewholders/ProjectViewHolder.kt
```diff
@@ -513,9 +514,10 @@ class ProjectViewHolder(
             .compose(bindToLifecycle())
             .compose(observeForUI())
             .subscribe { url: String? ->
-                url?.let {
+                binding.projectSocialImage.isGone = url.isNullOrEmpty()
+                if (!url.isNullOrEmpty()) {
```
Arkariang, line 518: could the ViewModel just emit a non optional string url, instead of checking that logic on the ViewHolder

hadia: we do the issue if its empty i i will remove the optional check

Arkariang (approved):

PR #1506, opened 2022-01-11
[no-jira]: Bug on UI 

Arkariang:
# 📲 What

- On large screens (pixel 6 pro in this case), the animation to expand the pledge fragment was not having in account the status bar, leaving a blank space on the bottom.

# 👀 See

Screen Shot 2022-01-10 at 5 10 19 PM

|  |  |

# 📋 QA

- Load the app on several simulator with several screen images,


hadia (approved):
Tested on emulator working fine 😍

PR #1524, opened 2022-01-25
NTV-356: Joined queries to fecth projects

Arkariang:
# 📲 What

- On the spike we created two queries following V1 structure of endpoints, those two queries have been joined in only one, and added all the missing parameter.

# 🤔 Why

- V1 to GraphQL migration.

# 👀 See
- No user facing changes, the queries have not yet being called on a ViewModel level, just polishing the queries.

# 📋 QA

- Uncomment the code that makes the call to Apollo client, and comment the code that makes the networking call to V1. Except for the drawer option `Backed by people you follow` you should not appreciate any change when using Apollo client.

# Story 📖

[NTV-356](https://kickstarter.atlassian.net/browse/NTV-356)


leighdouglas (approved):
lgtm! 🎉

PR #1535, opened 2022-02-03
[no-jira]: percentage funded

Arkariang:
# 📲 What

- The percentage was missing after the migration to GraphQL

# 🤔 Why

- `percentage` field was not queried

# 👀 See
| Before 🐛 | A
Screen Shot 2022-02-03 at 3 56 39 PM
fter 🦋 |

|  |  |

# 📋 QA

- Check the percentage and progress bar are loading correctly


hadia (approved):
👌

PR #1554, opened 2022-03-02
NTV-366: DiscoveryParams migrated to kotlin and deprecated autoparcel

Arkariang:
# 📲 What

- Migrated to kotlin
- Adopt @Parcelize 

# 🤔 Why
- Remove all references to AutoParcel library in Models, eventually we will be able to remove that dependency.

# 👀 See
- No user facing changes

|  |  |

# 📋 QA
- Launch the app logged out
- Launch the app logged in
- Try the project deeplink (load any project from staging on the device browser, you should see the button to open the app, that would test the deeplink)

# Story 📖

[NTV-366](https://kickstarter.atlassian.net/browse/NTV-366)


hadia (approved):
👌

PR #1560, opened 2022-03-07
NTV-423: bug fixing on Campaign tabg

Arkariang:
# 📲 What
- Sometimes you could see image placeholder on top of the images available to pinch and zoom
image

- Campaign tab name replicated on the rest of the tabs
Screen Shot 2022-03-07 at 2 12 37 PM

# 🛠 How
- The images able to pinch and zoom had the same image below it to keep the structure of the content when doing the pinch  and zoom, the placeholder was loaded with the same drawable as the top image exchanged `setImageBitmap(target.drawable.toBitmap())` over `.setImageDrawable(target.drawable)` in order to avoid the `.toBitmap` transformation
- When a cell from the recyclerview is recycled clean the previous image 

# 📋 QA

- Load many campaigns, you should not see the placeholder image on top anymore. 
- Make sure you don't see the repeated Campaign name on the tabs

# Story 📖

[NTV-423](https://kickstarter.atlassian.net/browse/NTV-423)


leighdouglas (approved):
Lgtm! 🎉

PR #1564, opened 2022-03-15
NTV-386: Migrate SurveyResponse.java to kotlin and deprecate Autoparcel

hadia:
# 📲 What

- Migrated to kotlin
- Adopt @Parcelize 

# 🤔 Why

# 👀 See

https://user-images.githubusercontent.com/1075310/158448589-a8953132-5c80-4a67-8428-e197792f848c.mp4

# 📋 QA
- Launch the app Kickstarter native account on prod and check survey items 

# Story 📖
https://kickstarter.atlassian.net/browse/NTV-386


app/src/main/java/com/kickstarter/models/SurveyResponse.kt
```diff
@@ -0,0 +1,131 @@
...
+            ) : Parcelable {
+                fun survey(survey: String?) = apply { this.survey = survey }
+                fun build() = Web(
+                    survey = survey
+                )
+            }
+
+            fun toBuilder() = Builder(
+                survey = survey
+            )
```
Arkariang, line 92: Same as previous line

Arkariang (approved):

PR #1569, opened 2022-03-21
NTV-440 : Deprecate Autoparcel for ErroredBacking.kt

hadia:
# 📲 What

- Migrated to kotlin
- Adopt @Parcelize 

# 🤔 Why

# Story 📖
https://kickstarter.atlassian.net/browse/NTV-440


Arkariang (approved):
lgtm! Tested it on production with my personal account 👍

PR #1581, opened 2022-03-25
NTV-445 : Migrate ProjectNotificationBody.java to kotlin and deprecate Autoparcel

hadia:
# 📲 What

- Migrated to kotlin
- Adopt @Parcelize 

# 🤔 Why

# Story 📖
https://kickstarter.atlassian.net/browse/NTV-445


Arkariang (approved):
lgtm!

PR #1585, opened 2022-03-29
NTV-443 : Migrate LoginWithFacebookBody.java to kotlin and deprecate Autoparcel

hadia:
# 📲 What

- Migrated to kotlin
- Adopt @Parcelize 

# 🤔 Why

# 📋 QA
test login with Facebook

# Story 📖
https://kickstarter.atlassian.net/browse/NTV-443


Arkariang (approved):
lgtm! Tested with log in to facebook with the native account 👍

PR #1592, opened 2022-04-05
NTV-476 :Migrate AccessTokenEnvelope.java to kotlin and deprecate Autoparcel

hadia:
# 📲 What

- Migrated to kotlin
- Adopt @Parcelize 

# 🤔 Why

# 👀 See
Uploading device-2022-04-06-000408.mp4…

# 📋 QA

Test Access token after login 

# Story 📖
https://kickstarter.atlassian.net/browse/NTV-476


Arkariang (approved):
lgtm!
- Tested login in with an account without two factor auth
- Tested login in with facebook
- Tested logn in with an account with 2fa active

PR #1609, opened 2022-04-18
NTV-494: Deprecate Autoparcel for ProjectData.kt

hadia:
# 📲 What

- Migrated to kotlin
- Adopt @Parcelize 

# 🤔 Why

# 👀 See

Screen Shot 2022-04-18 at 5 14 57 PM

# 📋 QA
test Project data parameter in segment events 

# Story 📖
https://kickstarter.atlassian.net/browse/NTV-494


Arkariang (approved):
lgtm

PR #1616, opened 2022-04-27
[No-jira]: Update schema.json

Arkariang:
# 📲 What

- Updated schema.json due to the addition of new queries to GrahpQL


hadia (approved):
thanks ☺️

PR #1626, opened 2022-05-12
NTV-380  :Migrate Environment.java to kotlin and deprecate Autoparcel

hadia:
# 📲 What

- Migrated to kotlin
- Adopt @Parcelize 

# 🤔 Why

# 📋 QA
Test  Who App for Environment object used for local, user object ,sharedpreference

# Story 📖
https://kickstarter.atlassian.net/browse/NTV-380


leighdouglas (approved):
Awesome! 🚀🚀

PR #1645, opened 2022-06-16
WEB-568: [Android] Filter root comments with no replies from canceled backer in view model

hadia:
# 📲 What
 As a creator, if a backer cancels their pledge, I want to minimize the visual impact of comment tombstones in the comments of my project.

# 🤔 Why
Comments from canceled backers with no replies do not appear in the list of comments

# 🛠 How
Filter the response in the view model before mapping to remove the Comments from canceled backers with no replies 

# 👀 See
https://user-images.githubusercontent.com/1075310/174097204-ae9df316-b566-426d-9ece-013beaa84c77.mp4

# 📋 QA
1. Back project 
2. Add comments and reply for project and updates 
3. Cancel a pledge

# Story 📖
https://kickstarter.atlassian.net/browse/WEB-568


leighdouglas (approved):
lgtm!

PR #1650, opened 2022-06-28
WEB-605: Update string for flagged comments

leighdouglas:
# 📲 What

Product has updated the flagged comment copy to be a little more clear. 

# 👀 See

New copy: 
Screenshot_20220628-130146

# 📋 QA

Comments should all work as they did before, tests all states between project and update comments. 

# Story 📖

[WEB-605: Update string for flagged comments](https://kickstarter.atlassian.net/browse/WEB-605)


Arkariang (approved):
lgtm, just fix the linter and good to go!

PR #1651, opened 2022-06-30
NTV-569 :  Migrate Signin/Login related classes to kotlin

hadia:
# 📲 What
Migrate classes from java to kotlin

1. ` LoginReason `
2.  `LoginToutViewModel `
3.  `SignupViewModel `
4.  `TwoFactorViewModel `
6.  `LoginPopupMenu `

# 🤔 Why
Migrate All classes to java.

# 📋 QA
test Message Sign/Sigup 2FA 

# Story 📖
https://kickstarter.atlassian.net/browse/NTV-569


app/src/main/java/com/kickstarter/viewmodels/LoginToutViewModel.kt
```diff
@@ -0,0 +1,297 @@
...
+        ActivityViewModel<LoginToutActivity?>(environment),
+        Inputs,
+        Outputs {
+        private var callbackManager: CallbackManager? = null
+        private val currentUser: CurrentUserType
+        private val client: ApiClientType
+
+        private fun clearFacebookSession(e: FacebookException) {
+            LoginManager.getInstance().logOut()
+        }
```
leighdouglas, line 90: Noticed there are a bunch of codecov warnings here, should these be covered by tests? Is this a codecov error related to the other issues we've been having with it?

leighdouglas (approved):
Left a comment about code coverage on one particular view model, otherwise tested on staging and prod and everything is working as it should! 🚀

app/src/main/java/com/kickstarter/viewmodels/LoginToutViewModel.kt
Arkariang: For being able to properly test those callbacks we should've created a wrapper client around FacebookSDK (Dependency Inversion Principle, as we do for Apollo, Retrofit, or Braze ->  [DIP](https://dev.to/tamerlang/understanding-solid-principles-dependency-inversion-1b0f#:~:text=The%20Dependency%20Inversion%20Principle%20(DIP,Details%20should%20depend%20upon%20abstractions.) ), that implementation it's out of the scope of this task for sure.
FYI I think there were some conversations around deprecating Facebook as login method, but nothing conclusive yet, so it's ok for now to not have coverage on those areas yet.

PR #1656, opened 2022-07-18
PAY-1760: create setup intent mutation

Arkariang:
# 📲 What

When loading the pledge screen, in case there is a logged in user, a network call to the CreateSetupIntent mutation will occur, giving back the information we need in next stories to present the paymentSheet

# 🤔 Why
- Present stripe's paymentSheet 

# 👀 See
DataDog trace 
Screen Shot 2022-07-18 at 3 45 46 PM

|  |  |

# 📋 QA

- Add your deviceID to the optimizely audience attached to the Android_PaymentSheet feature flag
- When pressing the new card button nothing will happen if user logged in, but if you analyze datadog or the network profiler you will see the call to the new mutation

# Story 📖

[NTV-1760](https://kickstarter.atlassian.net/browse/PAY-1760)


leighdouglas (approved):
awesome! 🎉

PR #1679, opened 2022-08-23
WEB-699: Android 13 Push Notification Permissions on bootstrap

leighdouglas:
# 📲 What

With android 13, we now request permission for push notifications from the user upon app install. More information here: https://developer.android.com/guide/topics/ui/notifiers/notification-permission

# 🤔 Why

With the launch of android 13, push notifications will now require requesting permissions from the user. This PR addresses this. 

# 🛠 How

Users will see the push permission request on bootstrap if:
- They are logged in
- They have not granted permission, and
- If they have not seen the permission request previously
- They are running android 13 or 12 on their phone (if our app points to api 13 but the users phone is still 12, the permissions request is not controlled by us but automatically sent by the system)

This is accomplished by checking if the user is logged in, checking if they have granted permission, and checking a boolean stored in shared preferences that flags if they have seen the dialog already. This is set to true regardless of if the user taps allow, deny, or swipes the dialog away without selecting an answer. 

I also converted the `DiscoveryViewModelTest` from java to kotlin
# 👀 See

Login allowed:

https://user-images.githubusercontent.com/19390326/186230732-12d4e39c-78b4-45e7-a746-e16fc1acb3be.mp4

# 📋 QA

- When user logs in after a fresh install, they should see the dialog.
- If you are just updating your app and it is not a fresh install, and the user has already granted notifications permissions on previous app versions, that should remain the same and no dialog should be shown
- If the user taps allow, they should receive notifications
- If the user taps deny, they should not receive notifications
- If the user swipes away, they should not receive notifications
- One the user has seen the dialog once, they should not see it again until they reinstall the app

Permissions can be cross checked in settings

# Story 📖

[WEB-699: Android 13 Push Notification Permissions on bootstrap](https://kickstarter.atlassian.net/browse/WEB-699)


Arkariang (approved):
Nice work 🔥 !

PR #1680, opened 2022-08-26
WEB-665   [Android] When Facebook-only users log in to Kickstarter, require them to set a password

hadia:
# 📲 What
When Facebook-only users log in to Kickstarter, require them to set a password

# 🤔 Why

Remove Facebook Login

# 🛠 How

Behind Feature Flag
As a User who does not have a Kickstarter password but can log in via Facebook
When I attempt to log in to Kickstarter using Facebook
I should be directed to set a new password
And if I try to navigate away without setting a password
I should be continually redirected to the set a password page until I have set one.

When I have successfully set a password, I should be directed as normal to the home page.

# 👀 See

https://user-images.githubusercontent.com/1075310/187357864-21ac9add-cdb3-4d68-93ef-25697d7bd39a.mp4

# 📋 QA

1. Add user Device id for feature flag "ANDROID_FACEBOOK_LOGIN_REMOVE" For production 
2. Remove the Facebook account Password with help from the "Web team" from database
3. Login with Facebook account and it would be nice if we test with 2FA and without 2FA
4. Check the app redirect to set Password Screen With user email masked
5. Kill app And reopen
6. Test Password different combination 
7. Set password and submit 
8. Check the app working normally  and can pledge projects 

 

# Story 📖

[[WEB-665]](https://kickstarter.atlassian.net/browse/WEB-665)

[WEB-665]: https://kickstarter.atlassian.net/browse/WEB-665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/res/values/strings.xml
```diff
@@ -84,8 +84,10 @@
 
   <!-- facebook remove -->
   <string name="FPO_reset_your_password">Reset your password</string>
-  <string name="FPO_reset_your_password_hint">FPO We’ve discontinued logging in through Facebook.  To access your Kickstarter account, enter the email associated to your Facebook account and we’ll send you a link to reset your password.</string>
-  <string name="FPO_reset_your_password_dialog_msg">FPO: We can no longer log you in through Facebook. Please log in with your Kickstarter password, or set a new password with your Facebook email.</string>
-  <string name="FPO_login_with_kickstarter">Log in</string>
+  <string name="FPO_we_discontinued_logging_in_through_Facebook">FPO We’ve discontinued logging in through Facebook.  To access your Kickstarter account, enter the email associated to your Facebook account and we’ll send you a link to reset your password.</string>
```
Arkariang, line 87: Quick question here, are the translations gonna be worked out in other story?

app/src/main/res/layout/activity_set_password.xml
```diff
@@ -0,0 +1,94 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:background="@color/window_background"
+    android:gravity="center_horizontal"
+    android:orientation="vertical">
+
+    <LinearLayout
```
Arkariang, line 10: We should try to avoid using containers, and go over `ConstraintLayout` I assume you are re-using some UI from other screen for velocity sake, but let's keep that in mind.

app/src/main/java/com/kickstarter/models/User.kt
```diff
@@ -379,6 +387,7 @@ class User private constructor(
                 unreadMessagesCount() == obj.unreadMessagesCount() &&
                 unseenActivityCount() == obj.unseenActivityCount() &&
                 weeklyNewsletter() == obj.weeklyNewsletter()
+            needsPassword() == obj.needsPassword()
```
Arkariang, line 390: missing the "&&" operator, that new line is not gonna be executed

Arkariang (approved):
Left a few comments, as soon as the equals operator for the user model is fixed feel free to merge. 
Really nice work @hadia

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -105,3 +106,27 @@ fun String.isGif(): Boolean {
     val gifPattern = "(?:\\/\\/.*\\.(?:gif))"
     return gifPattern.toRegex().find(this) != null
 }
+
+/**
```
Arkariang, line 110: 💖

app/src/main/res/values/strings.xml
hadia: yes

app/src/main/java/com/kickstarter/models/User.kt
hadia: Fixed

PR #1694, opened 2022-09-29
PAY-1988:  Fixed 3DS validation error

Arkariang:
# 📲 What

- Error during 3DS validation flow

# 🤔 Why

- `SavePaymentMethod` mutation was being called before the validation process was completed by the user, when receiving the user input through `onPaymentOption` there was no warrant for the 3DS validation process to be completed. To be completely sure the validation process is finalized `SavePaymentMethod` has being moved to `onPaymentSheetResult` ONLY when the result is completed.
- Added additional UI loading  starting from the moment where we have the payment method available, until some result is obtained from `onPaymentSheetResult`
- Differentiate between success state for deleting payment methods, and success state for saving payment method

# 👀 See
| Before 🐛 |
- Not saving the payment method, as it was attempting to save the payment methods once available, but not after having the complete Result

https://user-images.githubusercontent.com/4083656/193075692-b6c1f23d-13a2-4804-9e9c-5546c9ac4fbd.mp4

| After 🦋 |
- 3DS flow with success authentication, 3DS flow failing authentication

https://user-images.githubusercontent.com/4083656/193074473-15a1db57-ddfc-4911-9ede-93f8fc093e1f.mp4

| --- | --- |
|  |  |

# 📋 QA
- Try to add this payment method on user settings 4000002760003184 it always request 3DS validation, accept the validation, make sure the new payment method is added, and the black snackbar is presented
-  Try to add this payment method on user settings 4000002760003184 it always request 3DS validation, fail the validation, make sure no new payment methods is added and the red error snackbar is presented
- Add any other type of card nothing should have changed

# Story 📖

[PAY-1899](https://kickstarter.atlassian.net/browse/PAY-1988)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/PaymentMethodsSettingsActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/PaymentMethodsSettingsActivity.kt
@@ -97,7 +97,7 @@ class PaymentMethodsSettingsActivity : AppCompatActivity() {
         )
 
         compositeDisposable.add(
-            this.viewModel.success()
+            this.viewModel.successDeleting()
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
                 .subscribe { showSnackbar(binding.settingPaymentMethodsActivityToolbar.paymentMethodsToolbar, R.string.Got_it_your_changes_have_been_saved) }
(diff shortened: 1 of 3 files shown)


lukert33 (approved):
Thanks for the through videos! I can confirm that the before video shows the bug exactly as we encountered it during QA, and the after video shows it is thoroughly fixed!

hadia (approved):
Tested 😍

PR #1696, opened 2022-10-03
PAY-1931: Update stripe SDK

Arkariang:
# 📲 What

- Updating stripe version, they fixed a bug on the PaymentSheet when jumping from the top textfield to the next line. 

|  |  |

# 📋 QA

- Add a new paymentMethod on user settings
- Delete a payment method on user settings


hadia (approved):
👍

PR #1698, opened 2022-10-05
PAY-1870:  Update pledge flows with PaymentSheet

Arkariang:
# 📲 What

- Stripe's PaymentSheet added to the pledge and update pledge flows
- ℹ️ Additionally removed the codecoverage integration as currently failing showing 0%, and the internal builds keep failing at this point @hadia or @leighdouglas feel free to create a ticket or investigate while I'm on PTO

# 👀 See
- 103 tests Ok for `PledgeFragmentViewModel`
Screen Shot 2022-10-05 at 11 04 55 AM

- Pledge & change payment method flows 

https://user-images.githubusercontent.com/4083656/194020961-b1877da1-7a73-4a8d-a8bf-1c1d9f09ba8a.mp4

|  |  |

# 📋 QA

⚠️  remember that for being able to use the update flows the project needs to be still going (Screenshot attached), meaning a project with "0 seconds to go" is no valid. 
Screen Shot 2022-10-05 at 10 54 46 AM

- Pledge to a project
- Update a pledge
- Change a reward
- Change payment method -> add a new payment method using the PaymentSheet
- Try to Fix a pledge ⚠️ tested on production with my personal account, but having the same situation as described below 👇 

@lukert33 or @iambchan in the video you can see as well that after successfully changing the payment method (calling UpdateBacking mutation) the backing information seems to not be updated with the new payment method, or it added to the user payment methods either.

# Story 📖

[PAY-1870](https://kickstarter.atlassian.net/browse/PAY-1870)


hadia (approved):
Tested😍

lukert33:
@Arkariang good catch on that bug! I confirmed that for me, on production using the legacy form, updating my card to a new payment source works. 

I also confirmed that on web, the flow works using the new form but web does not attempt to display the updated payment source. When I reload the edit pledge form, the new payment source is applied which I can see is not the case in your video.

So this looks like a blocking bug that we can create a story for. Can you write a story for the bug, including the video and a link to the network request?

PR #1700, opened 2022-10-12
WEB-572  : [Android] Deleted feature flag for Android Comment Moderation epic

hadia:
# 📲 What

 Deleted feature flag for Android Comment Moderation 

# 📋 QA
Comments should all work as they did before, tests all states between project and update comments.
Test all states of flagged comments:

1. user is not author: Flagged comment
2. user is author: Normal comment
3. Back project -> Add comments and reply for project and updates -> Cancel a pledge

# Story 📖
https://kickstarter.atlassian.net/browse/WEB-572

### DIFF
--- a/app/src/test/java/com/kickstarter/viewmodels/CommentsViewHolderViewModelTest.kt
+++ b/app/src/test/java/com/kickstarter/viewmodels/CommentsViewHolderViewModelTest.kt
@@ -3,8 +3,6 @@ package com.kickstarter.viewmodels
 import com.kickstarter.KSRobolectricTestCase
 import com.kickstarter.libs.Environment
 import com.kickstarter.libs.MockCurrentUser
-import com.kickstarter.libs.models.OptimizelyFeature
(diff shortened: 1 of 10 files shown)


app/src/test/java/com/kickstarter/viewmodels/CommentsViewHolderViewModelTest.kt
```diff
@@ -598,20 +573,15 @@ class CommentsViewHolderViewModelTest : KSRobolectricTestCase() {
 
     @Test
     fun testCommentsViewModel_whenCommentModFFDisabled_shouldNotSetStatusToFlagged() {
```
leighdouglas, line 575: We should probably change the name of this test since it's not testing for the feature flag being enabled/disabled anymore.

leighdouglas (approved):
Left a small comment about changing the name of a test, but otherwise looks good to me! 🎉

PR #1736, opened 2023-01-31
WEB-858: [Android] Check consent preference in isEnabled method in the TrackingClient

leighdouglas:
# 📲 What

Now that we can save user consent preference, we want to check for that preferences and not send events to segment if the user has declined consent. 

# 🤔 Why

[Consent management](https://kickstarter.atlassian.net/wiki/spaces/NEW/pages/2227077121/Android+Consent+Management)

# 🛠 How

Check if the feature flag is enabled, and then if the user has set a consent preference. If they have not, we default to false until they have consented. 

# 👀 See

There are no user facing changes, but segment events should not have any identifying information if the user has denied consent or if they have not given consent yet. 

# 📋 QA

- Open app, check segment dashboard. Events should not fire or have any identifying information
- Choose "Allow", segment dashboard should show events
- Clear data from your app and reopen the app, consent management dialog should reappear.
- Choose "Deny", segment dashboard not show events besides the initial application opened event

These QA tests should work the same regardless of if a user is logged in or not

# Story 📖

[WEB-858: [Android] Check consent preference in isEnabled method in the TrackingClient](https://kickstarter.atlassian.net/browse/WEB-858)


hadia (approved):
👌

Arkariang:
@leighdouglas Have you tried, when choosing "Deny", the identify events when opting in-out of notification preferences do not get send either?
To opting in-out of notification preferences Drawer menu -> settings -> Notifications

PR #1754, opened 2023-03-15
WEB-967: [Android] Create FirebaseAnalyticsClient and FirebaseAnalyticEvents, and add instances in application model and environment

leighdouglas:
# 📲 What

Set up infrastructure for sending google analytic events and properties on android

# 🤔 Why

- Implementation of GA4

# 🛠 How

- `FirebaseAnalyticEvents` will act as an proxy to send the correctly formatted data (in the form of bundles) and event keys to `FirebaseAnalyticsClient`
- FirebaseAnalyticsClient will sends these events to firebase and check for user consent

# 👀 See

We can view which events are being sent via logcat (this is a test event with my name that i sent from a project screen): 
Screen Shot 2023-03-14 at 6 58 10 PM

# 📋 QA

There are currently no manually sent events, just autotracked ones. For future QA purposes, and to also see the auto tracked events, follow these directions. 

- Make sure your device or emulator is connected/debug turned on
- In the console, type the commands `adb shell setprop log.tag.FA VERBOSE` and `adb shell setprop log.tag.FA-SVC VERBOSE`
- In android studio, click the debug button next to the run arrow
- After the app launches, in logcat filter/search by FA-SVC
- You should see the automatically tracked events
- This is recommended as google analytics will batch events to be uploaded at a later time, so more often than not you will not be able to see them immediately in the console
- more information about debugging/testing here: https://www.youtube.com/watch?v=aaQoqGELrQ4&t=349s

# Story 📖

[WEB-967: [Android] Create FirebaseAnalyticsClient and FirebaseAnalyticEvents, and add instances in application model and environment](Trello link)


app/src/main/java/com/kickstarter/libs/FirebaseAnalyticsClient.kt
```diff
@@ -0,0 +1,24 @@
...
+import android.os.Bundle
+import com.google.firebase.analytics.FirebaseAnalytics
+import com.kickstarter.libs.models.OptimizelyFeature
+import com.kickstarter.ui.SharedPreferenceKey.CONSENT_MANAGEMENT_PREFERENCE
+
+open class FirebaseAnalyticsClient(
+    private var optimizely: ExperimentsClientType,
+    private var preference: SharedPreferences,
+    private val firebaseAnalytics: FirebaseAnalytics?,
+) {
```
Arkariang, line 13: You could have created here an interface having two methods { track, isEnabled} in a facade way. And add to the Dagger graph, the dependency with the interface type.

Arkariang (approved):

PR #1778, opened 2023-05-18
MBL-700 : Fix bonus amount not being able to go past 3 digits

mtgriego:
# 📲 What

Fixes an issue where the bonus amount on pledge pages could not pass 3 digits.  This makes it so it will allow any normal formatting from the locale of the device

# 🤔 Why

You could not pledge up to the limit before

# 🛠 How

Removed the use of deprecated calls and updated formatting/parsing to use standard libraries

# 👀 See

| Before (ANY) 🐛 | After USA 🦋 | After Germany | After France |
| --- | --- | --- | --- |
| https://github.com/kickstarter/android-oss/assets/7563288/fc5202f0-b8b4-405a-a058-84c60e4c81f8 | https://github.com/kickstarter/android-oss/assets/7563288/58d4c723-a121-4adb-9c8c-72de97c8a373 | https://github.com/kickstarter/android-oss/assets/7563288/a9d30296-8715-4153-8767-c07c9da19bf7 | https://github.com/kickstarter/android-oss/assets/7563288/889428a4-bf1f-4a5f-8b44-1f8cb130fa72 |

# 📋 QA

1. Set your location to wherever you want
2. Go through the pledge of a campaign and either 
- don't select an award and pledge
- select an award and proceed to the checkout page
3) Start entering amounts in bonus/pledge

You should see the number being formatted as you enter it and it should allow you to add decimals (though like web it does not limit it to 2 digits)

# Story 📖

[MBL-700](https://kickstarter.atlassian.net/browse/MBL-700)

[MBL-700]: https://kickstarter.atlassian.net/browse/MBL-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/NumberUtils.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/NumberUtils.java
@@ -109,6 +109,9 @@ public final class NumberUtils {
     return value == Math.round(value);
   }
 
+  /**
+   * @deprecated Use com.kickstarter.libs.utils.extensions.StringExt#parseToDouble(java.lang.String)
+   */
   @Deprecated
   public static double parse(final @NonNull String input) {
     return parse(input, Locale.getDefault());
--- a/app/src/main/java/com/kickstarter/libs/utils/NumberUtils.java
+++ b/app/src/main/java/com/kickstarter/libs/utils/NumberUtils.java
@@ -109,6 +109,7 @@ public final class NumberUtils {
     return value == Math.round(value);
   }
 
+  @Deprecated
   public static double parse(final @NonNull String input) {
     return parse(input, Locale.getDefault());
   }
--- a/app/src/main/res/layout/fragment_pledge_section_bonus_support.xml
+++ b/app/src/main/res/layout/fragment_pledge_section_bonus_support.xml
@@ -129,6 +129,7 @@
             android:imeOptions="actionDone"
             android:importantForAutofill="no"
             android:inputType="numberDecimal"
+            android:digits="0123456789.,"
             android:maxLength="@integer/max_length"
             tools:ignore="LabelFor"
             tools:targetApi="o"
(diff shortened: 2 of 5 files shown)


mtgriego:
There is an issue with our backend currently, but will provide examples asap

app/src/main/res/layout/fragment_pledge_section_bonus_support.xml
```diff
@@ -129,6 +129,7 @@
             android:imeOptions="actionDone"
             android:importantForAutofill="no"
             android:inputType="numberDecimal"
+            android:digits="0123456789.,"
```
mtgriego, line 132: This allows the user to enter both `.` and `,` characters, before it would only accept `.`

app/src/main/java/com/kickstarter/libs/utils/NumberUtils.java
```diff
@@ -109,6 +109,10 @@ private static Boolean isWholeNumber(final @NonNull Double value) {
     return value == Math.round(value);
   }
 
+  /**
+   * @deprecated Use com.kickstarter.libs.utils.extensions.StringExt#parseToDouble(java.lang.String)
+   */
+  @Deprecated
```
mtgriego, line 115: The functions this calls is deprecated, so this should be marked deprecated was my approach

app/src/main/res/layout/fragment_pledge_section_bonus_support.xml
ycheng-kickstarter: Do we want to accept , though? I noticed it's disabled on web, plus it doesn't behave the way a EU person would expect it to behave. Might as well just disable it?

ycheng-kickstarter: QA note: must change device language too!

ycheng-kickstarter (approved):

PR #1779, opened 2023-05-23
[MBL-836] Remove unneeded extension function, replace with core library usage

mtgriego:
# 📲 What

Removing a written extension function to replace it with the core library functionality

# 🤔 Why

We got the library, might as well use it

# 🛠 How

A keyboard and some effort

# 📋 QA

Visit the various pages/views affected by the change and confirm everything works as expected

# Story 📖

[MBL-836](https://kickstarter.atlassian.net/browse/MBL-836)

[MBL-836]: https://kickstarter.atlassian.net/browse/MBL-836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/libs/utils/extensions/ViewExt.kt
```diff
@@ -1,36 +0,0 @@
...
-    visibility = if (hidden) {
-        View.INVISIBLE
-    } else {
-        View.VISIBLE
-    }
-}
-
-fun View.setInvisible(): Action1<Boolean> {
-    return Action1 { invisible: Boolean -> this.setInvisible(invisible) }
-}
```
mtgriego, line 36: Only used here, not in any code elsewhere

mtgriego:
Also removed a couple unnecessary `@NotNull` annotations, kotlin does this for us as long as we don't add a `?`

Arkariang (approved):

PR #1794, opened 2023-06-07
[MBL-830] Update UpdateViewmodel/Activity/Tests to use RX2

mtgriego:
# 📲 What

See title

# 🤔 Why

Stuffs out of date, bring it up to date-ish

# 🛠 How

See code

# 👀 See

No visible changes

# 📋 QA

Go to the update page via these methods: Activity feed, Project page, Discovery page update, Push Notification

# Story 📖

[MBL-830](https://kickstarter.atlassian.net/browse/MBL-830)

[MBL-830]: https://kickstarter.atlassian.net/browse/MBL-830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/UpdateActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/UpdateActivity.kt
@@ -137,11 +137,6 @@ class UpdateActivity : AppCompatActivity() {
         viewModel.provideIntent(intent)
     }
 
-    override fun onNewIntent(intent: Intent?) {
-        super.onNewIntent(intent)
-        viewModel.provideIntent(intent)
-    }
-
     override fun onDestroy() {
         super.onDestroy()
         binding.updateWebView.setDelegate(null)


app/src/main/java/com/kickstarter/viewmodels/UpdateViewModel.kt
```diff
@@ -189,46 +220,61 @@ interface UpdateViewModel {
...
+                .subscribe {
+                    it?.let { pair ->
+                        startShareIntent.onNext(pair)
+                    }
+                }
+                .addToDisposable(disposables)
 
             goToCommentsActivity
-                .delay(2, TimeUnit.SECONDS, environment.scheduler()) // add delay to wait until activity subscribed to viewmodel
+                .delay(
```
Arkariang, line 231: Probably this delay can now be removed, as the view model is created now in a different way

app/src/main/java/com/kickstarter/ui/intentmappers/ProjectIntentMapper.kt
```diff
@@ -90,6 +92,30 @@ object ProjectIntentMapper {
             .mergeWith(projectFromParceledParam)
     }
 
+    /**
+     * Returns an observable of projects retrieved from intent data. May hit the API if the intent only contains a project
+     * param rather than a parceled project.
+     */
+    fun project(intent: Intent, client: ApiClientTypeV2): io.reactivex.Observable<Project> {
```
Arkariang, line 99: could we add the test for V2  on `ProjectIntentMapperTest`

app/src/main/java/com/kickstarter/viewmodels/UpdateViewModel.kt
```diff
@@ -266,36 +319,50 @@ interface UpdateViewModel {
             return Pair.create(projectParam, updateParam)
         }
 
-        override fun externalLinkActivated() = externalLinkActivated.onNext(null)
+        override fun externalLinkActivated() = externalLinkActivated.onNext(Unit)
```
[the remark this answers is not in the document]
mtgriego: I shall remove it then

app/src/main/java/com/kickstarter/ui/intentmappers/ProjectIntentMapper.kt
mtgriego: you got it!

app/src/main/java/com/kickstarter/ui/activities/UpdateActivity.kt
```diff
@@ -61,105 +74,112 @@ class UpdateActivity : BaseActivity<UpdateViewModel.ViewModel?>(), KSWebView.Del
...
+            .addToDisposable(disposables)
 
         binding.updateActivityToolbar.shareIconButton.setOnClickListener {
             viewModel.inputs.shareIconButtonClicked()
         }
+
+        viewModel.provideIntent(intent)
+    }
+
+    override fun onNewIntent(intent: Intent?) {
```
ycheng-kickstarter, line 142: Noticed this new logic - can you explain what it does? Is this activity supposed to be a single instance?

mtgriego: So before the activity was providing the intent to the viewmodel via a class function in BaseActivity, but since we are not using that for this anymore i just copied its functionality to this activity/viewmodel.  it just provides the intent to the viewmodel so it can grab its arguments

ycheng-kickstarter: From what I can see in documentation onNewIntent is for activities that are set to "singleTop", and it doesn't look like this activity is set to singleTop in the AndroidManifest, so I was curious if there is some other use of onNewIntent that we're using it for here?

mtgriego: nop[e, and if thats the case I can just remove it, only realy did this because its how the previous activity was doing it and I didnt want to rock the boat too much

PR #1797, opened 2023-06-12
[MBL-857] Add feature flag to remove dashboard feature

mtgriego:
# 📲 What

Added a feature flag that will hide the dashboard feature when set to true

# 🤔 Why

We will be removing this feature in the future so we want to be able to turn it off from this version onward before removal

# 🛠 How

Added the variable in the remote config and read that value where needed in the app

# 📋 QA

1) Log in as a creator
2) turn the feature flag on/off depending on what you are testing
3) Check to see if the entry point on both the discovery menu and the project page are gone when the flag is set to true, there when set to false

# Story 📖

[MBL-857](https://kickstarter.atlassian.net/browse/MBL-857)

[MBL-857]: https://kickstarter.atlassian.net/browse/MBL-857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ycheng-kickstarter (approved):
LGTM. Changes on the console took about 10 min for me fyi.

PR #1802, opened 2023-06-26
MBL-868 crash account

ycheng-kickstarter:
# 📲 What

Per Isa's doc https://docs.google.com/document/d/1ib2SodBYjbDXK_-Sx69ZfT64tIhLMhjVJ8JRrH-QgSQ/edit, a fix resolve a crash happening for null user properties. Rather than directly exposing the UserPrivacy GraphQL response which could contain null values, first convert to a UserPrivacy data class with default values before exposing it to ViewModels to consume.

Most important classes to look at are 
UserPrivacy.kt, 
KSApolloClientV2.kt, 
GraphQLTransformers.kt. 

All other changes are just refactors to accept type `UserPrivacy` rather than `UserPrivacyQuery.Data` from the apollo client.

# 🤔 Why

Users with null UserPrivacy properties will crash in 3.12.0 because of all the RxJava1->RxJava2 migrations that require non-null values.

# 🛠 How

Added a test in GraphQLTransformersTest to verify that a response with null chosenCurrency gets transformed into a data class with "USD" as default value.

# 👀 See

No visible changes.

# 📋 QA

Create a NEW account to ensure you definitely haven't set a chosenCurrency. (Do NOT test with therealnativesquad account for instance as that has a non-null chosenCurrency) View your Account. Ensure no crash, and verify that the default currency listed is $USD.

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-868


app/external_version_name.txt
```diff
@@ -1 +1 @@
-3.11.0
\ No newline at end of file
+3.12.0
```
ycheng-kickstarter, line 1: Note this was branched off 3.12.0, not master

app/src/main/java/com/kickstarter/models/UserPrivacy.kt
```diff
@@ -0,0 +1,11 @@
+package com.kickstarter.models
+
+data class UserPrivacy(
```
ycheng-kickstarter, line 3: Don't take the User query response and directly pass values from it around app. Create a new data class that the User query response must get transformed to before passing it around app.

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
```diff
@@ -236,18 +238,29 @@ class KSApolloClientV2(val service: ApolloClient) : ApolloClientTypeV2 {
...
+                        response.data?.me()?.let {
+                            val defaultCurrency = "USD"
+                            val userPrivacy = UserPrivacy(
+                                name = it.name(),
+                                email = it.email() ?: "",
+                                hasPassword = it.hasPassword() ?: false,
+                                isCreator = it.isCreator ?: false,
+                                isDeliverable = it.isDeliverable ?: false,
+                                isEmailVerified = it.isEmailVerified ?: false,
+                                chosenCurrency = it.chosenCurrency() ?: defaultCurrency
```
ycheng-kickstarter, line 260: For non-mandatory fields, include a default value so that it is never null.

app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
```diff
@@ -417,7 +417,7 @@ fun userTransformer(user: fragment.User?): User {
     val avatar = Avatar.builder()
         .medium(user?.imageUrl())
         .build()
-    val chosenCurrency = user?.chosenCurrency()
+    val chosenCurrency = user?.chosenCurrency() ?: "USD"
```
ycheng-kickstarter, line 420: Also fixed this chosenCurrency in the userTransformer that didn't have a default value

app/src/main/java/com/kickstarter/viewmodels/AccountViewModel.kt
```diff
@@ -72,18 +73,18 @@ interface AccountViewModel {
                 .compose(Transformers.neverErrorV2())
 
             userPrivacy
-                .map { it.me()?.chosenCurrency() }
+                .map { chosenCurrency() }
```
Arkariang, line 76: Now this RXJava chain can be exchanged over to: 
` userPrivacy
                .map { it.chosenCurrency }
                .subscribe { this.chosenCurrency.onNext(it) }
                .addToDisposable(disposables)`

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
Arkariang: Small suggestion, no need to change it if you don't see the need for it. but we could move this piece into `fun userPrivacyTransformer(UserPrivacyQuery.Data): UserPrivacy` within `GraphQLTransformers` and add some isolated unit tests on `GraphQLTransformersTest`

app/src/main/java/com/kickstarter/viewmodels/AccountViewModel.kt
ycheng-kickstarter: Ah ok, yeah I wasn't sure what the ObjectUtils.coalesce was doing. Are you sure you want me to remove ```           .map { ObjectUtils.coalesce(it, CurrencyCode.USD.rawValue()) }```

Arkariang: yeap, it's redundant now, if you look at `ObjectUtils,coalesce` internal code it basically adds the second parameter as default value if the first parameter is null

app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
Arkariang: probably is better to use instead of "USD", CurrencyCode.USD.rawValue() as default value, is not a value likely gonna change but to not have hardcoded strings around if avoidable

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
Arkariang: for default currency here, same as below comment, let's use CurrencyCode.USD.rawValue()

ycheng-kickstarter: Yup! I had made same comment in your doc about moving logic to GraphQLTransformers. Will do.

Arkariang (dismissed):
amazing work! 💖

app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
```diff
@@ -427,6 +429,24 @@ fun userTransformer(user: fragment.User?): User {
...
+    return UserPrivacy(
+        name = userPrivacy.name(),
+        email = userPrivacy.email() ?: "",
+        hasPassword = userPrivacy.hasPassword() ?: false,
+        isCreator = userPrivacy.isCreator ?: false,
+        isDeliverable = userPrivacy.isDeliverable ?: false,
+        isEmailVerified = userPrivacy.isEmailVerified ?: false,
+        chosenCurrency = userPrivacy.chosenCurrency() ?: defaultCurrency
+    )
+}
```
leighdouglas, line 448: 🎉

leighdouglas (dismissed):
lgtm! 🎉

Arkariang (approved):

PR #1811, opened 2023-07-10
[MBL-871] Add compose visuals based on design system

mtgriego:
# 📲 What

Added many default views to be used in the future, added a screen where you can preview all the default views and typography 

# 🤔 Why

We need to prepare for our future visuals building

# 🛠 How

Compose and a bit of learning

# 👀 See

In App on internal builds: Go to internal tools and design system

# Story 📖

[MBL-871](https://kickstarter.atlassian.net/browse/MBL-871)

[MBL-871]: https://kickstarter.atlassian.net/browse/MBL-871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/ui/extensions/SafeLet.kt
```diff
@@ -0,0 +1,24 @@
+package com.kickstarter.ui.extensions
+
+inline fun <T1 : Any, T2 : Any, R : Any> safeLet(p1: T1?, p2: T2?, block: (T1, T2) -> R?): R? {
```
Arkariang, line 3: I love this! but a question, could we user varargs to reduce the amount of `safeLet` declarations?

Arkariang:
We could probably add to `jacoco.gradle` any new or modified folder containing only UI, that or create a spike ticket to explore UI testing with compose. The only UI testing we currently have it's screenshot testing and it's not adapted to compose yet.

Arkariang:
Small comments but overall looks fantastic!

Arkariang (dismissed):

app/src/main/java/com/kickstarter/ui/compose/designsystem/KSControls.kt
```diff
@@ -0,0 +1,223 @@
...
+            checkmarkColor = colors.kds_white,
+            disabledColor = kds_support_500,
+        ),
+        enabled = enabled
+    )
+}
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun KSStringDropdown(
```
Arkariang, line 159: A bit worried about the `ExperimentalMaterialApi` for dropdowns, but one of the first UI pieces I would like to kill and substitute over this component will be the Shipping rules dropdown. The usability will dramatically improve.

app/src/main/java/com/kickstarter/ui/extensions/SafeLet.kt
mtgriego: I am not sure, but I'll see what I can find!

app/src/main/java/com/kickstarter/ui/compose/designsystem/KSControls.kt
mtgriego: Yeah this was the only one that needed it fortunately.  The non experimental one didn't seem to match what we needed so I opted for this, but if something better replaces it all the better

app/src/main/java/com/kickstarter/ui/extensions/SafeLet.kt
mtgriego: from what i can tell I can do a vararg, but the problem is that seems to make it so all the elements need to be the same type, whereas this allows each thing being passed to be whatever it wants to be as long as its not null.  ill keep digging on that when i can, but for now I will push this up

Arkariang (dismissed):

Arkariang (approved):

PR #1820, opened 2023-07-24
[NoTicket] Adjust usage and visuals of compose screen based on learnings

mtgriego:
# 📲 What

- Updated KS Alerts to have actual snackbars and name components accordingly
- Use scaffold for changepassword screen
- move topbar into scaffold
- use new snackbars for the snackbar message

# 🤔 Why

Had to do custom work to make current snackbar show at the top, but we will be moving forward with android standards where we can, so started with that here

# 🛠 How

- Added keyboard options to compose ks input texts
- created new snackbars for ks alerts
- updated logic on change password screen to use error as a state

# 👀 See

https://github.com/kickstarter/android-oss/assets/7563288/6277c9e8-26bf-48ba-8add-0849f531be07

# Story 📖

None/My Head/Slack convos

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ChangePasswordActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ChangePasswordActivity.kt
@@ -2,11 +2,11 @@ package com.kickstarter.ui.activities
 
 import android.content.Intent
 import android.os.Bundle
-import android.os.Handler
-import android.os.Looper
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.viewModels
+import androidx.compose.material.rememberScaffoldState
+import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/ui/activities/ChangePasswordActivity.kt
```diff
@@ -57,21 +60,18 @@ class ChangePasswordActivity : ComponentActivity() {
...
-                .subscribe {
-                    errorMessage = it
-                    if (it.isNotEmpty()) {
-                        Handler(Looper.getMainLooper()).postDelayed({
-                            viewModel.resetError()
-                        }, 2750)
+            when {
+                error.isNotEmpty() -> {
+                    LaunchedEffect(scaffoldState) {
+                        scaffoldState.snackbarHostState.showSnackbar(error)
```
Arkariang, line 70: 💖

Arkariang (approved):

PR #1828, opened 2023-08-03
[Master] update version codes/names

mtgriego:
# 📲 What

Updating the version codes and names based on stable release 3.13.0

# 🤔 Why

3.13 is stable so updating master


app/external_version_code.txt
```diff
@@ -1 +1 @@
-2014150878
\ No newline at end of file
+2014150879
```
Arkariang, line 1: Take the version number, from firebase, CI will increase it by 1
Screenshot 2023-08-09 at 9 56 15 AM

Arkariang (approved):

PR #1830, opened 2023-08-09
MBL-798: SetPasswordViewModel.kt to Jetpack ViewModel and RxJava2.2

leighdouglas:
# 📲 What

# 🤔 Why

# 👀 See

No user facing changes

# 📋 QA

This feature is tricky to test - I had to manually set the `needsPassword` field in the User struct to `true`. I then had to make sure the feature flag for facebook login deprecation was on or the condition was commented out in the discovery view model. This is what made the set password screen appear after logging in.

On the actual set password screen: 
Make sure that the warning text changes if the passwords in each field do not match, or if the password in the first field is fewer than 6 characters. 

To test success or failure in setting the password, some work is required from the backend if you want to QA this

# Story 📖

[MBL-766: FrequentlyAskedQuestionsViewHolderViewModel.kt to Jetpack ViewModel and RxJava2.2](https://kickstarter.atlassian.net/browse/MBL-766)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/SetPasswordActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/SetPasswordActivity.kt
@@ -16,21 +16,22 @@ import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.android.schedulers.AndroidSchedulers
 
 class SetPasswordActivity : AppCompatActivity() {
+    private lateinit var viewModelFactory: SetPasswordViewModel.Factory
     private val viewModel: SetPasswordViewModel.SetPasswordViewModel by viewModels { viewModelFactory }
     private lateinit var binding: ActivitySetPasswordBinding
     private var errorTitleString = R.string.general_error_oops
-    private lateinit var viewModelFactory: SetPasswordViewModel.Factory
     private val disposables = CompositeDisposable()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
         this.getEnvironment()?.let { env ->
             viewModelFactory = SetPasswordViewModel.Factory(env)
         }
-        viewModel.configureWith(intent)
 
         binding = ActivitySetPasswordBinding.inflate(layoutInflater)
 
+        viewModel.configureWith(intent)
         setContentView(binding.root)
         setSupportActionBar(binding.resetPasswordToolbar.loginToolbar)
         binding.resetPasswordToolbar.loginToolbar.setTitle(getString(R.string.Set_your_password))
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/viewmodels/SetPasswordViewModel.kt
```diff
@@ -93,65 +99,69 @@ interface SetPasswordViewModel {
 
             setNewPassword
                 .map { it.warning() }
+                .filter { ObjectUtils.isNotNull(it) }
+                .map { requireNotNull(it) }
+                .filter { it != 0 }
```
leighdouglas, line 104: I had to update the function `warning()` to return a non-null value, as it was still causing the app to crash due to a the mapper function returning a null value, even with the `isNotNull` checks in the chain. I'm not exactly sure why this is the case here or what makes this different, as opposed to other places where a null check in the chain is enough to not cause a crash (the intent rx chain above is an example of this, where the getStringExtra returns a null value if there is no extra with that key attached to the intent, but it does not crash.)

app/src/main/java/com/kickstarter/ui/activities/SetPasswordActivity.kt
```diff
@@ -34,51 +44,52 @@ class SetPasswordActivity : BaseActivity<SetPasswordViewModel.ViewModel>() {
...
 
         this.viewModel.outputs.setUserEmail()
-            .compose(bindToLifecycle())
-            .compose(Transformers.observeForUI())
+            .observeOn(AndroidSchedulers.mainThread())
             .subscribe {
-                binding.setPasswordHint.text = this.environment().ksString()?.format(getString(R.string.We_will_be_discontinuing_the_ability_to_log_in_via_FB), "email", it)
-            }
+                binding.setPasswordHint.text = getEnvironment()?.ksString()?.format(getString(R.string.We_will_be_discontinuing_the_ability_to_log_in_via_FB), "email", it)
+            }.addToDisposable(disposables)
```
mtgriego, line 56: i know this was in the last iteration of this as well, but do we have a translated string for "email"?

mtgriego (approved):

app/src/main/java/com/kickstarter/ui/activities/SetPasswordActivity.kt
```diff
@@ -88,8 +99,4 @@ class SetPasswordActivity : BaseActivity<SetPasswordViewModel.ViewModel>() {
     private fun setFormDisabled(isDisabled: Boolean) {
         setFormEnabled(!isDisabled)
     }
-
-    override fun back() {
```
Arkariang, line 92: @leighdouglas I think the disable back action was on purpose, to not allow the user to leave the screen

Arkariang (approved):

app/src/main/java/com/kickstarter/ui/activities/SetPasswordActivity.kt
leighdouglas: Good catch, there is no back button, but a user can swipe or press the backbutton to exit. I'll bring this up during sprint planning and add this as a QA script.

app/src/main/java/com/kickstarter/ui/activities/SetPasswordActivity.kt
leighdouglas: We do! i can update in a subsequent PR. Thanks for catching this.

PR #1832, opened 2023-08-14
MBL-751: CreatePasswordViewModel.kt to Jetpack ViewModel and RxJava2.2

leighdouglas:
# 📲 What

# 🤔 Why

# 👀 See

No user facing changes

# 📋 QA

**one thing to note:** i updated the functionality that shows the warning text on the createpassword screen as well as the setpassword screen because the warning text was not disappearing when there was no longer any warning. Previously, for example, if the passwords fields did not match, the text would appear, but then if you changed the passwords to match, that warning text would not disappear. Now the function returns a 0 if there is no warning, and the warning text is set accordingly in the activity. 

To launch the create password screen:
I had to manually set the create password row in the account activity to visible so that i could tap the row and launch the activity. 

On the actual create password screen: 
Make sure that the warning text changes if the passwords in each field do not match, or if the password in the first field is fewer than 6 characters, and also disappears if there should not be a warning. 
**To test success or failure in setting the password, some work is required from the backend if you want to QA this**

This will require additional QA before it goes live. 

# Story 📖

[MBL-751: CreatePasswordViewModel.kt to Jetpack ViewModel and RxJava2.2](https://kickstarter.atlassian.net/browse/MBL-751)


app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
```diff
@@ -669,4 +670,32 @@ class KSApolloClientV2(val service: ApolloClient) : ApolloClientTypeV2 {
...
+                    .build()
+            )
+                .enqueue(object : ApolloCall.Callback<CreatePasswordMutation.Data>() {
+                    override fun onFailure(exception: ApolloException) {
+                        ps.onError(exception)
+                    }
+
+                    override fun onResponse(response: Response<CreatePasswordMutation.Data>) {
+                        if (response.hasErrors()) {
+                            ps.onError(java.lang.Exception(response.errors?.first()?.message))
```
Arkariang, line 690: `java.lang` can be removed

Arkariang (approved):

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
leighdouglas: okay, will update in subsequent PR!

PR #1843, opened 2023-08-23
MBL-977: Deprecate ObjectUtils, migrate to kotlin Ext functions

leighdouglas:
# 📲 What

Created an AnyExt class with equivalent extension functions to replace ObjectUtils.java, and removed unused or unneeded functions. 

# 🤔 Why

ObjectUtils was legacy and still in java.

# 🛠 How

Replace objectutils functions with ext functions

# 👀 See

No visible changes

# 📋 QA

This will need QA of all of our important functions in the app, as this touches almost every part of the app.

# Story 📖

[MBL-977: Deprecate ObjectUtils, migrate to kotlin Ext functions #1843
](https://kickstarter.atlassian.net/browse/MBL-977)


app/src/main/java/com/kickstarter/libs/utils/extensions/AnyExt.kt
```diff
@@ -0,0 +1,54 @@
...
+    return when (this) {
+        is Long -> this.toString()
+        is Float -> this.toString()
+        is Int -> this.toString()
+        is Double -> this.toString()
+        else -> null
+    }
+}
+
+fun <T : Any> T?.requireNonNull(): T {
```
leighdouglas, line 40: I think all of these requireNonNull methods can be removed in favor of `Preconditions.kt requireNotNull`

Arkariang: correct, kotlin already has a [`requireNotNull()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require-not-null.html#requirenotnull), as example take a look at `ProjectPageViewModel` every `requireNotNull` would be the `Preconditions` one, so for any kotlin code use kotlin available precondition and not  `ObjectUtils.requireNotNull()`. 

So I suggest a quick search over the project and exchange any  `ObjectUtils.requireNotNull()` over the precondition instead of the extension methods. And remove the extension methods and possible `ObjectUtils.requireNotNull()`  as well (We can keep the latest one in java code if needed).

In `ProjectCardViewHolder.kt` we could:
`
        val projectAndParams = requireNotNull(data as? Pair<Project, DiscoveryParams>)
`
instead of using the extension method: 
`
        val projectAndParams = (data as? Pair<Project, DiscoveryParams>).requireNonNull()
`

app/src/main/java/com/kickstarter/libs/utils/extensions/AnyExt.kt
```diff
@@ -0,0 +1,54 @@
...
+}
+
+fun <T : Any> T?.coalesce(theDefault: T): T {
+    if (this.isNotNull()) {
+        return this!!
+    }
+    return theDefault
+}
+
+fun <T : Any> coalesceWith(theDefault: T): Func1<T, T> {
```
Arkariang, line 22: This one is not used, probably could be removed.

app/src/main/java/com/kickstarter/libs/utils/extensions/AnyExt.kt
```diff
@@ -0,0 +1,54 @@
...
+    return Func1 { value: T -> ObjectUtils.coalesce(value, theDefault) }
```
Arkariang, line 23: Here substitute the `ObjectUtils.coalesce(value, theDefault)` over the `value.coalesce(theDefault)` to use the extension method on line 15

app/src/main/java/com/kickstarter/libs/utils/extensions/AnyExt.kt
```diff
@@ -0,0 +1,54 @@
+package com.kickstarter.libs.utils.extensions
+
+import com.kickstarter.libs.utils.ObjectUtils
+import io.reactivex.functions.Function
+import rx.functions.Func1
```
Arkariang, line 5: `Fun1` is rxJava1, `Function` is RXJava 2` a bit confused around the `coalesceWith` extension methods looks like they are transformers, but as they lived previously on `ObjectUtils` let's keep them for now.

app/src/main/java/com/kickstarter/libs/utils/extensions/AnyExt.kt
```diff
@@ -0,0 +1,54 @@
...
+    }
+    return theDefault
+}
+
+fun <T : Any> coalesceWith(theDefault: T): Func1<T, T> {
+    return Func1 { value: T -> ObjectUtils.coalesce(value, theDefault) }
+}
+
+fun <T : Any> coalesceWithV2(theDefault: T): Function<T, T> {
+    return Function { value: T -> ObjectUtils.coalesce(value, theDefault) }
```
Arkariang, line 27: same as line 23

app/src/main/java/com/kickstarter/libs/utils/extensions/AnyExt.kt
```diff
@@ -0,0 +1,54 @@
...
+}
+
+fun <T : Any> T?.numToString(): String? {
```
Arkariang, line 30: This extension method makes sense for the code still in java, otherwise could probably be remove

app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -180,3 +180,13 @@ fun String.toHashedSHAEmail(): String {
         .digest(this.toByteArray())
         .fold("") { str, it -> str + "%02x".format(it) }
 }
+
+fun String?.toInteger(): Int? {
```
Arkariang, line 184: 👍

app/src/main/java/com/kickstarter/services/DiscoveryParams.kt
Arkariang, line 1: nice one!

app/src/main/java/com/kickstarter/ui/viewholders/RewardsItemViewHolder.kt
```diff
@@ -11,7 +12,7 @@ class RewardsItemViewHolder(private val binding: RewardsItemViewBinding) : KSVie
         val rewardsItem = ObjectUtils.requireNonNull(data as RewardsItem?)
         val title = ksString.format(
             "rewards_info_item_quantity_title", rewardsItem.quantity(),
-            "quantity", ObjectUtils.toString(rewardsItem.quantity()),
+            "quantity", rewardsItem.quantity().numToString(),
```
Arkariang, line 15: for kotlin I don't really think in makes sense, the language already provides `.toString()` in the primitive types

app/src/main/java/com/kickstarter/services/ApiClientV2.java
```diff
@@ -279,8 +279,8 @@ public ApiClientV2(final @NonNull ApiServiceV2 service, final @NonNull Gson gson
 
   @Override
   public @NonNull Observable<Update> fetchUpdate(final @NonNull Update update) {
-    final String projectParam = ObjectUtils.toString(update.projectId());
-    final String updateParam = ObjectUtils.toString(update.id());
+    final String projectParam = AnyExtKt.numToString(update.projectId());
```
Arkariang, line 282: For java I see it somewhat useful as `Long.toString(update.projectId())` might throw exceptions and the extensions validates the types, but we need to add some tests for the method

app/src/main/java/com/kickstarter/libs/utils/extensions/AnyExt.kt
leighdouglas: Woops, didn't realize i left that there 😅😅

app/src/main/java/com/kickstarter/libs/utils/extensions/AnyExt.kt
leighdouglas: I kept it as there are 2 java classes that are currently using it (ApiClient and ApiClientV2)

app/src/main/java/com/kickstarter/libs/utils/extensions/AnyExt.kt
leighdouglas: I ended up keeping the AnyExt require non-null since there is a java class currently using it, ActivityListViewHolder. Once we migrate that to java or deprecate it, we can remove these extension methods in favor of the kotlin `Preconditions.requireNotNull()`

mtgriego (approved):
Good work!

PR #1847, opened 2023-08-29
[no-jira] Download graphql schema

leighdouglas:
# 📲 What

Download graphql schema changes, updated android app

# 👀 See

no user facing changes

# 📋 QA

Everything should work as normal

### DIFF
--- a/app/src/main/graphql/schema.json
+++ b/app/src/main/graphql/schema.json
@@ -606,6 +606,33 @@
             "deprecationReason": null
           },
           {
+            "name": "order",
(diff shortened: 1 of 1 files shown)


Arkariang:
@leighdouglas for updating the schema.json, instead of upload the file obtained, just modify the currently existing one https://github.com/kickstarter/android-oss/blob/master/app/src/main/graphql/schema.json , modifications to that file are the only change that shoule be added when updating the schema.

Arkariang (approved):

PR #1849, opened 2023-08-29
[MBL-790] Migrate project search result viewhold view model to rxjava2.2

leighdouglas:
…e viewmodel inheritance from envrionmental commitments view holder view model, tests, linter

# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/EnvironmentalCommitmentsViewHolderViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/EnvironmentalCommitmentsViewHolderViewModel.kt
@@ -1,6 +1,5 @@
 package com.kickstarter.viewmodels
 
-import androidx.lifecycle.ViewModel
 import com.kickstarter.libs.EnvironmentalCommitmentCategories
 import com.kickstarter.libs.utils.ObjectUtils
 import com.kickstarter.libs.utils.extensions.addToDisposable
@@ -14,6 +13,9 @@ interface EnvironmentalCommitmentsViewHolderViewModel {
     interface Inputs {
         /** Configure the view model with the [EnvironmentalCommitment]. */
         fun configureWith(environmentalCommitmentInput: EnvironmentalCommitment)
+
+        /** Clear subscriptions, called from ViewHolder when view is destroyed. */
+        fun onCleared()
     }
 
     interface Outputs {
@@ -23,8 +25,7 @@ interface EnvironmentalCommitmentsViewHolderViewModel {
         fun category(): Observable<Int>
     }
 
-    class EnvironmentalCommitmentsViewHolderViewModel() :
-        ViewModel(), Inputs, Outputs {
+    class EnvironmentalCommitmentsViewHolderViewModel() : Inputs, Outputs {
         val inputs: Inputs = this
         val outputs: Outputs = this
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/viewmodels/EnvironmentalCommitmentsViewHolderViewModel.kt
```diff
@@ -23,8 +25,7 @@ interface EnvironmentalCommitmentsViewHolderViewModel {
         fun category(): Observable<Int>
     }
 
-    class EnvironmentalCommitmentsViewHolderViewModel() :
-        ViewModel(), Inputs, Outputs {
```
mtgriego, line 27: Just curious why the viewmodel extension was removed here, if its no longer a viewmodel should this just be a util class?

app/src/main/java/com/kickstarter/viewmodels/EnvironmentalCommitmentsViewHolderViewModel.kt
```diff
@@ -55,7 +56,6 @@ interface EnvironmentalCommitmentsViewHolderViewModel {
 
         override fun onCleared() {
             disposables.clear()
-            super.onCleared()
         }
```
mtgriego, line 59: onCleared came from extending viemodel, removing that extension also means you needed to add this to the inputs

app/src/main/java/com/kickstarter/viewmodels/ProjectSearchResultHolderViewModel.kt
```diff
@@ -50,18 +52,18 @@ interface ProjectSearchResultHolderViewModel {
         fun displayPrelaunchProjectBadge(): Observable<Boolean>
     }
 
-    class ViewModel(@NonNull environment: Environment) :
-        ActivityViewModel<ProjectSearchResultViewHolder>(environment), Inputs, Outputs {
```
mtgriego, line 54: Same question here, instead of activityviewmodel which was our own design, why not just use a viewmodel extension?

mtgriego (changes requested):
I am not sure about removing the viewmodel extensions for these classes, doing say removes view lifecycle safety and the onCleared calls we were making.  If there is a reason thats fine but we should probably rename these from viewmodels if that is the case

app/src/main/java/com/kickstarter/viewmodels/EnvironmentalCommitmentsViewHolderViewModel.kt
leighdouglas: This is the input function that is being called by the viewholder. Take a look at line 34 in the EnvironmentalCommitmentsViewHolder.kt in this pull request

leighdouglas:
> I am not sure about removing the viewmodel extensions for these classes, doing say removes view lifecycle safety and the onCleared calls we were making. If there is a reason thats fine but we should probably rename these from viewmodels if that is the case

Addressed this in a slack message but wanted to address here as well - inheritance from ViewModel only works with viewmodels on the fragment or activity lifecycle, and should never be made outside of that. This viewmodel is not created with a factory (documentation below)
image

mtgriego (approved):

Arkariang (approved):

PR #1853, opened 2023-09-12
MBL-908: Convert SetPasswordActivity visuals to Compose

leighdouglas:
# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/extensions/ActivityExt.kt
+++ b/app/src/main/java/com/kickstarter/ui/extensions/ActivityExt.kt
@@ -197,7 +197,7 @@ fun Activity.startPreLaunchProjectActivity(project: Project, previousScreen: Str
     TransitionUtils.transition(this, TransitionUtils.slideInFromRight())
 }
 
-fun Activity.startDisclaimerActivity(disclaimerItem : DisclaimerItems) {
+fun Activity.startDisclaimerActivity(disclaimerItem: DisclaimerItems) {
     val intent = when (disclaimerItem) {
         DisclaimerItems.TERMS -> Intent(this, HelpActivity.Terms::class.java)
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/kickstarter/ui/extensions/ActivityExt.kt
```diff
@@ -192,3 +196,13 @@ fun Activity.startPreLaunchProjectActivity(project: Project, previousScreen: Str
...
+
+fun Activity.startDisclaimerActivity(disclaimerItem: DisclaimerItems) {
+    val intent = when (disclaimerItem) {
+        DisclaimerItems.TERMS -> Intent(this, HelpActivity.Terms::class.java)
+        DisclaimerItems.PRIVACY -> Intent(this, HelpActivity.Privacy::class.java)
+        DisclaimerItems.COOKIES -> Intent(this, HelpActivity.CookiePolicy::class.java)
+        DisclaimerItems.HELP -> Intent(Intent.ACTION_VIEW, Uri.parse(Secrets.HelpCenter.ENDPOINT))
+    }
+    startActivity(intent)
+}
```
mtgriego, line 208: Love it, ill use this is my stuff too!

mtgriego (approved):
Nice first compose PR!

app/src/main/java/com/kickstarter/ui/extensions/ActivityExt.kt
leighdouglas: We might actually be able to use this as default values in the drop-down menu composable

PR #1854, opened 2023-09-13
Convert Login to Compose

mtgriego:
# 📲 What

Convert the login screen to compose

# 🤔 Why

Dark mode eventually and design system

# 🛠 How

Brainpower!

# 👀 See
Screenshot_20230913-124440
Screenshot_20230913-135316

# 📋 QA

make sure login still works as normal, forgot password dialog still works, etc

# Story 📖

[MBL-961](https://kickstarter.atlassian.net/browse/MBL-961)

[MBL-961]: https://kickstarter.atlassian.net/browse/MBL-961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/LoginActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/LoginActivity.kt
@@ -101,7 +101,6 @@ class LoginActivity : ComponentActivity() {
                 }
             }
 
-
             var changedPasswordMessage = ""
 
             this.viewModel.outputs.showChangedPasswordSnackbar()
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/kickstarter/ui/activities/LoginActivity.kt
```diff
@@ -3,129 +3,200 @@ package com.kickstarter.ui.activities
...
+        setContent {
+            var prefillEmail = viewModel.outputs.prefillEmail().subscribeAsState(initial = "").value
+
+            var isLoading = viewModel.isLoading().subscribeAsState(initial = false).value
+
+            var error by remember { mutableStateOf("") }
+
+            errorMessages().subscribe {
+                error = it
+            }.addToDisposable(disposables)
```
leighdouglas, line 77: `            var error by remember { mutableStateOf("") }

            errorMessages().subscribe {
                error = it
            }.addToDisposable(disposables)`

Curious about what this block of code is doing?

mtgriego: So this class was doing error messages in a different way, the error was not being put into the observable itself, but it was being mapped from a Unit observable, so I made a separate variable to hold the error so I can reset it, since I could not do so with the observable itself.

Let me know if that makes sense!

PR #1858, opened 2023-09-20
[MBL-955] POC for Couroutines: ChangePasswordViewmodel

mtgriego:
# 📲 What

WIP, Do not review yet, feel free to look at the code and comment what you like/done like for now.  Once tests are done we can decide if we want to merge this in or not

# 🤔 Why

Coroutines are google supported and much more human readable ways to interact with APIs/Apollo

# 🛠 How

see https://developer.android.com/kotlin/coroutines
and https://developer.android.com/kotlin/flow

# 👀 See

No visual changes, only code

# 📋 QA

Verify all functionality of change password still works (happy path and errors)

# Story 📖

[MBL-955](https://kickstarter.atlassian.net/browse/MBL-955)

[MBL-955]: https://kickstarter.atlassian.net/browse/MBL-955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -319,6 +319,10 @@ dependencies {
 
     // Desugaring Java 8 to < API 26
     coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.3'
+
+    // Coroutines
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.4'
+    implementation 'com.apollographql.apollo:apollo-coroutines-support:2.5.14'
 }
 
 // SHA and timestamp caching courtesy of https://github.com/gdg-x/frisbee/blob/develop/app/build.gradle#L193-L218
--- a/app/src/main/java/com/kickstarter/viewmodels/ChangePasswordViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/ChangePasswordViewModel.kt
@@ -1,165 +1,58 @@
 package com.kickstarter.viewmodels
 
-import UpdateUserPasswordMutation
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.viewModelScope
 import com.kickstarter.libs.Environment
-import com.kickstarter.libs.rx.transformers.Transformers.errorsV2
-import com.kickstarter.libs.rx.transformers.Transformers.takeWhenV2
-import com.kickstarter.libs.rx.transformers.Transformers.valuesV2
-import com.kickstarter.libs.utils.extensions.addToDisposable
-import com.kickstarter.libs.utils.extensions.isNotEmptyAndAtLeast6Chars
-import com.kickstarter.libs.utils.extensions.newPasswordValidationWarnings
-import io.reactivex.Observable
-import io.reactivex.disposables.CompositeDisposable
-import io.reactivex.subjects.BehaviorSubject
-import io.reactivex.subjects.PublishSubject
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/kickstarter/viewmodels/ChangePasswordViewModel.kt
```diff
@@ -1,165 +1,58 @@
...
+    val isLoading: Flow<Boolean> get() = mutableIsLoading
+
+    private val apolloClient = requireNotNull(this.environment.apolloClientV2())
+    private val analytics = this.environment.analytics()
+
+    fun updatePassword(oldPassword: String, newPassword: String) {
+        viewModelScope.launch {
+            mutableIsLoading.emit(true)
+            val response =
+                apolloClient.updateUserPasswordNew(oldPassword, newPassword, newPassword)
```
Arkariang, line 31: Could we explore something? We are re-writting with coroutines the Networking layer as well (KSApolloClientV2), which is perfectly correct, but in order to reduce the amount of layers and code we are changing could we explore as well the usage of [as-flow](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-rx2/kotlinx.coroutines.rx2/as-flow.html) method ? Or any of the other utilities present on https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-rx2/kotlinx.coroutines.rx2/as-flow.html 

I would like to figure out if to adopt coroutines is mandatory to re-write the network layer, or can it can be done in a more reduced scope, just on the UI layer with (Actvity/Frgments/VM's).

If the conclusion is that we need to rewrite the network layer I'm perfectly ok with it (long term it's what will happen anyway), but want to know all the available options here.

mtgriego: I can play around with that for sure.  I think eventually we would re-write it so this is my all in approach while still trying tnot to change very much when it comes to the network layer (you can see the call to updateUserPasswordNew doesnt change very much from the non new version other than the await() call which does the response vs error handling for us).  My guess is that while we can use asFlow the amount of effort to update it to use await (or newer calls in the newer library versions) its about the same.

Arkariang: want to make sure we do not block any thread by using `await` (it should not if we use it as a promise) but wanna explore all the options, if there is a bridge RX - coroutines wanna see how it behaves before doing a full rewrite. The main reason at the present moment to explore the bridge is to not juggle with so many migrations/adoptions as we currently are. But completely agree on the point that eventually there will be a full re-writte for the networking layer, but was hopping to delay that re-writting to the moment were we can adopt `repository` pattern and possibly get rid of the V1 API, but let's discuss further on the Android sync, it's quite a fascinating topic ❤️

app/build.gradle
```diff
@@ -319,6 +319,10 @@ dependencies {
 
     // Desugaring Java 8 to < API 26
     coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.3'
+
+    // Coroutines
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.4'
```
CoreFloDev, line 324: Hey that's not only what to do to add the coroutine on a project please have a look at the documentation https://github.com/Kotlin/kotlinx.coroutines/blob/master/README.md#android

PR #1859, opened 2023-09-21
MBL-962: Convert SignupActivity Visuals to Compose

leighdouglas:
# 📲 What

Compose migration! 

# 🛠 How

[Compose Technical Plan](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/2452357206/Compose+KS+Best+practices)

# 👀 See

Screenshot 2023-09-21 at 11 49 34 AM
Screenshot 2023-09-21 at 11 50 10 AM

# 📋 QA

A few things to note:

[Previously we had a bug ](https://kickstarter.atlassian.net/browse/MBL-847) that prevented users from opting in to newsletters on first try. This PR should fix this bug. To test this:

When signing up, toggle should initially be in the off position. tap the toggle **ONCE** to opt-in
Monitor network response, newsletter opt-in fields should be set to true
Screenshot 2023-09-21 at 11 34 24 AM
Visit settings > newsletters > see that user is opted in to receive newsletters
Sign up with an account, tap the toggle on or off a few times, then finally to off, tap signup and monitor the response. It should be set to false

Signup button should be disabled if:
- name field is empty
- email field is not an email
- password is < 6 characters and does not contain only whitespaces (see comment about containing spaces)

Test errors:
- Signup using a defunct email that ends with @kickstarter.com (try thisisanemailaddress@kickstarter.com), you should get an error snackbar
- Turn off wifi or cellular data, signup and see the timeout snackbar

# Story 📖

[MBL-962: Convert SignupActivity Visuals to Compose](https://kickstarter.atlassian.net/browse/MBL-962)


app/src/main/java/com/kickstarter/libs/utils/extensions/StringExt.kt
```diff
@@ -114,9 +114,9 @@ fun String.maskEmail(): String {
 }
 
 /**
- * validate password isNotEmptyAndAtLeast6Chars
+ * validate password is not empty, contains at least 6 characters, and does not contain spaces
  */
-fun String.isNotEmptyAndAtLeast6Chars() = this.isNotEmpty() && this.length >= MINIMUM_PASSWORD_LENGTH
+fun String.validPassword() = this.isNotEmpty() && this.length >= MINIMUM_PASSWORD_LENGTH && !this.contains(" ")
```
leighdouglas, line 119: I realized that users could create passwords using spaces, so i updated to remove that functionality

app/src/main/java/com/kickstarter/ui/activities/SignupActivity.kt
```diff
@@ -1,130 +1,87 @@
...
     private val viewModel: SignupViewModel.SignupViewModel by viewModels { viewModelFactory }
     private val disposables = CompositeDisposable()
+    var darkModeEnabled = false
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
         this.getEnvironment()?.let { env ->
             viewModelFactory = SignupViewModel.Factory(env)
+            darkModeEnabled = env.featureFlagClient()?.getBoolean(FlagKey.ANDROID_DARK_MODE_ENABLED) ?: false
```
mtgriego, line 33: Small change needed here:
If this is true it will always show dark mode rather than using the system setting, do something like this:
`useDarkTheme = if (darkModeEnabled) isSystemInDarkTheme() else false`

mtgriego (changes requested):
Nicely done, just one small change needed, but otherwise solid work

app/src/main/java/com/kickstarter/ui/activities/SignupActivity.kt
leighdouglas: If you take a look at line 51, we are checking for both the dark mode feature flag as well as if the system is in dark mode

mtgriego: oh woops

mtgriego (approved):
Woopsies!

PR #1897, opened 2023-11-08
[MBL-786] Adjust project card to use rx2

mtgriego:
# 📲 What

Rx 2 adjustments for the project card viewholder

# 🤔 Why

Refactoring is good

# 🛠 How

🧠 s

# 📋 QA

Make sure the project cards act as expected:
- projects we love are labeled
- they are clickable and lead to the projects
- the category is correct
- etc

# Story 📖

[MBL-786](https://kickstarter.atlassian.net/browse/MBL-786)

[MBL-786]: https://kickstarter.atlassian.net/browse/MBL-786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/viewmodels/ProjectCardHolderViewModel.kt
```diff
@@ -307,27 +326,25 @@ interface ProjectCardHolderViewModel {
                 .subscribe(this.heartDrawableId)
```
leighdouglas, line 326: I believe we have to use a lambda expression here: 
`.subscribe { this.heartDrawableId.onNext(it)`

mtgriego: good catch!

leighdouglas (approved):
lgtm! left an additional comment about some null checks happening in the viewmodel, but they arent a blocker for merging!

PR #1900, opened 2023-11-13
MBL-813:  EditProfileViewModel.kt to Jetpack ViewModel and RxJava2.2

leighdouglas:
# 📲 What

# 🤔 Why

# 👀 See

No user facing changes

# 📋 QA

Discovered a bug with the edit profile page in that location and bio do not save. However, test that checking the private profile switch works properly - toggle the switch on, leave the screen and return, should still be checked. 

# Story 📖

[MBL-813: EditProfileViewModel.kt to Jetpack ViewModel and RxJava2.2](https://kickstarter.atlassian.net/browse/MBL-813)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/utils/extensions/ProjectExt.kt
+++ b/app/src/main/java/com/kickstarter/libs/utils/extensions/ProjectExt.kt
@@ -160,7 +160,7 @@ fun isProjectNamePunctuated(name: String): Boolean = name.substring(name.length 
 /**
  * Returns the metadata for the project based on some backing, starred, or featured, otherwise returns null
  */
-fun Project.metadataForProject(): ProjectMetadata? =
+fun Project.metadataForProject(): ProjectMetadata =
     when {
         this.isBacking() -> ProjectMetadata.BACKING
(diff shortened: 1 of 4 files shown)


app/src/test/java/com/kickstarter/viewmodels/EditProfileViewModelTest.kt
```diff
@@ -0,0 +1,195 @@
...
+    private val hidePrivateProfileRow = TestSubscriber<Boolean>()
+    private val user = TestSubscriber<User>()
+    private val userAvatarUrl = TestSubscriber<String>()
+    private val userName = TestSubscriber<String>()
+    private val unableToSavePreferenceError = TestSubscriber<String>()
+
+    private val disposables = CompositeDisposable()
+
+    private fun setUpEnvironment(environment: Environment) {
+        this.vm = EditProfileViewModel.EditProfileViewModel(environment)
```
Arkariang, line 29: Fyi, non mandatory, you can use the factory to create the VM, so it is also covered by the tests: 
`Factory(env).create(EditProfileViewModel::class.java)` you can take a look for example at `ProjectSocialViewModelTest:28`

PR #1913, opened 2023-12-06
[MBL-1087] update viewmodel with more null filtering and adjusting subscribing

mtgriego:
# 📲 What

A crash came up with this viewmodel:
https://console.firebase.google.com/u/0/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/efe8a43e0b95638aae8bfefd3099ce72?time=last-seven-days&versions=3.17.0%20(2014150890)&types=crash&sessionEventKey=6570B56D01E3000150EE85831ED32812_1887987293388166730

So this is to make more null safetey and fix the refactoring of this class

# 🤔 Why

Crashes are bad

# 🛠 How

Add more null filtering before mapping in some places and add better subscribe handling

# 📋 QA

Go to a project and confirm all rewards are being displayed

# Story 📖

[MBL-1087](https://kickstarter.atlassian.net/browse/MBL-1087)

[MBL-1087]: https://kickstarter.atlassian.net/browse/MBL-1087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/RewardViewHolderViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/RewardViewHolderViewModel.kt
@@ -203,7 +203,8 @@ interface RewardViewHolderViewModel {
                         this.ksCurrency
                     )
                 }
-                .subscribe(this.minimumAmountTitle)
+                .subscribe { this.minimumAmountTitle.onNext(it) }
+                .addToDisposable(disposables)
 
             val userCreatedProject = this.currentUser.observable()
                 .compose<Pair<KsOptional<User>?, Project>>(combineLatestPair(project))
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/kickstarter/viewmodels/RewardViewHolderViewModel.kt
```diff
@@ -226,7 +228,8 @@ interface RewardViewHolderViewModel {
             projectAndReward
                 .map { it.first }
```
Arkariang, line 229: Filter before these maps, if I recall correctly the currency field might be null

mtgriego: in this case its an == check so it will always return a boolean, so even if one is null (currentCurrency in this case) it will still return a boolean.  it.currency is always a string, the currentCurrency is nullable, but thats okay

Arkariang (approved):

PR #1917, opened 2023-12-11
MBL-1053: Identify if https links in messages are KSR links

leighdouglas:
# 📲 What

Goal is to validate links sent in messages by creators to open app links in-app where applicable.

# 🤔 Why

Previously, urls sent by creators would always open an external browser. Now, when it is a kickstarter link, it will open in app. All other links will continue to go to an external browser. 

# 🛠 How

- Update messages view holder and view model to rxjava 2
- Add functionality to check if textview contains a url
- Open deeplink activity and supply with url, this will handle the url validation

# 👀 See

https://github.com/kickstarter/android-oss/assets/19390326/e46e5b09-4043-4090-b20d-c642465873c3

# 📋 QA

- Sign-in
- Go to your messages
- Find a kickstarter link, tap, should not redirect out of app
- Find a non-kickstarter link, tap, should redirect to system browser

# Story 📖

[MBL-1053: Identify if https links in messages are KSR links](https://kickstarter.atlassian.net/browse/MBL-1053)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/MessageHolderViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/MessageHolderViewModel.kt
@@ -2,7 +2,6 @@ package com.kickstarter.viewmodels
 
 import android.util.Pair
 import com.kickstarter.libs.ActivityViewModel
-import com.kickstarter.libs.CurrentUserType
 import com.kickstarter.libs.CurrentUserTypeV2
 import com.kickstarter.libs.Environment
 import com.kickstarter.libs.utils.PairUtils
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/viewmodels/MessageHolderViewModel.kt
```diff
@@ -48,7 +48,7 @@ interface MessageHolderViewModel {
         ActivityViewModel<MessageViewHolder?>(environment),
         Inputs,
         Outputs {
-        private val currentUser: CurrentUserType?
+        private val currentUser: CurrentUserTypeV2?
```
mtgriego, line 51: very minor, but currentUser cant be null (due to requireNotNull(*)) so it doesn't need to be nullable here

mtgriego (approved):

Arkariang (approved):

app/src/main/java/com/kickstarter/viewmodels/MessageHolderViewModel.kt
leighdouglas: I'll fix in a subsequent PR! Thanks!

PR #1922, opened 2024-01-05
[MBL-1127] Dark mode

mtgriego:
# 📲 What

Add forced dark mode to the app so the app matches the system theme and shows dark mode content

# 🤔 Why

Dark mode is one of the most requested features for android

# 🛠 How

Added a new theme for v29+ that includes
`<item name="isLightTheme">true</item>`
which forces the app into a dark theme.  This is not a permanent fix, but a quick win

# 👀 See

Screenshot_20240102-124202
Screenshot_20240102-124127
Screenshot_20240102-124038I![Screenshot_20240102-124043]
Screenshot_20240102-124023
Screenshot_20240102-123849
(https://github.com/kickstarter/android-oss/assets/7563288/be3814b0-2d86-4fe0-b4e5-3580d6961574)

# 📋 QA

Go through every screen in the app and confirm it looks good in both light and dark modes
NOTE:  The help page webviews (privacy policy, terms etc) will only appear in light mode

# Story 📖

[MBL-1127](https://kickstarter.atlassian.net/browse/MBL-1127)

[MBL-1127]: https://kickstarter.atlassian.net/browse/MBL-1127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/ui/activities/CreatePasswordActivity.kt
```diff
@@ -62,6 +63,8 @@ class CreatePasswordActivity : AppCompatActivity() {
                         AppThemes.LIGHT.ordinal -> false
                         else -> false
                     }
+                } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
Arkariang, line 66: [not mandatory], but we could try to move all theme related logic to the composable `KickstarterApp`, trying to avoid having to repeat in every single screen the theming setup.

Arkariang (dismissed):

app/src/main/java/com/kickstarter/ui/activities/CreatePasswordActivity.kt
mtgriego: thats a good idea, just not sure how much access the kickstarter app has to builds and stuff, but ill play around with that, would save a lot of time

Arkariang (approved):

PR #1930, opened 2024-01-16
[MBL-1135] Add analytics value for session_force_dark_mode

mtgriego:
# 📲 What

Add an analytic property for the session if the user is in force dark mode

# 🤔 Why

We want to see if there is any sort of impact for users in dark mode vs not

# 🛠 How

Add the session property after checking the system version and dark mode setting

# 📋 QA

Run the app and confirm the `session_force_dark_mode=XXXX` is correct based on system setting and version

# Story 📖

[MBL-1135](https://kickstarter.atlassian.net/browse/MBL-1135)

[MBL-1135]: https://kickstarter.atlassian.net/browse/MBL-1135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/TrackingClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/TrackingClient.kt
@@ -124,4 +124,17 @@ abstract class TrackingClient(
             cm.getNetworkCapabilities(it)?.hasTransport(NetworkCapabilities.TRANSPORT_WIFI)
         } ?: false
     }
+
+    override fun sessionForceDarkMode(): Boolean {
+        return if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.Q) {
+            when (context.resources?.configuration?.uiMode?.and(Configuration.UI_MODE_NIGHT_MASK)) {
+                Configuration.UI_MODE_NIGHT_YES -> {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/libs/TrackingClient.kt
```diff
@@ -124,4 +124,17 @@ abstract class TrackingClient(
             cm.getNetworkCapabilities(it)?.hasTransport(NetworkCapabilities.TRANSPORT_WIFI)
         } ?: false
     }
+
+    override fun sessionForceDarkMode(): Boolean {
```
Arkariang, line 128: just an fyi, no need to act on it. This method could probably just be a private method, not added to the interface. 
The trackingclienttype interface could use some refactoring (@leighdouglas might remember some conversations we had around the topic), the interface should probably contain only generic action methods (tack, identify, reset, isEnabled ... ) that can be applied to any tracking system (Segment, Google Analytics or whatever tracking third party we might want). The rest are properties and not necessarily behavioural

Arkariang (approved):

PR #1940, opened 2024-01-31
[MBL-1174] Add support for the CreateCheckout mutation

mtgriego:
# 📲 What

Adds support for the new create checkout mutation used in the post campaign checkout

Note:  when no rewards is selected we need to send a blank list for the rewardsIds, this will happen in implementation

# 🤔 Why

Implementation is to come later, but this is to add it into the codebase for use when we are moving to implementation

# 🛠 How

Created the mutation and data classes needed to execute the mutation successfully with the backend

# 👀 See

Below is the plaintext and graphql input and output for the call

```
CreateCheckout Input: ProjectId: 769466129, ammount: 248.0, locationId: 23424977, rewardIds: [com.kickstarter.models.Reward@f432eda, com.kickstarter.models.Reward@e7d8f0b], refTag: null

X-APOLLO-OPERATION-NAME: CreateCheckout
{"operationName":"CreateCheckout","variables":{"projectId":"UHJvamVjdC03Njk0NjYxMjk=","amount":"248.0","rewardIds":["UmV3YXJkLTk0NTg0OTE=","UmV3YXJkLTk0NTg1MjM="],"locationId":"23424977","refParam":null},"query":"mutation CreateCheckout($projectId: ID!, $amount: String!, $rewardIds: [ID!], $locationId: String, $refParam: String) { createCheckout(input: {projectId: $projectId, amount: $amount, rewardIds: $rewardIds, locationId: $locationId, refParam: $refParam}) { __typename checkout { __typename id paymentUrl } } }"}

{"data":{"createCheckout":{"__typename":"CreateCheckoutPayload","checkout":{"__typename":"Checkout","id":"Q2hlY2tvdXQtMTkyODQzNzY2","paymentUrl":"/checkouts/192843766/payments/new"}}}}

CreateCheckout Result: id: 192843766, url: /checkouts/192843766/payments/new
```

# 📋 QA

At the moment it is not implemented, so testing involves adding this call to the pledge fragment and calling with the available data to check for correct result.  As long as there is an ID that is all we ned for future calls for now

# Story 📖

[MBL-1174](https://kickstarter.atlassian.net/browse/MBL-1174)

[MBL-1174]: https://kickstarter.atlassian.net/browse/MBL-1174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/models/CheckoutPayment.kt
```diff
@@ -0,0 +1,3 @@
+package com.kickstarter.models
+
+data class CheckoutPayment(val id: Long?, val paymentUrl: String?)
```
Arkariang, line 3: Nitpick here, the ID is mandat
Screenshot 2024-02-01 at 9 06 11 AM
ory, should not be optional

app/src/main/java/com/kickstarter/services/mutations/CreateCheckoutData.kt
```diff
@@ -0,0 +1,7 @@
+package com.kickstarter.services.mutations
+
+import com.kickstarter.libs.RefTag
+import com.kickstarter.models.Project
+import com.kickstarter.models.Reward
+
+data class CreateCheckoutData(val project: Project, val amount: String, val locationId: String?, val rewardsIds: List<Reward>?, val refTag: RefTag?)
```
Arkariang, line 7: Another nitpick here, 
reswarIds list should not be null ever, it generates an internal server error. It could should be an empty list when no values, so you can change it to `rewardsIds: List<Reward> = emptyList()`.

Here server response when you send null on that field:
Screenshot 2024-02-01 at 9 08 21 AM

Here server response when you send empty list on that field:

Screenshot 2024-02-01 at 9 02 55 AM

app/src/main/java/com/kickstarter/models/CheckoutPayment.kt
mtgriego: I was marking it nullable because the decode function is nullable, but ill throw an exception if its null instead.

app/src/main/java/com/kickstarter/services/mutations/CreateCheckoutData.kt
mtgriego: good point, it should also be an empty list if the use picks "no reward" so ill change that

Arkariang (approved):

PR #1949, opened 2024-02-13
POC: Address Collection on Pledge Screen

leighdouglas:
# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖

### DIFF
--- a/app/src/main/java/com/kickstarter/mock/services/MockApolloClient.kt
+++ b/app/src/main/java/com/kickstarter/mock/services/MockApolloClient.kt
@@ -43,12 +43,14 @@ import com.kickstarter.services.apiresponses.DiscoverEnvelope
 import com.kickstarter.services.apiresponses.ShippingRulesEnvelope
 import com.kickstarter.services.apiresponses.commentresponse.CommentEnvelope
 import com.kickstarter.services.apiresponses.updatesresponse.UpdatesGraphQlEnvelope
+import com.kickstarter.services.mutations.CreateAddressData
 import com.kickstarter.services.mutations.CreateBackingData
 import com.kickstarter.services.mutations.CreateCheckoutData
 import com.kickstarter.services.mutations.PostCommentData
 import com.kickstarter.services.mutations.SavePaymentMethodData
 import com.kickstarter.services.mutations.UpdateBackingData
 import com.kickstarter.viewmodels.usecases.TPEventInputData
+import com.stripe.android.paymentsheet.addresselement.AddressDetails
 import rx.Observable
 import type.CurrencyCode
 import java.util.Collections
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/kickstarter/mock/services/MockApolloClient.kt
```diff
@@ -43,12 +43,14 @@ import com.kickstarter.services.apiresponses.DiscoverEnvelope
...
 import com.kickstarter.services.apiresponses.commentresponse.CommentEnvelope
 import com.kickstarter.services.apiresponses.updatesresponse.UpdatesGraphQlEnvelope
+import com.kickstarter.services.mutations.CreateAddressData
 import com.kickstarter.services.mutations.CreateBackingData
 import com.kickstarter.services.mutations.CreateCheckoutData
 import com.kickstarter.services.mutations.PostCommentData
 import com.kickstarter.services.mutations.SavePaymentMethodData
 import com.kickstarter.services.mutations.UpdateBackingData
 import com.kickstarter.viewmodels.usecases.TPEventInputData
+import com.stripe.android.paymentsheet.addresselement.AddressDetails
```
Arkariang, line 53: nitpick for the future implementation, it is perfectly ok for the POC no need to change anythin, but we should try to define our own dataModel on the real implementation, instead using Stripes's defined one. If we ever want to substitute the 3rd party, or adopt for example Google Place and build the address collection piece ourself we should not be tied to an external data model.

app/src/main/java/com/kickstarter/ui/activities/ThanksActivity.kt
```diff
@@ -55,6 +65,16 @@ class ThanksActivity : AppCompatActivity() {
             orientation = RecyclerView.VERTICAL
         }
 
+        val addressConfiguration = AddressLauncher.Configuration(
+                additionalFields = AddressLauncher.AdditionalFieldsConfiguration(
+                phone = AddressLauncher.AdditionalFieldsConfiguration.FieldConfiguration.REQUIRED
+        ),
+                allowedCountries = setOf("US", "CA", "GB"),
```
Arkariang, line 72: For the real implementation, perfectly ok on the POC, `config.launchedCountries` will contain the array of allowed countries if the shipping location is worlwide. If the shipping location is restricted we should send only the allowed countries.

app/src/main/java/com/kickstarter/mock/services/MockApolloClient.kt
leighdouglas: I think this is an unused import that i forgot to delete, as i created a datamodel for this `CreateAddressData`, totally agree that we should use our own over the third party implementation!

PR #1964, opened 2024-02-29
[MBL-1229] Add Compose Views for Add-ons screen

mtgriego:
# 📲 What

Create the composables to be used in the new post campaign add-ons fragment in the flow

# 🤔 Why

All new things should be done in compose and coroutines and this is a new flow

# 🛠 How

Took a lot of reference from the existing views and classes to recreate as best as possible

# 👀 See

Please note colors are not final as I am working with design to review these colors, but they are using as much of the new colors as I could figure out.
Screenshot 2024-02-29 at 10 56 37 AM

Please note that the pricing, shipping, and some other functionality of the page needs an environment to parse our current strings so they wont show up in the preview
Screenshot 2024-02-29 at 10 57 08 AM

# 📋 QA

This will come more strongly during implementation, the container works as expected in the preview, but the acreen needs an environment to fully function so some functionality seems wonky without it.

# Story 📖

[MBL-1229](https://kickstarter.atlassian.net/browse/MBL-1229)

[MBL-1229]: https://kickstarter.atlassian.net/browse/MBL-1229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/AddOnsScreen.kt
```diff
@@ -0,0 +1,320 @@
...
+                    .build(),
+                onItemAddedOrRemoved = {},
+                onContinueClicked = {}
+            )
+        }
+    }
+}
+
+@Composable
+fun AddOnsScreen(
```
Arkariang, line 89: Could we add tests? All parameters seem able to be mocked

mtgriego: i was planning to add tests once we get to implementation because im not 100% sure ive covered all possible cases for these views yet.

app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/AddOnsContainer.kt
```diff
@@ -0,0 +1,211 @@
...
+import com.kickstarter.ui.compose.designsystem.KSTheme.dimensions
+import com.kickstarter.ui.compose.designsystem.KSTheme.typography
+import com.kickstarter.ui.compose.designsystem.shapes
+
+@Composable
+@Preview(name = "Light", uiMode = Configuration.UI_MODE_NIGHT_NO)
+@Preview(name = "Dark", uiMode = Configuration.UI_MODE_NIGHT_YES)
+private fun AddOnsContainerPreview() {
+    KSTheme {
+        AddOnsContainer(
```
Arkariang, line 42: Not for now, if you guys are in a rush to deliver for project late pledges, but ideally we should start creating re-usable components for these type of cards.
The addOns one could potentially be re-used for rewards in both rewards-carousel and manage pledge screen.

Same for project cards in the future! 

cc: @ycheng-kickstarter @leighdouglas

mtgriego: The functionality of the flows are so similar my eventual plan would be to use the existing screens with whatever copy change we need.  iOS is going to be using these in their existing flows.  If we had more time I would just refactor all the flows into these screens, but that will be a down the road initiative for sure.

PR #1971, opened 2024-03-07
MBL-1262: QA party bug fixes

Arkariang:
# 📲 What
3 bug fixes in this PR :) 

# 👀 See

1 - Navigate away to other apps while chrome tab is presented: 
- the chrome tab no longer gets killed when lost foreground.
- the activity task remains the parent, not a new one.

https://github.com/kickstarter/android-oss/assets/4083656/187a88af-64de-4d1a-8b9d-5a5874b51217

2 - Not related with OAuth bot found it by chance :) the issue is with verified deeplinks, the '*' wildcard regex, even though specified in the manifest as auto-verified, it had to manually by added as verified link, exchanged over `www` to not have to manually add it.
| Before 🐛 | 
Screenshot 2024-03-07 at 1 26 55 PM

https://github.com/kickstarter/android-oss/assets/4083656/54e14052-0e26-4020-a1eb-1020de560a41

After 🦋 |
Screenshot 2024-03-07 at 1 28 20 PM

https://github.com/kickstarter/android-oss/assets/4083656/05a7a959-a41a-4532-8a56-6b74414766a9

3 - During Change Password flow, the user gets logged out after the password was changed successfully, and presented the login flow to access again. The presented flow was the non-OAuth version. 

| --- | --- |
|  |  |

# 📋 QA
1 - For number 1 just navigate away from the app to retrieve credentials, (any password manager etc), come back to the app. There should be only 1 instance of the app, the Chrome tab parent task is now the app and not a new one, Chrome tab no longer killed after losing foreground.

2 - For number two, fresh install a version of the app with this branch, and take a look App info -> open by default -> 1 verified links. Previous versions to this branch will show 0 verified links, even though the documentation says the wildcard regex is valid for hosts, it seems not be able to autoverify automatically when used.
Screenshot 2024-03-07 at 1 38 23 PM

3 - Follow the change password flow -> change your password -> when feature flag on now you should be presented with `LoginToutActivity`, when feature flag off you should be presented with `LoginActivity`

# Story 📖

[MBL-1262](https://kickstarter.atlassian.net/browse/MBL-1262)

[MBL-1262]: https://kickstarter.atlassian.net/browse/MBL-1262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -70,7 +70,6 @@
             android:label="@string/app_name"
             android:exported="true"
             android:launchMode="singleTop"
-            android:taskAffinity=""
             android:theme="@style/SplashTheme">
             <intent-filter android:autoVerify="true">
                 <action android:name="android.intent.action.VIEW" />
@@ -111,7 +110,6 @@
             android:name=".ui.activities.ResetPasswordActivity"
             android:parentActivityName=".ui.activities.LoginActivity"
             android:windowSoftInputMode="adjustResize"
-            android:taskAffinity=""
             android:theme="@style/Login" />
         <activity
             android:name=".ui.activities.SetPasswordActivity"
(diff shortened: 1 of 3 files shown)


mtgriego (approved):
Tested locally and working as expected!

PR #1972, opened 2024-03-07
(WIP) MBL-1236 Call CreateAttributionEvent on ProjectPage for project page …

ycheng-kickstarter:
…viewed event

# 📲 What

This PR is for Step 6 of the event attribution eng plan https://kickstarter.atlassian.net/wiki/spaces/NT/pages/2699100255/Event+Attribution+Engineering+Plan+AKA+How+To+Wire+Up+a+New+GraphQL+Endpoint+on+Android#6.-Add-call-site-in-ProjectPageViewModel

# 🤔 Why

This is where we wire up the call site on the Project page for the endpoint we set up in steps 1-5.

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/AttributionEvents.kt
+++ b/app/src/main/java/com/kickstarter/libs/AttributionEvents.kt
@@ -1,0 +1,60 @@
+package com.kickstarter.libs
+
+import android.annotation.SuppressLint
+import com.kickstarter.libs.utils.AnalyticEventsUtils
+import com.kickstarter.libs.utils.ContextPropertyKeyName
+import com.kickstarter.libs.utils.EventContextValues
+import com.kickstarter.libs.utils.EventName
+import com.kickstarter.services.ApolloClientTypeV2
+import com.kickstarter.services.mutations.CreateAttributionEventData
+import com.kickstarter.ui.data.ProjectData
+import timber.log.Timber
+
+/**
+ * Similar to the [AnalyticEvents] class but specifically for sending attribution events to KSR
+ * event attribution backend. The trackingClient is [KsEventAttributionClient].
+ *
+ */
+class AttributionEvents(
+    val apolloClient: ApolloClientTypeV2
+) {
+    /**
+     * Sends data to the backend for event attribution when the projects screen is loaded.
+     *
+     * @param projectData: The selected project data.
+     * @param pageSectionContext: The section of the project page being viewed.
+     */
+    fun trackProjectPageViewed(projectData: ProjectData, pageSectionContext: String) {
+        val props: HashMap<String, Any> = hashMapOf(ContextPropertyKeyName.CONTEXT_PAGE.contextName to EventContextValues.ContextPageName.PROJECT.contextName)
+        props[ContextPropertyKeyName.CONTEXT_SECTION.contextName] = pageSectionContext
+        props["context_page_url"] = projectData.fullDeeplink().toString()
+
+        //props.putAll(AnalyticEventsUtils.projectProperties(projectData.project(), trackingClient?.loggedInUser()))
+        props.putAll(AnalyticEventsUtils.refTagProperties(projectData.refTagFromIntent(), projectData.refTagFromCookie()))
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/kickstarter/libs/AttributionEvents.kt
```diff
@@ -0,0 +1,52 @@
...
+     * @param projectData: The selected project data.
+     */
+    fun trackProjectPageViewed(projectData: ProjectData) {
+        val props: HashMap<String, Any> = hashMapOf(ContextPropertyKeyName.CONTEXT_PAGE.contextName to EventContextValues.ContextPageName.PROJECT.contextName)
+        props["context_page_url"] = projectData.fullDeeplink().toString()
+        props.putAll(AnalyticEventsUtils.refTagProperties(projectData.refTagFromIntent(), projectData.refTagFromCookie()))
+        track(EventName.PROJECT_PAGE_VIEWED.eventName, props, projectData.project())
+    }
+
+    @SuppressLint("CheckResult")
```
ycheng-kickstarter, line 35: Lint is complaining that we're not using the result of the response. It will be a best effort to track this event so we don't need to do anything with the result.

app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
```diff
@@ -874,7 +874,6 @@ fun getCreateAttributionEventMutation(eventInput: CreateAttributionEventData, gs
             .eventName(eventInput.eventName)
             .eventProperties(eventPropertiesJson)
             .projectId(eventInput.projectId)
-            .clientMutationId(eventInput.clientMutationId)
```
ycheng-kickstarter, line 877: Pretty sure the graphql clients generate these on request, rather than needing to be specified. Regardless, the mutation works fine without me specifying it.

app/src/main/java/com/kickstarter/viewmodels/projectpage/ProjectPageViewModel.kt
```diff
@@ -969,7 +975,10 @@ interface ProjectPageViewModel {
                     }
 
                     val dataWithStoredCookieRefTag = storeCurrentCookieRefTag(data)
+                    // Send event to segment
                     this.analyticEvents.trackProjectScreenViewed(dataWithStoredCookieRefTag, OVERVIEW.contextName)
+                    // Send event to backend event attribution
+                    this.attributionEvents.trackProjectPageViewed(dataWithStoredCookieRefTag)
```
ycheng-kickstarter, line 981: Here we send the event to the backend for event attribution

app/src/main/java/com/kickstarter/libs/AttributionEvents.kt
```diff
@@ -0,0 +1,52 @@
...
+    val apolloClient: ApolloClientTypeV2
+) {
+    /**
+     * Sends data to the backend for event attribution when the projects screen is loaded.
+     *
+     * @param projectData: The selected project data.
+     */
+    fun trackProjectPageViewed(projectData: ProjectData) {
+        val props: HashMap<String, Any> = hashMapOf(ContextPropertyKeyName.CONTEXT_PAGE.contextName to EventContextValues.ContextPageName.PROJECT.contextName)
+        props["context_page_url"] = projectData.fullDeeplink().toString()
```
ycheng-kickstarter, line 30: Per product requirements, we send the full deeplink, not just the ref tag https://kickstarter.atlassian.net/wiki/spaces/NT/pages/2605875227/Event+Attribution+Spike#Adding-support-for-tracking-the-entire-deep-link-url

app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
mtgriego: yeah, we never really use it so should be fine to remove

mtgriego (approved):

PR #1980, opened 2024-03-19
MBL-1258 Add logic for add ons screen

ycheng-kickstarter:
# 📲 What

Update the new add ons screen with business logic for a new pledge flow (deal with existing pledge flow at later time).

# 🤔 Why

The new add ons screen needs actual logic to make the compose components function properly.

# 🛠 How

- Created a new AddOnsViewModel which will emit the following UI states:
   currentShippingRule 
   shippingSelectorIsGone 
   currentAddOnsSelection 
- Moved ShippingRules logic out of CheckoutFlowViewModel and into AddOnsViewModel
- Added logic for the location selector - hide selector when it's a digital reward, update shipping costs when location changes

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

Tap on a project that has addons and and play around with adding addons to pledge. Try to break it!
Some things I checked:
- If the reward is digital, only show digital add ons and hide location selector.
- If the reward is shippable, show all possible add ons and show location selector.
- If you change the shipping location from the selector, the shipping amounts change by location.
- Add different counts of different add-ons. The continue button should update with the correct count. Continuing to confirm pledge screen should display correct total.
- If you continue to confirm pledge screen, navigating back to the addons screen will display same state.
- If you navigate back and select a different reward, your previous addons selection is cleared. 

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-1258


app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/AddOnsContainer.kt
```diff
@@ -74,10 +70,9 @@ fun AddOnsContainer(
     buttonEnabled: Boolean,
     buttonText: String,
     environment: Environment,
-    onItemAddedOrRemoved: (count: Int) -> Unit
+    onItemAddedOrRemoved: (count: Int) -> Unit,
+    itemAddOnCount: Int
 ) {
-    var addOnCount by rememberSaveable { mutableStateOf(0) }
```
ycheng-kickstarter, line 79: We don't want rememberSaveable, because if we navigate back and select a different reward, the addOnCount will display the same as before when it actually should be cleared.

app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
```diff
@@ -0,0 +1,207 @@
...
+            ) -> ShippingRuleFactory.emptyShippingRule()
+            // sameReward -> backingShippingRule // TODO: When changing reward for manage pledge flow
+            else -> defaultShipping
+        }
+
+    fun provideProjectData(projectData: ProjectData) {
+        // no-op
+    }
+
+    // UI events
```
ycheng-kickstarter, line 150: Below are the UI events this VM will handle:
- user selects a reward
- user selects a shipping location
- user adds/removes add-on
- user clicks continue button

app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
```diff
@@ -0,0 +1,207 @@
...
+            .asSharedFlow()
+
+    init {
+        currentUserReward
+            .filter {
+                RewardUtils.isDigital(it) || !RewardUtils.isShippable(it) || RewardUtils.isLocalPickup(it)
+            }
+            .distinctUntilChanged()
+            .subscribe {
+                shippingSelectorIsGone = true
```
ycheng-kickstarter, line 68: If the currentUserReward is not shippable, then we hide the shipping selector UI

app/src/main/java/com/kickstarter/viewmodels/projectpage/CheckoutFlowViewModel.kt
```diff
@@ -31,18 +29,16 @@ data class FlowUIState(
 class CheckoutFlowViewModel(val environment: Environment) : ViewModel() {
```
Arkariang, line 29: Do we have tests associated with this ViewModel?

mtgriego: not yet

ycheng-kickstarter: Not for AddOnsVM either. Will add them as a followup ticket not in this PR

app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
```diff
@@ -0,0 +1,207 @@
...
+    var currentAddOnsSelection: MutableMap<Reward, Int> = mutableMapOf()
+)
+class AddOnsViewModel(val environment: Environment) : ViewModel() {
+    private val disposables = CompositeDisposable()
+    private val currentConfig = requireNotNull(environment.currentConfigV2())
+
+    private val currentUserReward = PublishSubject.create<Reward>()
+    private val shippingRules = BehaviorSubject.create<List<ShippingRule>>()
+    private val defaultShippingRule = PublishSubject.create<ShippingRule>()
+    private val shippingRuleSelected = PublishSubject.create<ShippingRule>()
```
mtgriego, line 41: If possible we would want to move away from RX as much as possible (some of our more established classes that use observables i understand the need to use).  In most cases just having a variable we can change and then emitting changes based on those variables can work.  

When a user selects a new reward (current user reward)instead of emitting an observable we can just make the calls as needed int he viewmodel and emit the UI state, etc.

For the sake of time we can keep this for now, but for future projects we want as little RX as possible added in.

app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
mtgriego, line 68: You set this to true here, but if the user goes back and selects a shippable reward this value doesn't get changed to false anywhere

app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
```diff
@@ -0,0 +1,207 @@
...
+            RewardUtils.isDigital(rw) || !RewardUtils.isShippable(rw) || RewardUtils.isLocalPickup(
+                rw
+            ) -> ShippingRuleFactory.emptyShippingRule()
+            // sameReward -> backingShippingRule // TODO: When changing reward for manage pledge flow
+            else -> defaultShipping
+        }
+
+    fun provideProjectData(projectData: ProjectData) {
+        // no-op
+    }
```
mtgriego, line 148: this is not necessary, please remove

mtgriego (changes requested):
The only strictly needed change is the shippingSelectorIsGone needs to change when the reward is shippable, in testing if i select a non-shippable and then go back and select a shippable it still doesnt show the country selector

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ base_job: &base_job
   executor:
     name: android/android-machine
     resource-class: xlarge
-    tag: 2023.11.1 #https://circleci.com/developer/images/image/cimg/android
```
ycheng-kickstarter, line 8: Copying over Alex's change in master to fix CI: https://github.com/kickstarter/android-oss/pull/1985

app/src/main/java/com/kickstarter/viewmodels/projectpage/CheckoutFlowViewModel.kt
Arkariang: Sorry folks, but I do not feel comfortable not having VM's tested, gonna have to do a request changes here

.circleci/config.yml
Arkariang: super nitpick comment here, could we bring back the comment or add some kind of documentation in guru ?
There will be a moment when testing for new operative system, (Android 15 in this year) when we will need to do spikes, and likely `default` won't contain Android 15 SDK nor all the software releated and we will have to use a tag

mtgriego (approved):

app/src/main/java/com/kickstarter/viewmodels/projectpage/CheckoutFlowViewModel.kt
mtgriego: We will have tickets for fast follow to add tests to the VM's

PR #1991, opened 2024-03-27
[MBL1260] Add Checkout Screen to Flow

mtgriego:
# 📲 What

Adding the checkout screen to the late pledge flow.  Currently will run through the calls needed to make a purchase, but does not proceed to thanks page afterwards

# 🤔 Why

Its the last step and screen in the flow

# 🛠 How

See code

# 👀 See

Screenshot_20240327-101728

# 📋 QA

Go through a late pledge flow and reach this screen.  To monitor calls use logcat

# Story 📖

[MBL-1260](https://kickstarter.atlassian.net/browse/MBL-1260)

[MBL-1260]: https://kickstarter.atlassian.net/browse/MBL-1260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/viewmodels/projectpage/LatePledgeCheckoutViewModel.kt
```diff
@@ -0,0 +1,162 @@
...
+                    apolloClient.getStoredCards().asFlow().map { cards ->
+                        storedCards = cards
+                        mutableLatePledgeCheckoutUIState.emit(
+                            LatePledgeCheckoutUIState(
+                                storeCards = storedCards,
+                                userEmail = userEmail,
+                            )
+                        )
+                    }.catch {
+                        // Some error
```
leighdouglas, line 75: do we have an error path for stored cards not fetching?

mtgriego: not really, i think we would just throw out a generic error, but ill double check to see what the current flow does

Arkariang: on PledgeFragment we do the generic error, red snackbar with "something went wrong". Same pattern could be use throughout the late pledge chedkout flow. And can be a separated ticket :)

leighdouglas: I'm cool with it being a separate ticket, just something that im currently trying to address in the QA script and was curious how we were handling it

Arkariang (dismissed):

Arkariang (approved):

app/src/main/java/com/kickstarter/viewmodels/projectpage/LatePledgeCheckoutViewModel.kt
Arkariang: [MBL-1306](https://kickstarter.atlassian.net/browse/MBL-1306) ticket here

PR #1995, opened 2024-03-28
[MBL-1303] Add adding payment method to the checkout screen

mtgriego:
# 📲 What

Allows the user to add a payment method and use it in the late pledge checkout

# 🤔 Why

Users may need to add new methods if their current one isnt accepted, etc

# 🛠 How

Borrowing stripe interactions from pledgefragment

# 👀 See

Screenshot_20240328-151847

https://github.com/kickstarter/android-oss/assets/7563288/9584eef2-7e47-4ccb-a1f8-2b9f98e13fd1

# 📋 QA

Go to any late pledge project and get to the checkout screen, from there, add a new payment method and attempt to use to check out.  Watch logs to see if all goes well.  If not, you should be presented with an error

# Story 📖

[MBL-1303](https://kickstarter.atlassian.net/browse/MBL-1303)

[MBL-1303]: https://kickstarter.atlassian.net/browse/MBL-1303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ProjectPageActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ProjectPageActivity.kt
@@ -54,10 +54,12 @@ import com.kickstarter.libs.utils.ApplicationUtils
 import com.kickstarter.libs.utils.ViewUtils
 import com.kickstarter.libs.utils.extensions.addToDisposable
 import com.kickstarter.libs.utils.extensions.getEnvironment
+import com.kickstarter.libs.utils.extensions.getPaymentSheetConfiguration
 import com.kickstarter.libs.utils.extensions.showLatePledgeFlow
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/ui/activities/ProjectPageActivity.kt
```diff
@@ -1085,9 +1123,96 @@ class ProjectPageActivity :
         }
     }
 
+    // Update the UI with the returned PaymentOption
```
Arkariang, line 1126: For everything related with the PaymentSheet, we should try instead of copy and paste the code from PledgeFragment, to create some kind of Extension function or utils for Stripe's integration in a way the is re-usable.

PR #2009, opened 2024-04-03
MBL-1325: Rewards that are unavailable for late pledging are still selectable

leighdouglas:
# 📲 What

Rewards that are marked as unavailable for late pledging were still selectable, you could get through the entire checkout until time to actually pledge, and then an error was returned. This PR fixes that. 

# 🤔 Why

Bugs are bad mmkay

# 🛠 How

Updated the conditions for checking the button availability to include checking the isAvailable field. This part of the code was designed accommodate manage pledge and other use cases, hence the extra conditions, but right now this screen is only viewable on the late pledge flow. should cleanup and refactor later tho

# 👀 See

https://github.com/kickstarter/android-oss/assets/19390326/acd13912-3e93-4f30-bc21-13d5b0be39d0

# 📋 QA

Open the project "zan's late pledge" on staging, open the rewards carousel, should see correct available and unavailable rewards

# Story 📖

[MBL-1325: Rewards that are unavailable for late pledging are still selectable](https://kickstarter.atlassian.net/browse/MBL-1325)


mtgriego (approved):

app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/RewardCarouselScreen.kt
```diff
@@ -144,19 +144,24 @@ fun RewardCarouselScreen(
             ) { reward ->
 
                 val ctaButtonEnabled = when {
+                    RewardUtils.isNoReward(reward) && (project.postCampaignPledgingEnabled() ?: false && project.isInPostCampaignPledgingPhase() ?: false) -> true
```
Arkariang, line 147: fyi, no need to act on it now, but all this logic could be moved to an extension function or the rewards utils, and unit tested, so we are sure future changes won't break anything

PR #2026, opened 2024-04-25
[MBL-1383] Update adding new payment method flow in late pledge flow

mtgriego:
# 📲 What

Change it so that instead of creating a payment intent and then using that to set up a card via stripe (which would charge immediately) we are using a setupintent and saving the card to backend, similar to what we do in the account card screen.  We then refresh the users stored cards so they show on the list.

# 🤔 Why

When doing the payment intent way it would charge cards immediately, causing double charges via stripe

# 🛠 How

Changed the code to save the new cards instead of have a temporary copy via payment intent

# 📋 QA

1) Go to a project that is accepting late pledges ("testing conversion" on staging)
2) Proceed to the checkout screen
3) add a new card
4) proceed with checkout with the new card
5) Confirm card was only charged once (this may require backend help)

# Story 📖

[MBL-1383](https://kickstarter.atlassian.net/browse/MBL-1383)

[MBL-1383]: https://kickstarter.atlassian.net/browse/MBL-1383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/projectpage/LatePledgeCheckoutViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/projectpage/LatePledgeCheckoutViewModel.kt
@@ -11,14 +11,13 @@ import com.kickstarter.models.Project
 import com.kickstarter.models.Reward
 import com.kickstarter.models.ShippingRule
 import com.kickstarter.models.StoredCard
+import com.kickstarter.services.mutations.SavePaymentMethodData
 import com.kickstarter.ui.data.CheckoutData
 import com.kickstarter.ui.data.PledgeData
 import com.kickstarter.ui.data.ProjectData
-import com.stripe.android.ApiResultCallback
 import com.stripe.android.Stripe
 import com.stripe.android.confirmPaymentIntent
 import com.stripe.android.model.ConfirmPaymentIntentParams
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kickstarter/viewmodels/projectpage/LatePledgeCheckoutViewModel.kt
```diff
@@ -149,40 +131,39 @@ class LatePledgeCheckoutViewModel(val environment: Environment) : ViewModel() {
...
 
-    fun onPledgeButtonClicked(selectedCard: StoredCard?, project: Project, totalAmount: Double) {
-        if (selectedCard == newStoredCard) {
-            stripe.retrievePaymentIntent(
-                clientSecret = clientSecretForNewCard,
-                callback = object : ApiResultCallback<PaymentIntent> {
-                    override fun onError(e: Exception) {
-                        errorAction.invoke(null)
-                    }
+    private suspend fun refreshUserCards() {
```
Arkariang, line 152: This one doesn't need to be a suspend function I think

Arkariang (approved):
tested and reviewed with bessie only one charged for my created payment intent

PR #2030, opened 2024-05-06
MBL-1326 bonus support input

ycheng-kickstarter:
# 📲 What

In late pledge flow, allow user to input bonus support amount, not just step increase/decrease

# 🤔 Why

Match checkout flow.

# 🛠 How

Ended up choosing BasicTextField over TextField but this came with tradeoffs. 
Pros: BasicTextField allowed me to use Modifier.width(IntrinsicSize.Min) to properly wrap-content for width and end-align to row. I was struggling to not have a TextField take up the entire width of row otherwise.
Cons: BasicTextField does not allow a hint, which can make for a poor user experience if you try to delete a "0" amount. e.g. If the initial bonus amount is 0 and I try to delete it to enter amount "2", it enters "20" instead.

# 👀 See

| Light mode | Dark mode |
| --- | --- |
| Screenshot_20240506-092355 | Screenshot_20240506-092329 |

# 📋 QA

Go through the late pledge flow and input bonus support amount on Confirm Pledge Details screen.

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-1378


app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/ConfirmPledgeDetailsScreen.kt
```diff
@@ -565,22 +605,40 @@ fun BonusSupportContainer(
...
+                    start = dimensions.paddingXSmall,
+                    top = dimensions.paddingXSmall,
+                    bottom = dimensions.paddingXSmall,
+                    end = dimensions.paddingXSmall
+                ),
+
+            ) {
+                Text(
+                    text = currencySymbolAtStart ?: "",
+                    color = colors.kds_create_700
```
ycheng-kickstarter, line 623: @mtgriego KSColors tells me this is an old color, so pls let me know what I should be using instead.

app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/ConfirmPledgeDetailsScreen.kt
```diff
@@ -565,22 +605,40 @@ fun BonusSupportContainer(
...
+                )
+                BasicTextField(
+                    modifier = Modifier.width(IntrinsicSize.Min),
+                    value = if (totalBonusSupport % 1.0 == 0.0) totalBonusSupport.toInt().toString() else totalBonusSupport.toString(),
+                    onValueChange = {
+                        if (it.length <= bonusAmountMaxDigits) onBonusSupportInputted(it.parseToDouble())
+                    },
+                    textStyle = typography.title1.copy(color = colors.kds_create_700),
+                    keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Number),
+                    cursorBrush = SolidColor(Color.LightGray)
```
ycheng-kickstarter, line 633: @mtgriego If I needed this to change for Light/Dark theme what would I change?

app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/ConfirmPledgeDetailsScreen.kt
mtgriego: As this is text, it should probably be `textAccentGreen`, i believe thats the closest new green for text we have

app/src/main/java/com/kickstarter/libs/utils/RewardViewUtils.kt
```diff
@@ -83,6 +83,43 @@ object RewardViewUtils {
...
+    }
+
+    /**
+     * Returns a Pair of SpannableString and a Boolean where the
+     * SpannableString represents a project's currency symbol that shrinks currency symbol if it's necessary and the
+     * Boolean represents whether the symbol should be shown at the end or the start of the currency.
+     * Special case: US people looking at US currency just get the currency symbol.
+     *
+     */
+    fun currencySymbolAndPosition(project: Project, ksCurrency: KSCurrency): Pair<SpannableString, Boolean> {
```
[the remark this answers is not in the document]
Arkariang: Any reason why we could not use ProjectViewUtils method? or if it doesn't fit within "Project" or "rewards" just extract to to a generic extension function of ViewUtils and do not duplicate code

app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/ConfirmPledgeDetailsScreen.kt
mtgriego: Pretty much anything you want to change with the system theme needs to use the theme colors we created.
In this case I see you are using out theme colors for the text (but use `textAccentGreen` here as well), but the cursor brush is using a defined color.  Instead of `Color.LightGray` you should use one of our colors (in this case perhaps` iconSubtle`. 
 Not all of our color names match their use cases yet).

app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/ConfirmPledgeDetailsScreen.kt
```diff
@@ -565,22 +605,40 @@ fun BonusSupportContainer(
...
+                    end = dimensions.paddingXSmall
+                ),
+
+            ) {
+                Text(
+                    text = currencySymbolAtStart ?: "",
+                    color = colors.kds_create_700
+                )
+                BasicTextField(
+                    modifier = Modifier.width(IntrinsicSize.Min),
```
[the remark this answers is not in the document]
mtgriego: From what I've been able to tell textfield is a bit limited in customization, so I am not sure, but I can help dig for a solution for the future.  I ran into a couple things I could quite change with the normal textfield and switch to basictextfield for the country input too.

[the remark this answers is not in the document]
ycheng-kickstarter: Ok glad we had similar experiences!

mtgriego:
Screenshot_20240506-152904
I was doing a bit of testing and it looks like the amount is not being compared correctly to the maximum amount somewhere.

mtgriego:
For the above screenshot, it looks like the maximum amount its allowing is 7835 euro, when converting that to USD thats about 8500, so perhaps the currency conversion is not working correctly

mtgriego:
https://github.com/kickstarter/android-oss/assets/7563288/b5333047-ee10-44ba-b266-7d66b82fceea

also seeing some unusual behaviour here:
1) Go to a late pledge and select the no reward option
2) In the text field move the cursor behind the 1
3) click zero then backspace repeatedly

I think something similar happens on the current bonus support text box too

app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/ConfirmPledgeDetailsScreen.kt
```diff
@@ -741,3 +806,33 @@ fun ItemizedRewardListContainer(
         }
     }
 }
+
+@Composable
```
mtgriego, line 810: Methods that have a return should not be composable.  We can probably either inline this or make it a util function that takes in a context to get the string.

mtgriego (changes requested):
Overall pretty good, just one comment for a method introduced.

app/src/main/java/com/kickstarter/viewmodels/projectpage/ConfirmDetailsViewModel.kt
```diff
@@ -312,7 +312,7 @@ class ConfirmDetailsViewModel(val environment: Environment) : ViewModel() {
             ConfirmDetailsUIState(
                 rewardsAndAddOns = rewardAndAddOns,
                 initialBonusSupportAmount = initialBonusSupport,
-                totalBonusSupportAmount = initialBonusSupport + addedBonusSupport,
```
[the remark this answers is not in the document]
Arkariang: Seems worth to me to extract this piece in a method
`if (addedBonusSupport > 0) addedBonusSupport else initialBonusSupport`

and add as comment your explanation up there ☝️, so we don't loose that info living just on the Pr 💖

app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/ConfirmPledgeDetailsScreen.kt
Arkariang: could perfectly be a extracted to `RewardViewUtils`

mtgriego (approved):

PR #2038, opened 2024-05-14
MBL-1403 Stripe Link

ycheng-kickstarter:
# 📲 What

Implement Stripe Link

# 🤔 Why

To enable 1-click checkout 

# 🛠 How

Update Stripe sdk version to 20.42.0 
Pass in user email to PaymentSheet.Configuration
Hide Stripe Link on user's payment methods settings screen to match iOS and web behavior

# 👀 See
User adds a new payment method and can check the box at the bottom that saves the info to Link OR click on green Pay with Link button on top: 

A return user sees the option to Pay with Link with their associated email:

Users should not see Link enabled in their settings payment methods. Unfortunately there isn't a clientside way to turn it completely off; instead they'll see an extra email field which is not ideal:

# 📋 QA

Manually override the feature flag value in ContextExt.kt to make it easier to QA for now.
**When you get to the verification step in the Stripe Link checkout flow, enter 000000** 

Try out the new Stripe Link flow for adding new payment methods on 
- Pledge flow
- Late pledge flow

Stripe Link should NOT be an option on the payment methods settings screen
- Settings-> Account -> Payment methods

Reminder that test cards can be found here: https://docs.stripe.com/testing

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-1403


app/src/main/java/com/kickstarter/ui/activities/compose/FormularyScreen.kt
```diff
@@ -112,7 +112,7 @@ fun FormularyScreen(
             value = outputs.email().subscribeAsState(initial = "").value,
             onValueChange = {},
             readOnly = true,
-            label = { Text(stringResource(id = R.string.email)) },
```
ycheng-kickstarter, line 115: Can someone explain to me where this `R.string.email` string went? It's not in strings_i18n.xml

mtgriego (approved):

Arkariang (approved):

PR #2042, opened 2024-05-21
[MBL-1498] Fix updating payment method/fixing payment method for late pledge campaign

mtgriego:
# 📲 What

There is an edge case where:
1) User backs a campaign during the normal campaign time
2) The reward chosen changes price for late pledges
3) The users card has an error for some reason
4) User tries to update their card

In this scenario currently there is an issue with the graph call we make.

Currently we call the update backing endpoint, but there are two issues:
1) The price was based on the reward ids, which since the price changes, this was reflected incorrectly on the manage pledge/re-do checkout screen
2) This call would not work in the scenario since the pledge is locked, but the campaign is still running

So the solution was:
1) Update the call we are using to not pass in the amount

# 🤔 Why

See above

# 🛠 How

Add logic to look at the type of backing (late pledge or non late pledge) and adjust the displayed pricing and dont pass amount to the graph call if the amount has not changed.

# 👀 See

Screenshot_20240528-152614
Screenshot_20240528-152610

# 📋 QA

This is a bit tricky, you need to
1) Back a project using a reward that will change price during the post campaign phase using a card that will fail (see stripe test cards)
2) Have the project end
3) Go to update your payment method via the project page or via the activity warning
4) Update your payment method

A few more things to cover since this is touching payments:
Make sure updating your pledge for a currently running campaign still works
- Updating bonus amount
- Changing reward selection
- Changing payment methods
- Cancelling the pledge

Any backing in post campaign will not run into this issue as they are charged immediately, but be sure everything is still displayed correctly in the "view my pledge" page for a post campaign backing

# Story 📖

[MBL-1498](https://kickstarter.atlassian.net/browse/MBL-1498)

[MBL-1498]: https://kickstarter.atlassian.net/browse/MBL-1498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/viewmodels/PledgeFragmentViewModel.kt
```diff
@@ -1690,11 +1695,29 @@ interface PledgeFragmentViewModel {
...
-        private fun getPledgeAmount(rewards: List<Reward>): Double {
+        private fun getPledgeAmount(rewards: List<Reward>, isLatePledge: Boolean): Double {
             var totalPledgeAmount = 0.0
             rewards.forEach {
-                totalPledgeAmount += if (RewardUtils.isNoReward(it) && !it.isAddOn()) it.minimum() // - Cost of the selected Reward
-                else it.quantity()?.let { q -> (q * it.minimum()) } ?: it.minimum() // - Cost of each addOn
+                totalPledgeAmount += if (isLatePledge) {
+                    if (it.latePledgeAmount() > 0) {
+                        if (RewardUtils.isNoReward(it) && !it.isAddOn()) it.latePledgeAmount() // - Cost of the selected Reward
+                        else it.quantity()?.let { q -> (q * it.latePledgeAmount()) } ?: it.latePledgeAmount() // - Cost of each addOn
```
Arkariang, line 1704: Question here, on line 1704, does addOns have latePledgeAmount associated as well?

Arkariang (approved):

PR #2044, opened 2024-05-23
MBL-1394 Remove RXJava usage from new AddOnsViewModel

ycheng-kickstarter:
# 📲 What

Refactor the AddOnsViewModel  to remove all RxJava usages.

# 🤔 Why

AddOnsViewModel was originally written with a mix of RxJava and Coroutines, and we no longer want it to contain RxJava.

# 🛠 How

# 👀 See

No user facing changes.

# 📋 QA

Open a late pledges project and confirm the AddOnsVM is working the same as before.

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-1394


app/src/test/java/com/kickstarter/viewmodels/AddOnsViewModelTest.kt
```diff
@@ -58,18 +58,17 @@ class AddOnsViewModelTest : KSRobolectricTestCase() {
         viewModel.userRewardSelection(reward)
 
         assertEquals(
-            uiState.last(),
-            AddOnsUIState(
-                shippingSelectorIsGone = true
-            )
```
ycheng-kickstarter, line 64: This `test_hide_location_selection_on_reward_not_shippable` test and `test_show_location_selection_on_reward_is_shippable` are practically identical aside from the shipping type, but this test somehow does not pass, so I ended up comparing the specific shippingSelectorIsGone field instead.

app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
```diff
@@ -114,13 +76,9 @@ class AddOnsViewModel(val environment: Environment) : ViewModel() {
                     apolloClient.getShippingRules(
                         reward = reward
                     ).subscribe { shippingRulesEnvelope ->
```
Arkariang, line 78: There is still a subsciption and disposable being used here. We should not use RXJava, but `.asFlow` and use flow streams.
Screenshot 2024-05-28 at 5 50 44 PM

[Take a look to this document in case of doubt ](https://kickstarter.atlassian.net/wiki/spaces/NT/pages/2602729524/Coroutines+KS+Best+practices#Consume-Network-Layer-still-using-RXJava2)

ycheng-kickstarter: Oops good catch. Thanks Isa.

Arkariang (approved):

PR #2059, opened 2024-06-18
MBL-1474: PPO Message creator navigation

leighdouglas:
# 📲 What

Added message creator navigations to the PPO card

# 🤔 Why

PPO work

# 🛠 How

When message creator is tapped, performs a network call to grab the project/backing and passes this on to the message activity. 

# 👀 See

https://github.com/kickstarter/android-oss/assets/19390326/3080ad16-36cb-4b73-9ac2-47cbb7d981e4

# 📋 QA

Tested using mocked data, so you will need to grab the slug of one of your backed projects. in the `PledgedProjectsOverviewViewModel`, replace the ppo card list:
Screenshot 2024-06-18 at 22 26 41
with your slug:
```private val ppoCards = MutableStateFlow<PagingData<PPOCardDataMock>>(PagingData.from(listOf(PPOCardDataMock(projectSlug = "insert slug here"))))```
inserting your own slug
to test errors, leave the slug as an empty string or pass in a random string of characters

# Story 📖

[MBL-1474: Message creator navigation](Trello link)

### DIFF
--- a/app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewActivity.kt
@@ -1,6 +1,5 @@
 package com.kickstarter.features.pledgedprojectsoverview.ui
 
-import android.content.Intent
 import android.os.Build
 import android.os.Bundle
 import androidx.activity.OnBackPressedCallback
@@ -16,17 +15,13 @@ import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Modifier
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.paging.compose.collectAsLazyPagingItems
-import com.kickstarter.R
 import com.kickstarter.features.pledgedprojectsoverview.viewmodel.PledgedProjectsOverviewViewModel
 import com.kickstarter.libs.MessagePreviousScreenType
 import com.kickstarter.libs.utils.TransitionUtils
 import com.kickstarter.libs.utils.extensions.getEnvironment
 import com.kickstarter.libs.utils.extensions.isDarkModeEnabled
-import com.kickstarter.models.Project
-import com.kickstarter.ui.IntentKey
 import com.kickstarter.ui.SharedPreferenceKey
 import com.kickstarter.ui.activities.AppThemes
-import com.kickstarter.ui.activities.MessageCreatorActivity
 import com.kickstarter.ui.compose.designsystem.KickstarterApp
 import com.kickstarter.ui.extensions.startCreatorMessageActivity
 import com.kickstarter.ui.extensions.transition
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewActivity.kt [hunk 1]
```diff
@@ -60,11 +68,28 @@ class PledgedProjectsOverviewActivity : AppCompatActivity() {
...
+                    viewModel.projectFlow
+                        .collect {
+                            startCreatorMessageActivity(it, previousScreen = MessagePreviousScreenType.PLEDGED_PROJECTS_OVERVIEW)
+                        }
+                }
+
+                val coroutineScope = rememberCoroutineScope()
+
+                viewModel.provideSnackbarMessage {
+                    coroutineScope.launch {
```
leighdouglas, line 88: Not sure if this is the best way to do this, but `showSnackBar` must be called from a coroutine or other suspend function.

mtgriego, line 88: I recommend you use `lifecycleScope` here as it is more aware of the activity lifecycle and will close the job if the activity/fragment dies:
https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope

mtgriego (changes requested):
Overall very good!  One comment about the coroutine scope used, but otherwise pretty solid

app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewActivity.kt
mtgriego: You can see my comment for a recommendation/link for more info on what to use here

leighdouglas: Ah okay, i did see that as another option, though I wasn't sure if it was overkill. I'll definitely update!!

mtgriego (approved):

app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewScreen.kt
```diff
@@ -58,14 +62,21 @@ fun PledgedProjectsOverviewScreen(
...
+    onSendMessageClick: (projectName: String) -> Unit
 ) {
     Box(
         modifier = Modifier.fillMaxSize(),
         contentAlignment = Alignment.Center
     ) {
         Scaffold(
+            snackbarHost = {
+                SnackbarHost(
+                    hostState = errorSnackBarHostState
```
ycheng-kickstarter, line 77: Hey! If this is indeed supposed to be an error snackbar should we use the design system's KSErrorRoundedText in KSAlerts.kt?

leighdouglas: Good callout, but I opted not to for now since we can only attach one snackbar to the screen in this way, and we may need to use this snackbar for more than one type of non-error message in subsequent work. I figured I'd spend time figuring out how to show more than one type of snackbar in a subsequent PR, and i should have renamed the this to be more generic instead of errorSnackbarHostState

ycheng-kickstarter: Ah, well you see that is the issue I'm running into right now 😅 since my Confirm Address flow also involves a snackbar

PR #2062, opened 2024-06-21
[No Ticket] Add Support for Surveys via Deeplinks

mtgriego:
# 📲 What

Adds support for survey deeplinks for emails etc

# 🤔 Why

We will soon be needed surveys linked into the app

# 🛠 How

Add the links to the supported list, redirect based on log in status

# 📋 QA

Kinda tricky, but you need to be logged in as a user who has a survey for a project available.  Click the link for the survey (usually from an email) and see that it open in app or asks you to log in if not already logged in before being redirected to the survey page.

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/DeepLinkActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/DeepLinkActivity.kt
@@ -10,6 +10,7 @@ import android.view.animation.AnticipateInterpolator
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import com.kickstarter.R
+import com.kickstarter.libs.ActivityRequestCodes
 import com.kickstarter.libs.RefTag
 import com.kickstarter.libs.utils.ApplicationUtils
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/kickstarter/ui/activities/DeepLinkActivity.kt
```diff
@@ -109,6 +112,24 @@ class DeepLinkActivity : AppCompatActivity() {
...
                 startPreLaunchProjectActivity(it.first, it.second, "DEEPLINK")
             }.addToDisposable(disposables)
+
+        viewModel.outputs.startProjectSurvey()
+            .observeOn(AndroidSchedulers.mainThread())
+            .subscribe {
+                val uri = it.first
+                val isLoggedIn = it.second
+                val surveyResponse =
+                    SurveyResponse.builder().urls(
```
Arkariang, line 122: We could move this logic to the VM, and emit as outputs `startSurveyResponseActivity` and `startLoginForSurveys`, and add test for both use cases for maintainance sake

Arkariang (dismissed):

Arkariang (approved):

PR #2066, opened 2024-07-01
Leigh/mbl 1464 ppo query

leighdouglas:
# 📲 What

A description of the change.

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖


app/src/main/graphql/fragments.graphql
```diff
@@ -362,6 +362,32 @@ fragment shippingRule on ShippingRule {
     }
 }
 
+
+
+fragment ppoCard on Backing {
```
leighdouglas, line 367: As of right now, each card only has the backing in the response from graph, so im retrieving the information we need from it with this fragment. this is subject to change depending on how the rest of the response is structured

app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewScreen.kt
```diff
@@ -144,8 +144,13 @@ fun PledgedProjectsOverviewScreen(
                             showBadge = it.showBadge,
                             onActionButtonClicked = { },
                             onSecondaryActionButtonClicked = {
-                                confirmedAddress = it.shippingAddress
-                                openConfirmAddressAlertDialog.value = true
+                                when (it.viewType) {
+                                    PPOCardViewType.CONFIRM_ADDRESS -> {
```
leighdouglas, line 148: Amending this to only call this particular action when the card type is confirm address.

app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewScreen.kt
```diff
@@ -131,23 +134,28 @@ fun PledgedProjectsOverviewScreen(
...
-                            shippingAddress = it.shippingAddress,
-                            showBadge = it.showBadge,
+                            onProjectPledgeSummaryClick = { onProjectPledgeSummaryClick(it.backingDetailsUrl() ?: "") },
+                            projectName = it.projectName(),
+                            pledgeAmount = it.amount(),
+                            imageUrl = it.imageUrl(),
+                            imageContentDescription = it.imageContentDescription(),
+                            creatorName = it.creatorName(),
+                            sendAMessageClickAction = { onSendMessageClick(it.projectSlug() ?: "") },
+                            shippingAddress = it.address()?.addressLine1() ?: "", // TODO replace with formatted address from PPO response
```
leighdouglas, line 146: we will need to update this with a formatted string. for now im just grabbing the shipping address from the backing object and using the first line, but this should be different when the backend piece is complete (we will either receive a formatted string or will have to format it ourselves).

mtgriego (approved):

ycheng-kickstarter (approved):

PR #2073, opened 2024-07-15
MBL-1479: Fix payment card - fix pledge CTA flow

leighdouglas:
# 📲 What

- Tapping fix payment on fix payment card now opens the manage pledge flow
- If a user successfully changes their payment method, when returning to the PPO screen, the list will requery/refresh

# 🤔 Why

PPO!

# 🛠 How

Manage pledge opens with manage pledge sheet expanded. 
ActivityResultLauncher is used to track when the user completes changing payment, ppo list is refreshed

# 👀 See

Please note that since this is still mock data, the refreshing behavior doesn't actually update the list in this PR, which is why you see still see the card in the list after it reloads. The empty state is also briefly shown. This will change in subsequennt PR's, but the refreshing action is there. 

https://github.com/user-attachments/assets/8f4e67fb-a564-40a9-bf3f-5f5bf56ea5e6

# 📋 QA

This is kinda tricky to test.

1. Back a project in staging with an attachable errored card - `4000000000000341`
2. Modify the fields in the mock card created in `PPOCardFactory.fixPayment()` to have the data from the project you just backed, most importantly the slug
3. Replace line 37 in `PledgedProjectsOverviewViewModel` with     
```
private val mutablePpoCards = MutableStateFlow<PagingData<PPOCard>>(PagingData.from(listOf(PPOCardFactory.fixPaymentCard())))
```
4. Set `totalAlerts = 1` in `PledgedProjectsOverviewUIState()`
5. Run the app, tap on project alerts, tap fix pledge, go through manage pledge flow, return to the PPO screen, should see the loading spinner for a brief moment
6. If you do not fix the pledge, when you return to the PPO screen, you should not see a loading spinner or any change

# Story 📖

[MBL-1479: Fix payment card - fix pledge CTA flow](https://kickstarter.atlassian.net/browse/MBL-1479)


Arkariang (approved):
👍

PR #2078, opened 2024-07-22
[MBL-821] Upgrade profile activity to RX2

mtgriego:
# 📲 What

Remove RX1 and replace it with RX2 in profile activity/viewmodel

# 🤔 Why

RX1 is old and busted, we need the new hotness

# 📋 QA

Confirm profile activity still works as normal

# Story 📖

[MBL-821](https://kickstarter.atlassian.net/browse/MBL-821)

[MBL-821]: https://kickstarter.atlassian.net/browse/MBL-821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ProfileActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ProfileActivity.kt
@@ -2,29 +2,37 @@ package com.kickstarter.ui.activities
 
 import android.content.Intent
 import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.viewModels
 import androidx.core.view.isGone
 import androidx.recyclerview.widget.GridLayoutManager
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.kickstarter.R
 import com.kickstarter.databinding.ProfileLayoutBinding
-import com.kickstarter.libs.BaseActivity
-import com.kickstarter.libs.RecyclerViewPaginator
-import com.kickstarter.libs.qualifiers.RequiresActivityViewModel
-import com.kickstarter.libs.rx.transformers.Transformers.observeForUI
+import com.kickstarter.libs.recyclerviewpagination.RecyclerViewPaginatorV2
+import com.kickstarter.libs.rx.transformers.Transformers.observeForUIV2
 import com.kickstarter.libs.utils.ApplicationUtils
 import com.kickstarter.libs.utils.ViewUtils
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/ui/activities/ProfileActivity.kt
```diff
@@ -2,126 +2,138 @@ package com.kickstarter.ui.activities
...
 import com.kickstarter.ui.IntentKey
 import com.kickstarter.ui.adapters.ProfileAdapter
 import com.kickstarter.ui.extensions.loadCircleImage
+import com.kickstarter.ui.extensions.startActivityWithTransition
 import com.kickstarter.viewmodels.ProfileViewModel
+import io.reactivex.disposables.CompositeDisposable
 
-@RequiresActivityViewModel(ProfileViewModel.ViewModel::class)
-class ProfileActivity : BaseActivity<ProfileViewModel.ViewModel>() {
+class ProfileActivity : ComponentActivity() {
```
leighdouglas, line 27: are we clearing the disposables anywhere here?

app/src/main/java/com/kickstarter/viewmodels/ProfileViewModel.kt
```diff
@@ -80,13 +82,15 @@ interface ProfileViewModel {
         fun userNameTextViewText(): Observable<String>
     }
 
-    class ViewModel(environment: Environment) : ActivityViewModel<ProfileActivity>(environment), ProfileAdapter.Delegate, Inputs, Outputs {
-        private val client = requireNotNull(environment.apiClient())
-        private val currentUser = requireNotNull(environment.currentUser())
+    class ProfileViewModel(environment: Environment) : ViewModel(), ProfileAdapter.Delegate, Inputs, Outputs {
```
leighdouglas, line 85: same question as above, we should be clearing the disposables when the viewmodel is destroyed

leighdouglas (commented):
Left a

app/src/main/java/com/kickstarter/ui/activities/ProfileActivity.kt
mtgriego: good catch, forgot about that

Arkariang (approved):

PR #2080, opened 2024-07-23
[MBL-1480] Add error snackbar visuals/a way to call them for PPO

mtgriego:
# 📲 What

Added the visuals and a way to call the visuals for an error message for PPO

I also updated some of the colors for these to match the newer standards

# 🤔 Why

There can be various times we need to display an error to the user

# 🛠 How

In this case I leverage that we can intervep the snackbarhost data and use the actionLabel in this case to change which visuals are shown

# 👀 See

Alternating between an error and heads up message type
https://github.com/user-attachments/assets/39a03479-610b-4aee-afbe-0a0f9f7ff065

# 📋 QA

You will need to make a way to trigger these calls (i made the back action invoke a method in the viewmodel to display an error and normal message alternatingly)

# Story 📖

[MBL-1480](https://kickstarter.atlassian.net/browse/MBL-1480)

[MBL-1480]: https://kickstarter.atlassian.net/browse/MBL-1480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewActivity.kt
@@ -2,7 +2,6 @@ package com.kickstarter.features.pledgedprojectsoverview.ui
 
 import android.app.Activity
 import android.content.Intent
-import android.os.Build
 import android.os.Bundle
 import androidx.activity.OnBackPressedCallback
 import androidx.activity.compose.setContent
@@ -10,7 +9,6 @@ import androidx.activity.result.ActivityResultLauncher
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
-import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material.SnackbarHostState
 import androidx.compose.runtime.LaunchedEffect
@@ -67,7 +65,6 @@ class PledgedProjectsOverviewActivity : AppCompatActivity() {
 
                 val ppoUIState by viewModel.ppoUIState.collectAsStateWithLifecycle()
 
-                val darkModeEnabled = this.isDarkModeEnabled(env = env)
                 val lazyListState = rememberLazyListState()
                 val snackbarHostState = remember { SnackbarHostState() }
                 val totalAlerts = viewModel.totalAlertsState.collectAsStateWithLifecycle().value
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/viewmodel/PledgedProjectsOverviewViewModel.kt
```diff
@@ -179,6 +180,14 @@ class PledgedProjectsOverviewViewModel(environment: Environment) : ViewModel() {
         )
     }
 
+    private fun showHeadsUpSnackbar(messageId: Int) {
+        snackbarMessage.invoke(messageId, KSSnackbarTypes.KS_HEADS_UP.name)
+    }
+
+    private fun showErrorSnackbar(messageId: Int) {
+        snackbarMessage.invoke(messageId, KSSnackbarTypes.KS_ERROR.name)
+    }
```
mtgriego, line 189: Depending on if its an error or normal message you can call one of these methods to display it

app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewScreen.kt
```diff
@@ -172,7 +175,16 @@ fun PledgedProjectsOverviewScreen(
...
-                    hostState = errorSnackBarHostState
+                    hostState = errorSnackBarHostState,
+                    snackbar = { data ->
+                        // Action label is typically for the action on a snackbar, but we can
+                        // leverage it and show different visuals depending on what we pass in
+                        if (data.actionLabel == KSSnackbarTypes.KS_ERROR.name) {
+                            KSErrorSnackbar(text = data.message)
+                        } else {
+                            KSHeadsupSnackbar(text = data.message)
+                        }
```
mtgriego, line 186: Wanted to re-use our visuals from the design system here

app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewActivity.kt
```diff
@@ -79,17 +76,7 @@ class PledgedProjectsOverviewActivity : AppCompatActivity() {
...
-                        when (theme) {
-                            AppThemes.MATCH_SYSTEM.ordinal -> isSystemInDarkTheme()
-                            AppThemes.DARK.ordinal -> true
-                            AppThemes.LIGHT.ordinal -> false
-                            else -> false
-                        }
-                    } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-                        isSystemInDarkTheme()
-                    } else false
+                    useDarkTheme = isDarkModeEnabled(env = env)
```
mtgriego, line 79: noticed this was calling this:
`@Composable
fun Context.isDarkModeEnabled(env: Environment): Boolean {
    val darkModeEnabled = env.featureFlagClient()?.getBoolean(FlagKey.ANDROID_DARK_MODE_ENABLED) ?: false
    val theme = env.sharedPreferences()
        ?.getInt(SharedPreferenceKey.APP_THEME, AppThemes.MATCH_SYSTEM.ordinal)
        ?: AppThemes.MATCH_SYSTEM.ordinal

    return if (darkModeEnabled) {
        when (theme) {
            AppThemes.MATCH_SYSTEM.ordinal -> isSystemInDarkTheme()
            AppThemes.DARK.ordinal -> true
            AppThemes.LIGHT.ordinal -> false
            else -> false
        }
    } else if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
        isSystemInDarkTheme() // Force dark mode uses system theme
    } else false
}`

so i just inlined it

leighdouglas (approved):
🎉

Arkariang (approved):

PR #2083, opened 2024-07-29
[MBL-800] Update survey viewholder viewmodel to RX2

mtgriego:
# 📲 What

Update the SurveyViewHolder Viewmodel to RX2

# 🤔 Why

We need to get rid of RX1

# 🛠 How

🧠 

# 👀 See

# 📋 QA

Make sure the viewholder in the activity page works as expected

# Story 📖

[MBL-800](https://kickstarter.atlassian.net/browse/MBL-800)

[MBL-800]: https://kickstarter.atlassian.net/browse/MBL-800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/viewholders/SurveyViewHolder.kt
+++ b/app/src/main/java/com/kickstarter/ui/viewholders/SurveyViewHolder.kt
@@ -6,17 +6,20 @@ import android.view.View
 import com.kickstarter.R
 import com.kickstarter.databinding.ActivitySurveyViewBinding
 import com.kickstarter.libs.rx.transformers.Transformers
+import com.kickstarter.libs.utils.extensions.addToDisposable
 import com.kickstarter.models.Project
 import com.kickstarter.models.SurveyResponse
 import com.kickstarter.ui.IntentKey
 import com.kickstarter.ui.activities.SurveyResponseActivity
 import com.kickstarter.ui.extensions.loadCircleImage
 import com.kickstarter.viewmodels.SurveyHolderViewModel
+import io.reactivex.disposables.CompositeDisposable
 
 class SurveyViewHolder(private val binding: ActivitySurveyViewBinding) :
     KSViewHolder(binding.root) {
     private val ksString = requireNotNull(environment().ksString())
-    private val viewModel: SurveyHolderViewModel.ViewModel = SurveyHolderViewModel.ViewModel(environment())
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/viewmodels/SurveyHolderViewModel.kt
```diff
@@ -34,12 +30,9 @@ interface SurveyHolderViewModel {
         fun startSurveyResponseActivity(): Observable<SurveyResponse>
     }
 
-    class ViewModel(@NonNull environment: Environment) :
-        ActivityViewModel<SurveyViewHolder>(environment),
-        Inputs,
-        Outputs {
+    class ViewModel : androidx.lifecycle.ViewModel(), Inputs, Outputs {
```
leighdouglas, line 33: i dont think we're supposed to inherit from android `ViewModel` since this is only a view holder and it doesn't have a dedicated activity/fragment and therefore no attachment lifecycle.

mtgriego: Good call, dont think it needs it so ill remove it!

app/src/main/java/com/kickstarter/ui/viewholders/SurveyViewHolder.kt
```diff
@@ -50,20 +53,25 @@ class SurveyViewHolder(private val binding: ActivitySurveyViewBinding) :
...
         viewModel.outputs.startSurveyResponseActivity()
-            .compose(bindToLifecycle())
-            .compose(Transformers.observeForUI())
+            .compose(Transformers.observeForUIV2())
             .subscribe { startSurveyResponseActivity(it) }
+            .addToDisposable(disposables)
+    }
+
+    override fun destroy() {
+        disposables.clear()
```
leighdouglas, line 74: we also need the clear the viewmodel subscriptions here, since we are not inheriting from android `ViewModel`

mtgriego: the viewmodel doesn't have any disposables to clear

leighdouglas: ohp you right!

leighdouglas (approved):
lgtm!

PR #2099, opened 2024-08-14
MBL-1635: Mock data to real data, and cleanup

leighdouglas:
# 📲 What

Small, mostly cosmetic cleanup for PPO
- Fix broken backing details page not loading
- Fix total alerts (0) showing when refreshing PPO screen
- Disable confirm address button when users address is null

All other cleanup is in separate tickets

# 🤔 Why

Cleanup cleanup

# 📋 QA

This will be tricky to test, and will require mock data. You can modify the ppo card factory and hardcode these cards into the graphql transformer function:

Testing backing details page:
- Add a mock card to the ppo list created in the graphql transformer function
- In the field for backing details url, you can use your own or this url "/projects/grimacecards/monochrome-muse-playing-cards/backing/details"
- Tapping on project title area of card should open a webview and load the backing details page
- Tapping "edit" on a confirm address or survey card should open a webview and load the backing details page

Testing total alerts heading
- When you have no alerts, visit the PPO page and swipe to refresh. You should not see the total alerts heading appear

Testing edit address button disabled
- Create a confirm address card with the PPOCardFactory, but leave the address field null. Add this to the ppo screen list in the graphql transformer function
- Go to the PPO screen, it should not display an address, and the edit button should be disabled and not clickable. 

# Story 📖

MBL-1635: Mock data to real data, and cleanup](https://kickstarter.atlassian.net/browse/MBL-1635)

### DIFF
--- a/app/src/main/graphql/fragments.graphql
+++ b/app/src/main/graphql/fragments.graphql
@@ -366,6 +366,7 @@ fragment shippingRule on ShippingRule {
 
 fragment ppoCard on Backing {
     id
+    backingDetailsPageUrl
     clientSecret
     amount {
         ...amount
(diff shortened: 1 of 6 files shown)


Arkariang (approved):
lgtm!

PR #2109, opened 2024-08-22
MBL-1659: Update card tags/pills to match backend schema and other small changes

leighdouglas:
# 📲 What

- PPO Flags are now rendered according to their formatting on the backing. Two types, "warning" and "alert".
- Other minor cosmetic updates: String extension method to format a string with a placeholder, and fixed a typo in a string
- Cleaned up ppo card model, fields are in alphabetical order

# 🤔 Why

PPO! And `environment.kt` is not invited to the compose party

# 🛠 How

PPO Flags have a type and icon, we will render the colors of the icon and background based on the type. The message comes from the backend. 

String ext method is taken from the KsStrings class, to avoid passing the environment or even just ksStrings into the compose function

Formatting "Alerts (#)" with this new string method

# 👀 See

Screenshot_20240822_183655 Screenshot_20240822_183739

# 📋 QA

Create a fix payment pledge
Open PPO screen
make sure alert count is correct, and the flags are displayed

# Story 📖

[MBL-1659: Update card tags/pills to match backend schema](https://kickstarter.atlassian.net/browse/MBL-1659)


app/src/test/java/com/kickstarter/ui/activities/compose/PPOCardViewKtTest.kt
```diff
@@ -120,17 +106,65 @@ class PPOCardViewKtTest : KSRobolectricTestCase() {
...
 
-        // Alerts
-        // TODO: Replace with translated strings
-        composeTestRule.onAllNodesWithText("Address locks in $timeNumberForAction hours")[0].assertIsDisplayed()
         // Shipping address hidden
         shippingAddressView.assertIsNotDisplayed()
         // CTA
         composeTestRule.onAllNodesWithText("Take Survey")[0].assertIsDisplayed()
     }
+
```
Arkariang, line 118: ❤️

Arkariang (approved):

PR #2119, opened 2024-09-03
[MBL-820] Update Privacy viewmodel to RX2

mtgriego:
# 📲 What

Updating Privacy Viewmodel to RX2

# 🤔 Why

Need to get rid of RX1

# 📋 QA

Go to the privacy activity and confirm everything works as expected

# Story 📖

[MBL-820](https://kickstarter.atlassian.net/browse/MBL-820)

[MBL-820]: https://kickstarter.atlassian.net/browse/MBL-820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
@@ -84,8 +84,8 @@ interface PrivacyViewModel {
             currentUser
                 .take(1)
                 .filter { it.isNotNull() && it.getValue().isNotNull() }
-                .map { it.getValue() }
-                .subscribe { this.userOutput.onNext(it!!) }
+                .map { requireNotNull(it.getValue()) }
+                .subscribe { this.userOutput.onNext(it) }
                 .addToDisposable(disposables)
 
             currentUser
@@ -105,7 +105,8 @@ interface PrivacyViewModel {
             updateSettingsNotification
                 .compose(errorsV2())
                 .filter { it.isNotNull() }
-                .subscribe { this.unableToSavePreferenceError.onNext(it!!) }
+                .map { it }
+                .subscribe { this.unableToSavePreferenceError.onNext(it) }
                 .addToDisposable(disposables)
 
             this.userInput
@@ -124,7 +125,8 @@ interface PrivacyViewModel {
             this.optIntoFollowing
                 .filter { checked -> checked }
                 .filter { this.userOutput.value.isNotNull() }
-                .subscribe { _ -> this.userInput.onNext(this.userOutput.value!!.toBuilder().social(true).build()) }
+                .map { requireNotNull(this.userOutput.value) }
+                .subscribe { this.userInput.onNext(it.toBuilder().social(true).build()) }
                 .addToDisposable(disposables)
 
             this.optIntoFollowing
@@ -135,7 +137,8 @@ interface PrivacyViewModel {
             this.optOutOfFollowing
                 .filter { optOut -> optOut }
                 .filter { this.userOutput.value.isNotNull() }
-                .subscribe { _ -> this.userInput.onNext(this.userOutput.value!!.toBuilder().social(false).build()) }
+                .map { requireNotNull(this.userOutput.value) }
+                .subscribe { this.userInput.onNext(it.toBuilder().social(false).build()) }
                 .addToDisposable(disposables)
 
             this.optOutOfFollowing
--- a/app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
@@ -1,20 +1,21 @@
 package com.kickstarter.viewmodels
 
-import androidx.annotation.NonNull
-import com.kickstarter.libs.ActivityViewModel
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
 import com.kickstarter.libs.Environment
 import com.kickstarter.libs.rx.transformers.Transformers
-import com.kickstarter.libs.rx.transformers.Transformers.errors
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
```diff
@@ -50,91 +51,97 @@ interface PrivacyViewModel {
...
+                .map { it.getValue() }
+                .subscribe { this.userOutput.onNext(it!!) }
+                .addToDisposable(disposables)
 
             currentUser
-                .compose(bindToLifecycle())
-                .filter { it.isNotNull() }
-                .map { user -> user.createdProjectsCount().isNonZero() }
-                .subscribe(this.hidePrivateProfileRow)
+                .filter { it.isNotNull() && it.getValue().isNotNull() }
```
leighdouglas, line 92: wouldnt this cause a crash if `it` / `currentUser` is null?

app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
```diff
@@ -50,91 +51,97 @@ interface PrivacyViewModel {
...
                 .subscribe { this.success(it) }
+                .addToDisposable(disposables)
 
             updateSettingsNotification
-                .compose(errors())
-                .compose(bindToLifecycle())
-                .subscribe(this.unableToSavePreferenceError)
+                .compose(errorsV2())
+                .filter { it.isNotNull() }
+                .subscribe { this.unableToSavePreferenceError.onNext(it!!) }
```
leighdouglas, line 108: we try to avoid the not-null forced assertion operator `!!`. we could use `.map { requireNotNull(it) }` after the `filter` instead

app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
```diff
@@ -50,91 +51,97 @@ interface PrivacyViewModel {
...
+                .compose<User>(Transformers.takeWhenV2<User, Throwable>(this.unableToSavePreferenceError))
+                .subscribe { this.userOutput.onNext(it) }
+                .addToDisposable(disposables)
 
             this.optIntoFollowing
-                .compose<Boolean>(bindToLifecycle<Boolean>())
                 .filter { checked -> checked }
-                .subscribe { _ -> this.userInput.onNext(this.userOutput.value.toBuilder().social(true).build()) }
+                .filter { this.userOutput.value.isNotNull() }
+                .subscribe { _ -> this.userInput.onNext(this.userOutput.value!!.toBuilder().social(true).build()) }
```
leighdouglas, line 127: same as above, we have historically avoided using `!!`

app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
```diff
@@ -50,91 +51,97 @@ interface PrivacyViewModel {
...
-                .subscribe { _ -> this.showConfirmFollowingOptOutPrompt.onNext(null) }
+                .subscribe { _ -> this.showConfirmFollowingOptOutPrompt.onNext(Unit) }
+                .addToDisposable(disposables)
 
             this.optOutOfFollowing
-                .compose<Boolean>(bindToLifecycle<Boolean>())
                 .filter { optOut -> optOut }
-                .subscribe { _ -> this.userInput.onNext(this.userOutput.value.toBuilder().social(false).build()) }
+                .filter { this.userOutput.value.isNotNull() }
+                .subscribe { _ -> this.userInput.onNext(this.userOutput.value!!.toBuilder().social(false).build()) }
```
leighdouglas, line 138: same as above, we have historically avoided using `!!`, tho im open to revisiting this

app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
mtgriego: no, since the operation happens in order it.isNotNull checks it first, then if it is null the filter fails since its an 'and' check, if it is not null, the second check happens to cover the getValue value

fun <T : Any> T?.isNotNull(): Boolean {
    return this != null
}

leighdouglas: Ah, what i mean is that im concerned with the first null check. to me, this is implying that the item could be null before hitting that first check, and my understanding is that any null value that's passed down the pipeline is an issue with rxjava 2

mtgriego: I believe the concept is called short circuiting.  Basically since its an && it will evaluate the operations in order and if any fail along the way it doesnt continue and just marks the result as false

leighdouglas: got it! thank you for explaining 👍🏽

leighdouglas (approved):
🎉

app/src/main/java/com/kickstarter/viewmodels/PrivacyViewModel.kt
mtgriego: > Ah, what i mean is that im concerned with the first null check. to me, this is implying that the item could be null before hitting that first check, and my understanding is that any null value that's passed down the pipeline is an issue with rxjava 2

I believe RX2 doesnt like nulls, but as long as you dont emit them you are okay.  I could be wrong about that, but since in this case we are just grabbing an observable im just covering the very odd case its null so that the subsequent it.getValue does not crash

mtgriego: i can remove them if you like

PR #2126, opened 2024-09-10
MBL-1722: Know the state of a feature flag in the backend for specific user

Arkariang:
# 📲 What

- We need to enroll batches or users into beta features, currently remote config does not allow adding several hundreds of unique identifiers rules to an specific feature flag, not create an audience with unique identifiers. 
- Flipper the feature flagging system used by our backed does allow adding unique userId's to an specific feature flag

Therefor we have for now decided to check both feature flags information (backend and remote config) for Pledged Dashboard Overview feature.

# 🛠 How

- query available in graphQL `me { enabledFeatures }`, it will return a list with the names of the feature flags that the user has access to.
- Every time the currentUser observable service is updated/refreshed on the app we query now and update the user with `me {enabledFeatures}` information.
- `FeatureFlagClient` now can receive an user and tell if the given `FlipperFlagKey` is enabled or not for given user.

|  |  |

# 📋 QA

- Kill the app, and come back again, your current user should still have access to the backend feature flags information. You can double check on datadog the network call for your user, but basically every time the app come active to Foreground you'll see 3 network calls : 
- - to the config endpoint (restApi), 
- - to the self endpoint(restApi),  
- - to graphql -> user privacy query
Screenshot 2024-09-10 at 3 10 33 PM

# Story 📖

[MBL-1722](https://kickstarter.atlassian.net/browse/MBL-1722)

[MBL-1722]: https://kickstarter.atlassian.net/browse/MBL-1722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/libs/FirebaseAnalyticsClient.kt
```diff
@@ -34,9 +34,7 @@ open class FirebaseAnalyticsClient(
 
     override fun sendUserId(user: User) {
         firebaseAnalytics?.let {
-            if (isEnabled()) {
-                firebaseAnalytics.setUserId(user.id().toString())
-            }
+            firebaseAnalytics.setUserProperty("ks_user_id", user.id().toString())
```
Arkariang, line 37: Can still be useful in the future so letting it here but let me know if there is any concern, using unique ID's is not possible currently for Firebase Remote Config, but for groups such as admin users or super backers can potentially be useful in the future

app/src/main/graphql/userprivacy.graphql
```diff
@@ -16,6 +16,7 @@ query UserPrivacy {
     isDeliverable
     isEmailVerified
     chosenCurrency
+    enabledFeatures
```
Arkariang, line 19: List of features the current user is enrolled in

app/src/main/java/com/kickstarter/libs/utils/ApplicationLifecycleUtil.kt
```diff
@@ -123,9 +138,16 @@ class ApplicationLifecycleUtil(private val application: KSApplication) :
         } else {
             if (accessToken.isNotNull() && accessToken.isNotEmpty()) {
                 client.fetchCurrentUser()
+                    .withLatestFrom(apolloClient.userPrivacy()) { user, userPrivacy ->
```
Arkariang, line 141: Every time the apps comes to Foreground the current user gets updated, it was updated only with V1 information, not with anything available on GraphQL, so updating here just the bare minimum

app/src/main/java/com/kickstarter/libs/featureflag/FeatureFlagClient.kt
```diff
@@ -52,6 +57,9 @@ interface FeatureFlagClientType {
      */
     fun getString(FlagKey: FlagKey): String
 }
+enum class FlipperFlagKey(val key: String) {
+    FLIPPER_PLEDGED_PROJECTS_OVERVIEW("pledge_projects_overview_2024")
```
Arkariang, line 61: "pledge_projects_overview_2024" is the name of the feature flag on flipper

app/src/main/java/com/kickstarter/viewmodels/LoggedInViewHolderViewModel.kt
```diff
@@ -112,8 +117,10 @@ interface LoggedInViewHolderViewModel {
                 .addToDisposable(disposables)
 
             Observable.just(
-                environment.featureFlagClient()
-                    ?.getBoolean(FlagKey.ANDROID_PLEDGED_PROJECTS_OVERVIEW) ?: false
+                featureFlagClient.getBoolean(FlagKey.ANDROID_PLEDGED_PROJECTS_OVERVIEW) &&
```
Arkariang, line 120: Now PPO is tied to both Feature Flags, the remote config one (will restrict app versions), and the backend one (will add/remove unique userID's )

PR #2131, opened 2024-09-18
MBL-1673 [Android 15] Update Views to be compatible with edge to edge enforcement

jlplks:
# 📲 What
Modify all the screens to be compatible with Android edge to edge.

# 🤔 Why

Android 15 introduced edge to edge view enforcement for apps targeting the version (sdk 35).  We need to update some views to look normal in this new context.  

# 🛠 How

For the screens that use composable fixed with` Modifier.systemBarsPadding(),`
For the others screens a utility was created `WindowInsetsUtil`

# Story 📖

[[Android 15] Update Views to be compatible with edge to edge enforcement](https://kickstarter.atlassian.net/browse/MBL-1673)


app/build.gradle
```diff
@@ -52,10 +52,10 @@ if (file('signing.gradle').exists()) {
 
 android {
     defaultConfig {
-        compileSdk 34
+        compileSdkPreview = "VanillaIceCream"
```
Arkariang, line 55: Could you use API level rather than the name? aka: 35

jlplks: No according to [this docs](https://developer.android.com/codelabs/edge-to-edge#2)
Since the stable release hasn't being published yet I use the `compileSdkPreview`. If not, edge to edge is not enable for default
Screenshot 2024-09-18 at 6 44 08 p m

mtgriego:
Screenshot_20241003_125838
Screenshot_20241003_125747
Screenshot_20241003_125703
Screenshot_20241003_125626

Tested a bunch of the screens i pointed out in the ticket as having issues with edge to edge on the branch and I still see the issues.  Many of the top bars cant be interacted with now as they are behind the notification bar

jlplks:
In what device are you testing, @matthewfamularo ? I test on an emulator targeting Android 15 and it works
Screenshot_20241003_132001
Screenshot_20241003_132011
Screenshot_20241003_132020


Screenshot 2024-10-03 at 1 19 11 p m
Screenshot_20241003_131940

Arkariang (approved):

PR #2148, opened 2024-10-08
MBL-1776: When getting a new Push Token from Firebase, sent it the backend endpoint

Arkariang:
# 📲 What

Affecting only KS Push notifications. Braze push notifications are not affected by this specific addition
There are a few scenarios were the push notification token 

Screenshot 2024-10-08 at 2 04 48 PM

The current Pr change will affect the last two scenarios described, for the 3 first ones `RegisterTokenWorker` will call `push_tokens` endpoint when logging in the user.

# 🤔 Why
- Some users complaining about not receiving push notifications, workaround is to clear data/log out/log in again but is a clear indication that the token was likely invalid and never registered a new one. 

# 🛠 How
- the endpoint `/v1/users/self/push_tokens` that we call using the retrofit V1 client  method `registerPushToken` is now called as well when the Firebase Service triggers `onNewToken`
- the endpoint will handle registering the new token 

|  |  |

# 📋 QA

- We cannot really QA the last two scenarios as cannot force trigger `onNewToken`, but if you need more information take a look [here](https://firebase.blog/posts/2023/04/managing-cloud-messaging-tokens/)

# Story 📖

[MBL-1776](https://kickstarter.atlassian.net/browse/MBL-1776)


app/src/main/java/com/kickstarter/services/firebase/MessageService.kt
```diff
@@ -0,0 +1,122 @@
...
+    override fun onNewToken(s: String) {
+        super.onNewToken(s)
+        val refreshHandler = object : RefreshPushToken {}
+        refreshHandler.invoke(
+            apiClient = apiClient,
+            newToken = s,
+            gson = gson,
+            successCallback = { message -> FirebaseCrashlytics.getInstance().log(message) },
+            errorCallback = { errorMessage -> FirebaseCrashlytics.getInstance().recordException(Exception(errorMessage)) }
+        )
```
Arkariang, line 66: From line 59 to 65 is the newly added code, everything else if just Java-Kotlin migration

app/src/main/java/com/kickstarter/services/firebase/MessageService.kt
```diff
@@ -0,0 +1,122 @@
...
+    fun invoke(apiClient: ApiClientTypeV2, newToken: String, gson: Gson, successCallback: (String) -> Unit, errorCallback: (String) -> Unit) {
+        try {
+            val response = apiClient.registerPushToken(newToken).blockingSingle()
+            val message = "$OK_MESSAGE $response"
+            successCallback.invoke(message)
+        } catch (exception: ApiException) {
+            val errorMessage = "$KO_MESSAGE ${
+            exception.errorEnvelope().errorMessage()
+            }"
+            if (exception.errorEnvelope().httpCode() != 401) // else OAuth token not authorized
```
Arkariang, line 37: `registerPushToken` requires a valid OAuth token if 401 triggered means the user is either not logged in, nor valid any longer. We want to register any error only with other http type of errors

leighdouglas (approved):
LGTM! this is great! 🎉

PR #2163, opened 2024-11-08
MBL-1850: Restricted shipping to EU wrong shipping cost being charged to the backer

Arkariang:
# 📲 What

When a reward is configured with cost for sub-regions like EU, this subregion is currently not recognized on the android app.

- - in a project offering rewards shipping restricted to the EU, when selecting on shipping selector a EU country, none of the restricted rewards shipping ONLY to EU will be visible to the backer.

- - If a reward was configured as globally shipping with a price ($10), and also to the EU with a different price ($5), if the backer on shipping selector selected a EU country, the price charged will be the price specified for global shipping($10) (not the correct one which will be the EU ($5))

# 🤔 Why
- Before we were retrieving the field `shippingRules` which is way better performance field to query, but does not allows us to identify if a country belongs to the EU subregion.
- Now we query `simpleShippingRulesExpanded` this field is a flattened representation (and better performance) from the field `shippingRulesExpanded` that shows ALL locations available to ship to for the reward.

# 🛠 How

- queryng now for `simpleShippingRulesExpanded` instead of `shippingRules` for rewards on the project query.
- new transformer added `simpleShippingRuleTransformer`,the given `SimpleShippingRulesExpanded` model will be transformed into the data model `ShippingRule` defined within the app, avoiding any extra side effect outside the networking layer.
- Network call on `GetShippingRulesUseCase` was used to populate the shipping selector is no longer needed. All shipping locations have been collected on the project query when loading project page. 

# 👀 See
- Filtering has been fixed
- Charging correct amounts has been fixed

https://github.com/user-attachments/assets/d453c163-48eb-4e84-b077-828bf1ef1030

|  |  |

# 📋 QA

- Go to this project in staging -> https://staging.kickstarter.com/projects/isabel-martin/burning-man-fort-watch
- - Select on the shipping selector any of the EU countries, you will see more rewards as the project has several rewards with restricted shipping to just EU.
- - Make sure the amount loaded on the checkout screen is correct
-- You should be able to pledge successfully with the right amount. (More work regarding this piece may come later).

# Story 📖

[MBL-1850](https://kickstarter.atlassian.net/browse/MBL-1850)

[MBL-1850]: https://kickstarter.atlassian.net/browse/MBL-1850?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
+++ b/app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
@@ -130,6 +130,7 @@ fun environmentalCommitmentTransformer(envCommit: fragment.EnvironmentalCommitme
 fun rewardTransformer(
     rewardGr: fragment.Reward,
     shippingRulesExpanded: List<fragment.ShippingRule> = emptyList(),
+    simpleShippingRules: List<ShippingRule> = emptyList(),
     allowedAddons: Boolean = false,
     rewardItems: List<RewardsItem> = emptyList(),
     addOnItems: List<RewardsItem> = emptyList()
@@ -161,15 +162,10 @@ fun rewardTransformer(
     val limit = if (isAddOn) chooseLimit(rewardGr.limit(), rewardGr.limitPerBacker())
     else rewardGr.limit()
 
-    val shippingRules = if (shippingRulesExpanded.isNotEmpty()) {
-        shippingRulesExpanded.map {
-            shippingRuleTransformer(it)
-        }
-    } else {
-        rewardGr.shippingRules().map {
-            shippingRuleTransformer(it.fragments().shippingRule())
-        }
-    }
+    val shippingRules =
+        shippingRulesExpanded.takeIf { it.isNotEmpty() }?.map { shippingRuleTransformer(it) }
+            ?: simpleShippingRules.takeIf { it.isNotEmpty() }?.map { it }
+            ?: emptyList()
 
     val localReceiptLocation = locationTransformer(rewardGr.localReceiptLocation()?.fragments()?.location())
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
```diff
@@ -200,6 +198,30 @@ fun rewardTransformer(
...
+    val country = simpleShippingRules.country() ?: ""
+    val displayName = simpleShippingRules.locationName()
+
+    val location = Location.builder()
+        .id(id)
+        .country(country)
+        .displayableName(displayName)
+        .name(displayName)
+        .build()
+    val cost = simpleShippingRules.cost()?.toDoubleOrNull() ?: 0.0
```
Arkariang, line 212: `toDoubleOrNull` if the string is not a double will return null, and default value will apply 0.0
`toDoubleOrNull` precision: rounds up from the 15th digit
Take a look at the test `test simpleShippingRuleTransformer provides appropriate shippingRule` where all those use cases are tested

app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
```diff
@@ -130,6 +130,7 @@ fun environmentalCommitmentTransformer(envCommit: fragment.EnvironmentalCommitme
 fun rewardTransformer(
     rewardGr: fragment.Reward,
     shippingRulesExpanded: List<fragment.ShippingRule> = emptyList(),
+    simpleShippingRules: List<FullProject.SimpleShippingRulesExpanded> = emptyList(),
```
Arkariang, line 133: `simpleShippingRules` -> when fetching rewards.
`shippingRulesExpanded` -> existed before, used when fetching addOns.

app/src/test/java/com/kickstarter/services/GraphQLTransformersTest.kt
```diff
@@ -176,4 +184,136 @@ class GraphQLTransformersTest : KSRobolectricTestCase() {
         assertTrue(post.publishedAt() == date)
         assertTrue(post.updatedAt() == date)
     }
+
+    @Test
+    fun `test simpleShippingRuleTransformer provides appropriate shippingRule`() {
+        val canadaSimpleSR = FullProject.SimpleShippingRulesExpanded(
+            "SimpleShippingRule",
+            "17.34562379823645234875620384756203847234",
```
Arkariang, line 192: This string value will be converted to a double by the `simpleShippingRuleTransformer` on line 220. 
The obtained double value can be checked on line 227

leighdouglas (approved):
lgtm! 🎉

PR #2164, opened 2024-11-08
MLB-1796 COLLECTION PLAN COMPONENT

jlplks:
# 📲 What

Create collection plan component for PLOT

# 🛠 How

Created component with compose library

# 👀 See

Screenshot 2024-11-11 at 11 49 25 a m

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/views/compose/checkout/CollectionPlan.kt
+++ b/app/src/main/java/com/kickstarter/ui/views/compose/checkout/CollectionPlan.kt
@@ -10,119 +10,189 @@ import androidx.compose.runtime.*
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
-import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
+import com.kickstarter.R
+import com.kickstarter.ui.compose.designsystem.KSTheme
+import com.kickstarter.ui.compose.designsystem.KSTheme.colors
+import com.kickstarter.ui.compose.designsystem.KSTheme.dimensions
+import com.kickstarter.ui.compose.designsystem.KSTheme.typography
+import androidx.compose.ui.res.stringResource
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/ui/views/compose/checkout/CollectionPlan.kt
```diff
@@ -0,0 +1,243 @@
...
+}
+
+@Composable
+fun CollectionPlan(isEligible: Boolean, initialSelectedOption: String = "Pledge in full") {
+    var selectedOption by remember { mutableStateOf(initialSelectedOption) }
+
+    Column(modifier = Modifier.fillMaxWidth()) {
+        PledgeOption(
+            optionText = stringResource(id = R.string.fpo_pledge_in_full),
+            selected = selectedOption == "Pledge in full",
```
Arkariang, line 77: @jlplks question here what will happen with this `selectedOption == "Pledge in full"` if the user has as selected language Spanish or any other language?
We could probably define some kind of Enum class to differentiate when the selected option is `Pledge in full` vs `Pledge over time`.

app/src/main/java/com/kickstarter/ui/views/compose/checkout/CollectionPlan.kt
```diff
@@ -0,0 +1,243 @@
...
+
+@Preview(showBackground = false, name = "Eligible - Pledge Over Time Selected")
+@Composable
+fun PreviewPledgeOverTimeSelected() {
+    KSTheme {
+        CollectionPlan(isEligible = true, initialSelectedOption = "Pledge Over Time")
+    }
+}
+
+@Preview(showBackground = false, name = "Not Eligible")
```
Arkariang, line 52: @jlplks nitpick here can we add the same previews as well for the darktheme?

jlplks: Screenshot 2024-11-12 at 3 39 48 p m

app/src/main/java/com/kickstarter/ui/views/compose/checkout/CollectionPlan.kt
jlplks: Enum added to manage the Collection plan options

Arkariang (approved):

PR #2166, opened 2024-11-13
MLB -1798: Added plot selected badge to PledgeItemizedDetails

jlplks:
# 📲 What

Created PLOT badge on PledgeItemizedDetails
# 🤔 Why

To indicates when PLOT is selected.

# 🛠 How

added plotSelected prop, created component, added tests and added previews

# 👀 See

Screenshot 2024-11-12 at 4 40 20 p m


app/src/main/java/com/kickstarter/ui/views/compose/checkout/PledgeItemizedDetails.kt
```diff
@@ -101,7 +128,8 @@ fun ItemizedRewardListContainer(
     totalBonusSupport: String,
     deliveryDateString: String = "",
     rewardsHaveShippables: Boolean = false,
-    disclaimerText: String = ""
+    disclaimerText: String = "",
+    plotSelected: Boolean
```
leighdouglas, line 132: We can pass in a default value of `false` here for `plotSelected`, and then it will only show when we pass in a `true` value. that way we aren't required to pass in a value for this field everywhere we call this function unless it's PLOT related

app/src/test/java/com/kickstarter/ui/activities/compose/PledgeItemizedDetailsTest.kt
```diff
@@ -45,6 +45,9 @@ class PledgeItemizedDetailsTest : KSRobolectricTestCase() {
     private val currencyConversion =
         composeTestRule.onNodeWithTag(PledgeItemizedDetailsTestTag.CURRENCY_CONVERSION.name)
 
+    private val plotSelectedBadge =
+        composeTestRule.onNodeWithTag(PledgeItemizedDetailsTestTag.PLOT_SELECTED_BADGE.name)
```
leighdouglas, line 49: 😃

app/src/test/java/com/kickstarter/ui/activities/compose/PledgeItemizedDetailsTest.kt
```diff
@@ -256,11 +266,39 @@ class PledgeItemizedDetailsTest : KSRobolectricTestCase() {
...
+                    plotSelected = false
                 )
             }
         }
         deliveryDate.assertIsDisplayed()
         deliveryDate.assertTextEquals("April 10")
     }
+
+    @Test
+    fun `test plot selected, when plot selected should show plot selected badge`() {
```
leighdouglas, line 279: 😃

leighdouglas (dismissed):
LGTM! Left a small optional comment that might clean things up a little bit, but otherwise this looks great!

app/src/main/java/com/kickstarter/ui/views/compose/checkout/PledgeItemizedDetails.kt
jlplks: Change added 😄

Arkariang (approved):

leighdouglas (approved):
lgtm!

PR #2173, opened 2024-11-19
Imartin/separate rewards from project

Arkariang:
# 📲 What

- Project query obtains now very minimum information about rewards,(ideally in the future will not query anything around rewards, but for now some details to pre-load quickly the rewards carousel is required).
- New query that will fetch the rewards + simple shipping rules + reward Items has being created, this new query is called within `RewardsSelectionViewModel`.
- Crash fixed related to changing the selected location for a pledge.

# 🤔 Why

Project query has modified to obtain all simple shipping rules, for all rewards, (Globally shipping rewards then will hold a collection of 246 shipping rules each). Project (as any other data model within the app) can be parzelized to be serialized/deserialized from Intents, but unfortunately we have hit again the `transactionTooLarge` exception due to Project being too big (we need to find other ways to persist information within the app, passing the hole object from screen to screen via intent to avoid extra network calls has clear limitations).

# 👀 See

|  |  |

# 📋 QA

# Story 📖

[MBL-1877](https://kickstarter.atlassian.net/browse/MBL-1877)
[MBL-1893](https://kickstarter.atlassian.net/browse/MBL-1893)


app/src/main/java/com/kickstarter/libs/utils/extensions/PledgeDataExt.kt
```diff
@@ -28,7 +28,7 @@ fun PledgeData.shippingCostIfShipping(): Double {
     var addOnsShippingCost = 0.0
     this.addOns()?.map {
         if (RewardUtils.shipsWorldwide(it) || RewardUtils.shipsToRestrictedLocations(it)) {
-            addOnsShippingCost += (it.shippingRules()?.first()?.cost() ?: 0.0) * (it.quantity() ?: 0)
+            addOnsShippingCost += (it.shippingRules()?.firstOrNull()?.cost() ?: 0.0) * (it.quantity() ?: 0)
```
Arkariang, line 31: The crash was due to using the reward and addOns from the backing object, where shipping locations information is not available

app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
```diff
@@ -229,10 +228,14 @@ class AddOnsViewModel(val environment: Environment, bundle: Bundle? = null) : Vi
...
-        // Take the backed AddOns, update with the backed AddOn information which will contain the backed quantity
-        backedAddOns.map {
-            holder[it.id()] = it
-            currentSelection[it.id()] = it.quantity() ?: 0
+        // Take the backed AddOns, update with matching addOn ID with the quantity information
+        backedAddOns.map { backedAddOn ->
+            val aux = holder[backedAddOn.id()]
+            if (aux != null) {
+                val updated = aux.toBuilder().quantity(backedAddOn.quantity()).build()
+                holder[backedAddOn.id()] = updated
```
Arkariang, line 236: before we were using the information from the baked object, now we update the "quanity" alone to not loose the "shipping rules" field

app/src/main/java/com/kickstarter/viewmodels/projectpage/CrowdfundCheckoutViewModel.kt
```diff
@@ -158,12 +157,11 @@ class CrowdfundCheckoutViewModel(val environment: Environment, bundle: Bundle? =
                 sharedPreferences
             )
 
-            when (flowContext) {
-                PledgeFlowContext.NEW_PLEDGE,
-                PledgeFlowContext.CHANGE_REWARD -> getPledgeInfoFrom(pData)
-                PledgeFlowContext.MANAGE_REWARD,
```
Arkariang, line 164: "Manage_Reward" is used for both updating selected reward and changing reward, it describes the fact of having to present to the user RewardsCarousel not granular enough. 
Exchanged to use `PledgeReason` instead of `PledgeFlowContext` as it's uniquely identified.

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
```diff
@@ -714,6 +717,45 @@ class KSApolloClientV2(val service: ApolloClient, val gson: Gson) : ApolloClient
...
+                                    rewardGr = it.fragments().reward(),
+                                    allowedAddons = it.allowedAddons().pageInfo().startCursor()?.isNotEmpty() ?: false,
+                                    rewardItems = complexRewardItemsTransformer(it.items()?.fragments()?.rewardItems()),
+                                    simpleShippingRules = it.simpleShippingRulesExpanded()
+                                )
+                            } ?: emptyList()
+                            // - API does not provide the Reward no reward, we need to add it first
+                            val minPledge = data.project()?.minPledge()?.toDouble() ?: 1.0
+                            val modifiedRewards = rwList.toMutableList()
+                            modifiedRewards.add(0, RewardFactory.noReward().toBuilder().minimum(minPledge).build())
```
Arkariang, line 749: Adding reward no reward to the list of rewards. It is not provided by the API

app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
```diff
@@ -150,7 +150,6 @@ class AddOnsViewModel(val environment: Environment, bundle: Bundle? = null) : Vi
                         bonusAmount = b.amount()
                     } else {
                         backedAddOns = b.addOns() ?: emptyList()
-                        currentUserReward = b.reward() ?: currentUserReward
```
Arkariang, line 153: this was causing the issue mentioned [here](https://github.com/kickstarter/android-oss/pull/2173/files#r1848962215)

leighdouglas (approved):
LGTM!!!!

PR #2178, opened 2024-11-27
MBL-1858: Payment Schedule component

jlplks:
# 📲 What
Crea te a Payment Schedule component on compose with tests

# 🤔 Why

To show the payment schedule on 'Your pledge section'

# 🛠 How

Created a compose class

# 👀 See

Screenshot_20241126_224813
Screenshot_20241126_224823


app/src/main/java/com/kickstarter/ui/views/compose/checkout/PaymentSchedule.kt
```diff
@@ -0,0 +1,232 @@
...
+            onExpandChange = { isExpanded = it }
+        )
+    }
+}
+
+@Composable
+fun PaymentSchedule(
+    isExpanded: Boolean = false,
+    onExpandChange: (Boolean) -> Unit = {},
+    // TODO: Add payment schedule data model when available
```
Arkariang, line 93: fyi, @jlplks we do create our own data models for the app. Once the API layer is ready (either GraphQL or REST) once the network call is made we will make the transformation required within the networking layer, take a look for example at  `KSApolloclient -> override fun getProject(slug: String): Observable<Project>`. The project returned is our own project -> `Project.kt`

leighdouglas (approved):
lgtm!

app/src/main/java/com/kickstarter/ui/views/compose/checkout/PaymentSchedule.kt
Arkariang: another fyi, if you want to know how the PaymentIncrement data model is gona look like on the API, take a look here -> https://kickstarter.atlassian.net/wiki/spaces/PAYM/pages/3210641443/PLOT+Mobile+API+Source+of+Truth?focusedCommentId=3212476418

jlplks: Created data model for it

Arkariang (approved):

PR #2180, opened 2024-11-28
Remove "Update Pledge" option on Manage Your Pledge menu

jlplks:
# 📲 What

Remove "Update Pledge" option on Manage Your Pledge menu

# 🤔 Why

If plot is selected we should remove all the options that interact with UpdateBacking mutation

# 🛠 How

Created another xlm menu and added a validation for show it or not


app/src/main/java/com/kickstarter/viewmodels/projectpage/ProjectPageViewModel.kt
```diff
@@ -1057,9 +1057,13 @@ interface ProjectPageViewModel {
         private fun managePledgeMenu(projectAndFragmentStackCount: Pair<Project, Int>): Int {
             val project = projectAndFragmentStackCount.first
             val count = projectAndFragmentStackCount.second
+            // Check the feature flag
+            // TODO: Add another validation when API is ready to add the value that PLOT is selected on the project
+            val isPledgeOverTimeEnabled = featureFlagClient.getBoolean(FlagKey.ANDROID_PLEDGE_OVER_TIME)
```
Arkariang, line 1062: @jlplks you need to check as well on project, the new field to be added on graphQL api `project.isPledgeOverTimeAllowed` = true.  Add in the todo which Jira ticket will be worked in at least if you wanna tackle this piece of the epic now.

app/src/main/res/menu/manage_pledge_plot_selected.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:app="http://schemas.android.com/apk/res-auto"
```
Arkariang, line 2: @jlplks there is a menu already, I much likely prefer to identify where that menu is, and hide the options not available rather than creating a new one

app/src/main/java/com/kickstarter/viewmodels/projectpage/ProjectPageViewModel.kt
jlplks: Adding it to Project data model

app/src/main/res/menu/manage_pledge_plot_selected.xml
jlplks: the way the menu options are managed inside the class is rendering differents xmls depending on the conditions, that's the reason for managePledgeMenu. That's why I opted for that way, do you prefer that I did it in another way?

jlplks: Added incremental to backing and added a test

Arkariang (dismissed):
👍 looks, good, this code can be merged, but wondering what tickets on the Jira epic will contain the connexion to the API where those two new fields (the one in Project, and the one in Backing ) are queried. Both fields are currently available on the API as shown on the screenshot comment

jlplks:
> 👍 looks, good, this code can be merged, but wondering what tickets on the Jira epic will contain the connexion to the API where those two new fields (the one in Project, and the one in Backing ) are queried. Both fields are currently available on the API as shown on the screenshot comment

I have 3 task for the API, I created them at the start of the PLOT I'll do it there, I will modify the task description to add that changes

leighdouglas (approved):

PR #2185, opened 2024-12-05
MBL-1918 API | Expose 'incremental' field on the `CreateBacking` mutation

jlplks:
# 📲 What

CreateBacking has a new field called incremental on it (available now); when true it is a PLOT backing. No plans to update UpdateBacking as of now, if a backer wants to change to PLOT for v1 they will have to cancel and re-back

This ticket just updates our GraphAPI. Implementation will be in a separate ticket
# 🤔 Why

To be able to pledge a project with PLOT
# 🛠 How

Added 'incremental' field to CreateBacking mutations

### DIFF
--- a/app/src/main/java/com/kickstarter/services/mutations/CreateBackingData.kt
+++ b/app/src/main/java/com/kickstarter/services/mutations/CreateBackingData.kt
@@ -4,6 +4,4 @@ import com.kickstarter.libs.RefTag
 import com.kickstarter.models.Project
 import com.kickstarter.models.Reward
 
-data class CreateBackingData(val project: Project, val amount: String, val paymentSourceId: String? = null, val setupIntentClientSecret: String? = null, val locationId: String?, val reward: Reward? = null, val rewardsIds: List<Reward>? = null, val refTag: RefTag?, val stripeCardId: String? = null, val incremental: Boolean = false) {
-
-}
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/services/mutations/CreateBackingData.kt
```diff
@@ -4,4 +4,6 @@ import com.kickstarter.libs.RefTag
 import com.kickstarter.models.Project
 import com.kickstarter.models.Reward
 
-data class CreateBackingData(val project: Project, val amount: String, val paymentSourceId: String? = null, val setupIntentClientSecret: String? = null, val locationId: String?, val reward: Reward? = null, val rewardsIds: List<Reward>? = null, val refTag: RefTag?, val stripeCardId: String? = null)
+data class CreateBackingData(val project: Project, val amount: String, val paymentSourceId: String? = null, val setupIntentClientSecret: String? = null, val locationId: String?, val reward: Reward? = null, val rewardsIds: List<Reward>? = null, val refTag: RefTag?, val stripeCardId: String? = null, val applePay: Boolean = false, val incremental: Boolean = false, val pledgeFlowContext: PledgeFlowContext? = null) {
+    data class PledgeFlowContext(val isFromBackingFragment: Boolean, val isFromManagePledge: Boolean)
```
leighdouglas, line 8: we only need to add the incremental field to this data class, and we dont need the PledgeFlowContextData class here either

jlplks: done

leighdouglas (approved):
lgtm! great job 🎉

PR #2192, opened 2025-01-08
MBL-1952 Add dynamic root padding to Message Creator screen

tonyteate:
# 📲 What

Fix hidden `Send` button in `MessageCreatorActivity`.

# 🤔 Why

When attempting to message a Creator (when not backing their Project), the keyboard hides the Send button. Users must dismiss the keyboard to send the message. This is a regression caused by Android 15 forced edge-to-edge. 

Screenshot_20250108_163309

# 🛠 How

Update layout root padding based on IME insets.


jlplks (approved):
LGTM

Arkariang (approved):

PR #2195, opened 2025-01-13
MBL-1916: Display payment plan in Manage Pledge Flow if available

jlplks:
# 📲 What

Changes on the  UI of **Manage Your Pledge Screen** to show the PLOT selection on the pledge changin the banner of the text an adding the Payment Schedule component.

# 🤔 Why

For the PLOT UI

# 🛠 How

Added the PaymentIncrement value to the backin object, modify GraphQLTransformer, changing the UI of the Manage your pledge screen to render the PaymentSchedule component, modifying the State enum class on PaymentIncrements to accept raw values in lowercase (cause that how we recibe it from the BE)

# 👀 See

Screenshot_20250113_085602

# 📋 QA

If you don't have a PLOT pledged project do it and then check it on the Manage Your Pledge Screen. The banner should say something about plot and the PaymentSchedule component should be rendered correctly
# Story 📖

[MBL-1916: API | Display payment plan in Manage Pledge Flow if available](https://kickstarter.atlassian.net/browse/MBL-1916)


app/src/main/java/com/kickstarter/ui/fragments/BackingFragment.kt
```diff
@@ -235,6 +241,20 @@ class BackingFragment : Fragment() {
 
             val boldPortionLength = text.toString().split(".").first().length
             setBoldSpanOnTextView(boldPortionLength, this, resources.getColor(R.color.kds_support_400, null))
+
+            binding?.paymentScheduleComposeView?.setContent {
```
leighdouglas, line 245: My inkling here is that this will appear on all manage pledge screens, regardless of if they are a plot pledge. We should probably gate this with the condition that it only shows if pledge is PLOT.

Arkariang: Agre with @leighdouglas

app/src/main/java/com/kickstarter/ui/views/compose/checkout/PaymentSchedule.kt
```diff
@@ -134,60 +133,53 @@ fun PaymentSchedule(
...
+                clickCallback = { onDisclaimerClicked.invoke(DisclaimerItems.TERMS) }
+            )
         }
     }
 }
 
 @Composable
 fun PaymentRow(paymentIncrement: PaymentIncrement) {
-    val formattedAmount = String.format(Locale.US, "%.2f", paymentIncrement.amount.amount.parseToDouble())
+    val formattedAmount =
```
Arkariang, line 181: For formatting ammounts I will suggest to taking a look at `KSCurrency`, we want currency string appropriate to the user's locale

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
```diff
@@ -249,6 +254,12 @@ interface BackingFragmentViewModel {
             }
                 .map { it.second.userIsCreator(it.first.getValue()) }
 
+            backing
```
Arkariang, line 257: For the bussines logic, I will check if the project is enabled or not for plot

app/src/main/java/com/kickstarter/ui/views/compose/checkout/PaymentSchedule.kt
leighdouglas: The backend is going to provide the formatted amounts for us, we are just waiting for this piece to be finished. We actually don't show these amounts in the users currency, but in the project currency.

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
```diff
@@ -249,6 +259,18 @@ interface BackingFragmentViewModel {
...
+                .map { it.paymentIncrements ?: emptyList() }
```
leighdouglas, line 263: i think we should check if payment increments is not null, or empty here, not just null.

leighdouglas (approved):

PR #2199, opened 2025-01-14
MBL-1989: Only call build payment plan query on PLOT enabled projects

leighdouglas:
# 📲 What

`BuildPaymentPlan` query should only be called on projects where plot is enabled
Update amounts reflecting backend changes

# 🤔 Why

An error is occuring because of changes being made to the backend, but they are effecting all of checkout because we are calling this query regardless of if the project is opted into PLOT or not. We should isolate this errored to PLOT projects only, so it doesn't effect all of checkout. Also, it's an unnecessary network call for most projects. 

# 🛠 How

Check project response to make sure PLOT is enabled, then we call the build payment plan query. 

# 📋 QA

Make sure regular checkout flows are not effected. Pledge to a project, make sure all of the numbers and amounts are correct and load properly. Also monitor the network to make sure build payment plan query is not being called on regular checkout. 

Then test PLOT projects. this will error until we update the graph schema, but make sure the query is at least being called here. 

note that amounts in the payment schedule will be fixed in a subsequent ticket. they are currently not formatted correctly.

# Story 📖

[MBL-1989: Only call build payment plan query on PLOT enabled projects](https://kickstarter.atlassian.net/browse/MBL-1989)

[MBL-1795: New amounts in backend response, plug into UI in payment schedule](https://kickstarter.atlassian.net/browse/MBL-1980)


jlplks (approved):
LGTM!

Tested on emulator, saw past pledged PLOT projects and pledge a new PLOT project and everything looks right!

app/src/main/java/com/kickstarter/libs/utils/ProjectViewUtils.kt
```diff
@@ -148,4 +148,41 @@ object ProjectViewUtils {
...
         return styledCurrency.trim()
     }
+
+    /**
+     * Returns a CharSequence representing a value in a project's currency based on a user's locale.
+     * The precision is shrunken and centered if the number is not whole.
+     * The currency symbol is shrunken and centered.
+     * Special case: US people looking at US currency just get the currency symbol.
+     */
+    fun styleCurrency(
```
Arkariang, line 158: For compose UI the styling of the currency will happen on this ticket [https://kickstarter.atlassian.net/browse/MBL-1992](url). And using the new fields available on the API that already provide the currency and symbol. So for now ignore this new method here.

Arkariang (commented):
lgtm

PR #2200, opened 2025-01-14
MBL-1991: Pledge Summary | Only show PLOT blurb in manage pledge while project is live

leighdouglas:
# 📲 What

PLOT disclaimer text on manage pledge should only display while a project is live. Otherwise, it should say "We have collected your pledge for this project"

# 🛠 How

If a user has backed a project with PLOT, the backing state remains in "PLEDGED". we need to check that the project is also still live to show this copy. 

# 👀 See

This is the copy that should be shown, only for live projects: 

Screenshot 2025-01-14 at 18 32 37

This is the copy for plot pledged projects that have ended: 
Screenshot 2025-01-14 at 18 33 49

# 📋 QA

Not sure how we will test that the correct copy shows upon the project ending, but 
1. navigate to a project pledged with PLOT that has ended
2. make sure the copy says "we have collected your pledge" 

Also navigate to a plot pledged project that is still live, and observe that the copy is the plot disclaimer string. 

# Story 📖
[MBL-1991: Pledge Summary | Only show PLOT blurb in manage pledge while project is live](https://kickstarter.atlassian.net/browse/MBL-1991)


jlplks (approved):
LGTM!

Arkariang (approved):

PR #2209, opened 2025-01-23
MBL-1971: Infinite loading on rewards carousel for late pledges projects fixed

Arkariang:
# 📲 What

- At times we found an infinite loading indicator for the rewards carousel, for projects with late pledges active

# 🤔 Why

- User cannot move forward on the checkout process for a late pledge project.

# 🛠 How
- Some context is required first. Late Pledge checkout navigation actually happens within an HorizontalPager which UI is initialized in `ProjectPageActivity` (take a look at [ProjectPledgeButtonAndFragmentContainer](https://github.com/kickstarter/android-oss/blob/31d4d5f367b5edcf0eae2b09766ac28e8dcc5762/app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/ProjectPledgeButtonAndFragmentContainer.kt#L121) )
- Before this update `LatePledgeCheckoutViewModel` had logic within an `init` method, that got triggered when the user first load the ProjectPageActivity due to previous point. What I suspect is the issue here, If the coroutine in which the flow is collected is canceled before the flow completes, the onCompletion block might not be executed. This can happen if the ViewModel is destroyed or if the user navigates away from the screen before the flow has finished (Another project is loaded for example).
- The logic that was living in the initialization block of the ViewModel has been moved to a new method `loadUserInfo()`, executed once the user navigates into the Checkout Screen.
- The calls to `UserPrivacy` and `storedCards` are now sequential and do not update states until both finished.

# 👀 See
| Before 🐛 | 

https://github.com/user-attachments/assets/bba12470-b534-40fc-90e1-77a25270f57e

| After 🦋 |

https://github.com/user-attachments/assets/7afb19df-1af0-4c4f-a128-d8b6fae019a9

| --- | --- |
|  |  |

# 📋 QA
The best way I found to reproduce this bug, is to load several projects (or the same one) from a deeplink as show in the Before video recording.
- Load as many late pledge projects as you can (from search, or discovery or a deeplink), should not run into the infinite loading indicator any longer.

# Story 📖

[MBL-1971](https://kickstarter.atlassian.net/browse/MBL-1971)

[MBL-1971]: https://kickstarter.atlassian.net/browse/MBL-1971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ycheng-kickstarter (dismissed):
LGTM

tonyteate (approved):

PR #2215, opened 2025-01-29
MBL-2052 PLOT projects not working after scheme change on BE

jlplks:
# 📲 What

After a change on BE they remove the prop `projectIsPledgeOverTimeAllowed` from PaymentPlan.
We have to remove it from the project too.

# 🤔 Why

Can't pledge PLOT projects with PLOT becuase the component doesn't render do a validation with that prop.

# 🛠 How

Remove all the uses for that prop and change the only validation where it's used.

# 👀 See

| Before 🐛 |
error_message

| After 🦋 |
[plot_fixed.webm](https://github.com/user-attachments/assets/328efafa-12c8-44fe-bdbf-07aedf85f353)

# 📋 QA

Enter with a valid user to staging an look for [Crossfit Watch Hot Sauce](https://staging.kickstarter.com/projects/1768690592/crossfit-watch-hot-sauce?ref=project_build)

the project is a PLOT project so if you follow the PLOT flow you should see the PLOT component rendering correctly and have no problem to pledge it

# Story 📖

[MBL-2052 PLOT projects not working after scheme change on BE](https://kickstarter.atlassian.net/browse/MBL-2052)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/projectpage/CrowdfundCheckoutViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/projectpage/CrowdfundCheckoutViewModel.kt
@@ -277,7 +277,7 @@ class CrowdfundCheckoutViewModel(val environment: Environment, bundle: Bundle? =
                     emitCurrentState(isLoading = false)
                 }
                 .collectLatest {
-                    showPlotWidget = !it.paymentIncrements.isNullOrEmpty()
+                    showPlotWidget = project.isPledgeOverTimeAllowed() == true
                     plotEligible = it.amountIsPledgeOverTimeEligible
                     paymentIncrements = it.paymentIncrements
                     emitCurrentState(isLoading = false)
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/viewmodels/projectpage/CrowdfundCheckoutViewModel.kt
```diff
@@ -277,7 +277,7 @@ class CrowdfundCheckoutViewModel(val environment: Environment, bundle: Bundle? =
                     emitCurrentState(isLoading = false)
                 }
                 .collectLatest {
-                    showPlotWidget = it.projectIsPledgeOverTimeAllowed
+                    showPlotWidget = !it.paymentIncrements.isNullOrEmpty()
```
leighdouglas, line 280: Won't the payment increments be null or empty if the pledge isn't PLOT eligible? Fundamentally, the widget visibility should be toggled based on whether the project has been opted into PLOT, not on whether the payment increments exist.

leighdouglas:
QAed this and the PLOT widget does not appear at all when the pledge does not meet the PLOT minimum. See my comment above to fix the issue. Thanks for catching this!

app/src/main/java/com/kickstarter/viewmodels/projectpage/CrowdfundCheckoutViewModel.kt
jlplks: Yes, Fixed that on the next commit. Thanks for noticing it

leighdouglas (approved):
lgtm!

PR #2216, opened 2025-01-29
MBL-2054: Manage Exceptions from Rx streams lacking `onError`

tonyteate:
# 📲 What

Rework the `RxJavaPlugins` error handler to report all previously fatal Exceptions to Crashlytics. Report `ApolloHttpException`s with status code `429` in particular with a custom key to simplify querying.

# 🤔 Why

Prevent Exceptions from Rx streams lacking `onError` handlers from becoming fatal:

Any consumer of an Rx stream that does not handle errors creates an opportunity for an unhandled `OnErrorNotImplementedException`. Currently, there are many Rx streams that are chained together across multiple layers of the app, which makes it challenging to pinpoint not only the root cause of unhandled exceptions, but also where to possibly add error handling.

If enabled, these particular kinds of exceptions can be globally reported to a custom error handler defined by `RxJavaPlugins.setErrorHandler()`, which is already in use. Previously, within this error handler, the app attempted to pass every exception that was not an instance of `UndeliverableException` to a default `UncaughtExceptionHandler`. However, that handler does not exist by default on Android, and thus all `OnErrorNotImplementedException`s are fatal.

# 🛠 How

Refactor the `RxJavaPlugins` error handler to instead report all previously fatal Exceptions to Crashlytics. For `ApolloHttpException`s with status code `429` in particular, report them with a custom key to simplify querying.

The test case represents a scenario that is currently possible wherein a `null` is "injected" into one of the app's Rx streams, leading to a fatal exception.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

Use the app normally. Experience less crashes.

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2054
https://kickstarter.atlassian.net/browse/MBL-1977


app/src/main/java/com/kickstarter/KSApplication.java
```diff
@@ -116,15 +118,24 @@ private void setVisitorCookie() {
   }
 
   private void createErrorHandler() {
-    RxJavaPlugins.setErrorHandler(e -> {
-      if (e instanceof UndeliverableException) {
-        Timber.w(e, "Undeliverable Exception");
-        if (e.getMessage() != null) {
-          FirebaseCrashlytics.getInstance().setCustomKey("Undeliverable Exception", e.getMessage());
+    RxJavaPlugins.setErrorHandler(t -> {
```
Arkariang, line 121: 👍

Arkariang (approved):

PR #2219, opened 2025-02-03
MBL-2068 Images on rewards - integrate with backend

ycheng-kickstarter:
# 📲 What

Backend integration for images on rewards

# 🤔 Why

We are adding images on rewards

# 🛠 How

Update graphql fragments, transformers, and data models

# 👀 See

No visual changes

# 📋 QA

Just needs to compile

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2068


app/src/main/graphql/fragments.graphql
```diff
@@ -315,6 +316,10 @@ fragment reward on Reward {
     localReceiptLocation {
         ... location
     }
+    image {
```
Arkariang, line 319: fyi @ycheng-kickstarter if you could make sure, for Images on rewards piece in graphQL, is only used within `getRewardsFromProject` query? Would like to avoid adding unnecessary loads for any of the discovery calls that might potentially use the `full` fragment or we might run into the the `TransactionTooLargeException` exception

Arkariang: also, could we add the ticket number on the PR? makes things easier when searching for specific issues + with the generated build on Firebase Internal Release should be easy to identify to which ticket is associated (now with the Design/Product QA column in jira will come extra handy).

ycheng-kickstarter: Ohh thanks for that context! Ok I moved the image field
```
   image {
        url(width: 1024)
        altText
    }
```
out of the reward fragment and into the FetchProjectRewards query only. Let me know if I understood you correctly! 

I didn't quite understand your comment about the `full` fragment though. The only change I made for the existing project image field was to add an `altText` field. Btw the naming confused me - why is the `full` fragment named that?

jlplks (approved):
LGTM!

app/src/main/graphql/project.graphql
Arkariang, line 1: Question, if I recall correctly there is images for reward Items as well. I assume that piece is not within the project you are currently tackling just asking to double check.

app/src/main/graphql/fragments.graphql
```diff
@@ -210,6 +210,14 @@ fragment rewardItems on RewardItemsConnection {
 fragment full on Project {
     image {
         url(width: 1024)
+        altText
+    }
+}
+
+fragment rewardImage on Reward {
```
Arkariang, line 217: As future enhancement we could pass from the VM to the query some parameter specifying quality for the image, so we could query for different images widths depending on device size for example :)

Arkariang (approved):
lgtm!

app/src/main/graphql/fragments.graphql
Arkariang: it is a good question, full makes reference to the entirety of the fields of the project page, we should not use `full` fragment for search or discovery queries that take projects as an example.
Said so we could use some renaming and review how/when our fragments are used (specially for project)

app/src/main/graphql/project.graphql
ycheng-kickstarter: An older iteration of the design showed images for reward items, but the latest figma does not.

app/src/main/graphql/fragments.graphql
ycheng-kickstarter: @Arkariang do you have context into why we chose width of 1024 for project photos? I used the same width for reward images but I'm wondering if we should reduce it to optimize loading speed.

PR #2223, opened 2025-02-11
Start coroutines migration of `KSApolloClientV2.getStoredCards()`

tonyteate:
Do not merge ❌

# 📲 What

Start coroutines migration of `KSApolloClientV2.getStoredCards()`

- Update PaymentMethodsViewModel + Tests
- Update LatePledgeCheckoutViewModel + Tests

# 🤔 Why

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖

### DIFF
--- a/app/src/main/java/com/kickstarter/mock/services/MockApolloClientV2.kt
+++ b/app/src/main/java/com/kickstarter/mock/services/MockApolloClientV2.kt
@@ -147,6 +147,9 @@ open class MockApolloClientV2 : ApolloClientTypeV2 {
         return io.reactivex.Observable.just("")
     }
 
+    override suspend fun _getStoredCards(): List<StoredCard> =
+        getStoredCards().blockingSingle() // or `blockingFirst()`
+
     override fun getStoredCards(): io.reactivex.Observable<List<StoredCard>> {
         return io.reactivex.Observable.just(Collections.singletonList(StoredCardFactory.discoverCard()))
     }
--- a/app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
+++ b/app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
@@ -4,6 +4,7 @@ import android.util.Pair
 import com.apollographql.apollo3.ApolloClient
 import com.apollographql.apollo3.api.ApolloResponse
 import com.apollographql.apollo3.api.Optional
+import com.apollographql.apollo3.exception.ApolloException
 import com.apollographql.apollo3.rx2.rxFlowable
 import com.apollographql.apollo3.rx2.rxSingle
 import com.google.android.gms.common.util.Base64Utils
@@ -115,6 +116,9 @@ import io.reactivex.Observable
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.schedulers.Schedulers
 import io.reactivex.subjects.PublishSubject
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.single
+import kotlinx.coroutines.rx2.asObservable
 import java.nio.charset.Charset
 
 interface ApolloClientTypeV2 {
@@ -123,6 +127,7 @@ interface ApolloClientTypeV2 {
     fun getProjects(discoveryParams: DiscoveryParams, slug: String?): Observable<DiscoverEnvelope>
     fun createSetupIntent(project: Project? = null): Observable<String>
     fun savePaymentMethod(savePaymentMethodData: SavePaymentMethodData): Observable<StoredCard>
+    suspend fun _getStoredCards(): List<StoredCard>
     fun getStoredCards(): Observable<List<StoredCard>>
     fun deletePaymentSource(paymentSourceId: String): Observable<DeletePaymentSourceMutation.Data>
     fun createFlagging(
@@ -231,6 +236,9 @@ class KSApolloClientV2(val service: ApolloClient, val gson: Gson) : ApolloClient
     override fun cleanDisposables() {
         disposables.clear()
     }
+
+    private fun <T : Any> emitAsObservable(block: suspend () -> T): Observable<T> =
+        flow { emit(block()) }.asObservable()
 
     override fun getProject(project: Project): Observable<Project> {
         return getProject(project.slug() ?: "")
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/kickstarter/mock/services/MockApolloClientV2.kt
```diff
@@ -147,6 +147,9 @@ open class MockApolloClientV2 : ApolloClientTypeV2 {
         return io.reactivex.Observable.just("")
     }
 
+    override suspend fun _getStoredCards(): List<StoredCard> =
+        getStoredCards().blockingSingle() // or `blockingFirst()`
```
Arkariang, line 151: blocking a thread worries me a bit, we need to be sure then to add to proper dispatchers. It is a common scenario to spam several calls within the same VM

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
```diff
@@ -232,6 +237,9 @@ class KSApolloClientV2(val service: ApolloClient, val gson: Gson) : ApolloClient
         disposables.clear()
     }
 
+    private fun <T : Any> emitAsObservable(block: suspend () -> T): Observable<T> =
```
Arkariang, line 240: 💖  , elegant!

app/src/main/java/com/kickstarter/viewmodels/projectpage/LatePledgeCheckoutViewModel.kt
```diff
@@ -139,35 +140,32 @@ class LatePledgeCheckoutViewModel(val environment: Environment) : ViewModel() {
...
+                emitCurrentState(isLoading = true)
 
-        val privacyAndCardsFlow = combine(
-            apolloClient.userPrivacy().asFlow(),
-            apolloClient.getStoredCards().asFlow()
-        ) { userPrivacy, cards ->
-            Pair(userPrivacy, cards)
-        }
+                @OptIn(FlowPreview::class)
+                val user = currentUser.timeout(1.seconds).runCatching { first() }.getOrNull()
```
Arkariang, line 149: Why the timeout?

tonyteate: When there is no logged in user, `environment.currentUserV2()?.loggedInUser()` does not emit, which leaves a Collector suspended forever. 

In tests, this is not as clear when running on `backgroundScope` since the scope cancels at the end of the test. In production, if we want to try to get the user, and fail if there isn't one, we need the timeout. Otherwise, if we want to collect forever intentionally, we should launch a separate coroutine earlier in the business logic. 

This currently appears to not be a bug in production because we check for a logged in user before we call `providePledgeData()`.

Arkariang: alternative path to the timeout, can change from `environment.currentUserV2()?.loggedInUser()` over to`environment.currentUser.observable()` it will operate with a KSOptional<User>

tonyteate: Yes, that's perfect 👍  I tried to stay close to the original in case I missed something. Thanks!

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
```diff
@@ -380,42 +388,38 @@ class KSApolloClientV2(val service: ApolloClient, val gson: Gson) : ApolloClient
...
+                StoredCard.builder()
+                    .expiration(cardData.expirationDate)
+                    .id(cardData.id)
+                    .lastFourDigits(cardData.lastFour)
+                    .type(cardData.type)
+                    .stripeCardId(cardData.stripeCardId)
+                    .build()
+            } ?: listOf()
+        } catch (apolloException: ApolloException) {
+            // parse network errors into specific type
```
Arkariang, line 413: If we wanted to provide some feedback to the user. For example a pretty used pattern now  within the app is displaying a red snackbar on the UI, usually with a generic message "something went wrong if no message provided on the API". It was achieved before with the `doOnError` block.
We could add some kind of callback the consumer of `_getStoredCards()` to still be able to provide feedback to the user when a error happens.

app/src/main/java/com/kickstarter/viewmodels/PaymentMethodsViewModel.kt
```diff
@@ -96,10 +110,22 @@ class PaymentMethodsViewModel(environment: Environment) : ViewModel(), PaymentMe
...
+                    this@PaymentMethodsViewModel.progressBarIsVisible.onNext(true)
+                    val ret = _getStoredCards()
+                    ret
+                }
+                .onSuccess {
+                    this@PaymentMethodsViewModel.progressBarIsVisible.onNext(false)
+                    this@PaymentMethodsViewModel.cards.onNext(it)
+                }
+                .onFailure {
+                    this@PaymentMethodsViewModel.progressBarIsVisible.onNext(false)
```
Arkariang, line 125: 👍

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
```diff
@@ -380,42 +388,38 @@ class KSApolloClientV2(val service: ApolloClient, val gson: Gson) : ApolloClient
...
 
-    override fun getStoredCards(): Observable<List<StoredCard>> {
-        return Observable.defer {
-            val ps = PublishSubject.create<List<StoredCard>>()
+    override suspend fun _getStoredCards(): List<StoredCard> {
+        val query = UserPaymentsQuery()
+        var ret = listOf<StoredCard>()
+        try {
+            // `single()` vs `singleOrNull()`
+            val response = this.service.query(query).toFlow().single()
```
Arkariang, line 396: could be changed over `val response = this.service.query(query).execute()` it internally uses `toFlow().single()`

app/src/main/java/com/kickstarter/viewmodels/PaymentMethodsViewModel.kt
```diff
@@ -140,9 +166,24 @@ class PaymentMethodsViewModel(environment: Environment) : ViewModel(), PaymentMe
         )
 
         compositeDisposable.add(
-            this.refreshCards
-                .switchMap { getListOfStoredCards() }
-                .subscribe { this.cards.onNext(it) }
+            this.refreshCards.subscribe {
```
Arkariang, line 169: question here, I assume this piece is for demonstration purposes, because this piece over here be just exchanging within the switch map `getListOfStoredCards()` over  `apolloClient.getStoredCards()`, let me know if I'm missing something!

PR #2240, opened 2025-02-25
[no-jira] Bottom nav bar appearing behind nav bar

leighdouglas:
# 📲 What

Bottom sheet on ppo screen was appearing behind nav bar for android 15 devices that opt to using the 3 button bottom navigation bar
Also, ModalBottomSheetLayout is meant to be top level

# 🛠 How

Adding navigation bar padding to bottom sheet layout
Moved bottom sheet layout to not be nested

# 👀 See
Before and after:
Screenshot_20250225_131159 Screenshot_20250225_132226

# 📋 QA

Test PPO ui on android 15 device as well as a device lower than 15. Make sure bottom sheet is not blocked by bottom nav bar

# Story 📖

no-jira


jlplks (approved):
LGTM

PR #2241, opened 2025-02-26
MBL-2144 Fix concurrency crash when preloading images

ycheng-kickstarter:
# 📲 What
Fatal Exception: java.util.ConcurrentModificationException
java.util.ArrayList$Itr.checkForComodification (ArrayList.java:1029)
java.util.ArrayList$Itr.next (ArrayList.java:982)
com.kickstarter.ui.fragments.RewardsFragment$onViewCreated$1$1$1$1.invoke (RewardsFragment.kt:78)

Per @tonyteate, make a copy of the rewards list before emitting it. That way we are not at risk of handling this list while the UseCase class is calling filteredRewards.clear() at various points in the filtering process.

Also use an iterator instead of a for loop just in case.

# 🤔 Why
Crash in 3.29.0 release:

https://console.firebase.google.com/u/0/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/92f9702e8d3411138f15d6e34957f0b8?versions=3.29.0%20(2014150933)&time=last-twenty-four-hours&types=crash&sessionEventKey=67BDCCCF033400012D758290E88FAAED_2053859627073866485

# 🛠 How

[Line 78](https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/ui/fragments/RewardsFragment.kt#L78) of RewardsFragment is:
```
                        for (reward in rewards) {
                            reward.image()?.let {
                                val request = ImageRequest.Builder(context)
                                    .data(reward.image()?.full())
                                    .build()
                                imageLoader.enqueue(request)
                            }
                        }
```

`rewards` comes from `val rewards = shippingUIState.filteredRw`. The GetShippingRulesUseCase filters out rewards and adds them to the list in an IO thread, which seems to be what the stacktrace is complaining about in terms of concurrent modification. So the patch is to use an iterator instead of a for loop.

# 👀 See

# 📋 QA
I wasn't able to reproduce the crash before, but please view rewards carousel and add ons screen and ensure no crashes.

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2144

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/usecases/GetShippingRulesUseCase.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/usecases/GetShippingRulesUseCase.kt
@@ -148,7 +148,7 @@ class GetShippingRulesUseCase(
                 loading = isLoading,
                 selectedShippingRule = defaultShippingRule,
                 error = errorMessage,
-                filteredRw = filteredRewards
+                filteredRw = filteredRewards.toList()
             )
         )
     }
--- a/app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/AddOnsScreen.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/AddOnsScreen.kt
@@ -28,7 +28,6 @@ import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import coil.Coil
-import coil.ImageLoader
 import coil.request.ImageRequest
 import com.kickstarter.R
 import com.kickstarter.libs.Environment
@@ -330,7 +329,7 @@ fun AddOnsScreen(
 }
 
 private fun preloadImages(context: Context, rewards: List<Reward>) {
-    val rewardsIterator = rewards.iterator();
+    val rewardsIterator = rewards.iterator()
     while (rewardsIterator.hasNext()) {
         val reward = rewardsIterator.next()
         reward.image()?.let {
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/kickstarter/viewmodels/usecases/GetShippingRulesUseCase.kt
```diff
@@ -148,7 +148,7 @@ class GetShippingRulesUseCase(
                 loading = isLoading,
                 selectedShippingRule = defaultShippingRule,
                 error = errorMessage,
-                filteredRw = filteredRewards
+                filteredRw = filteredRewards.toList()
```
ycheng-kickstarter, line 151: I think @tonyteate had the more correct solution here, which is to make a copy of the list before emitting it. That way we are not at risk of handling this list while the UseCase class is calling filteredRewards.clear() at various points in the filtering process.

ycheng-kickstarter:
@leighdouglas I think that might be a red herring! First the app crashes due to the concurrency exception, then when the ImageView tries to loadCircleImage, it correctly responds with `You cannot start a load for a destroyed activity` because the activity has crashed already.

app/src/main/java/com/kickstarter/viewmodels/usecases/GetShippingRulesUseCase.kt
leighdouglas: calling `.toList()` would make this an immutable list, would an immutable list be better here, since we are trying to modify the list while iterating through it?

ycheng-kickstarter: Nope, we want an immutable list here where it gets emitted. At the point where RewardsFragment is accessing this list, we don't want the list to be changing. We want the static values of the items in this list provided to RewardsFragment.

leighdouglas (approved):
LGTM but maybe get a second set of eyes on this! 🎉

Arkariang (approved):

Arkariang:
Nice work @ycheng-kickstarter and @tonyteate !

PR #2246, opened 2025-03-05
MBL-2157: Add Similar Projects GQL query & Apollo client method

tonyteate:
# 📲 What

Add a GraphQL query `FetchSimilarProjects` to the collection of Project operations, and introduce a new corresponding method `ApolloClientTypeV2.fetchSimilarProjects`.

# 🤔 Why

For any given Project, the app should be able to fetch a collection of similar or recommended Projects for a user. This is the base functionality for that capability.

# 🛠 How

In addition to the Apollo client method:
- Add another `projectTransformer` function to GraphQLTransformers for a `fragment.SimilarProject`
- Add `percentFunded` property to Project data model

# 📋 QA

# Story 📖

- https://kickstarter.atlassian.net/browse/MBL-2156
- https://kickstarter.atlassian.net/browse/MBL-2157

See the Epic for details

### DIFF
--- a/app/src/main/graphql/fragments.graphql
+++ b/app/src/main/graphql/fragments.graphql
@@ -292,6 +292,17 @@ fragment project on Project {
     slug
 }
 
+fragment similarProject on Project {
+    id
+    pid
+    name
+    slug
+    isLaunched
+    deadlineAt
+    percentFunded
+    imageUrl(width: $imageUrlWidth)
+}
+
 fragment reward on Reward {
     id
     name
--- a/app/src/main/graphql/project.graphql
+++ b/app/src/main/graphql/project.graphql
@@ -215,3 +215,18 @@ query BuildPaymentPlan($slug: String!, $amount: String!) {
     }
   }
 }
+
+query FetchSimilarProjects($first: Int = 4, $similarToPid: String!, $excludePid: Int!, $recommended: Boolean, $seed: Int, $imageUrlWidth: Int = 1024) {
+  projects(
+    first: $first
+    state: LIVE
+    similarToPid: $similarToPid
+    excludePids: [$excludePid]
+    seed: $seed
+    recommended: $recommended
+  ) {
+    nodes {
+      ...similarProject
+    }
+  }
+}
--- a/app/src/main/java/com/kickstarter/models/Project.kt
+++ b/app/src/main/java/com/kickstarter/models/Project.kt
@@ -40,6 +40,7 @@ class Project private constructor(
     private val launchedAt: DateTime?,
     private val location: Location?,
     private val name: String,
+    private val percentFunded: Int?,
     private val permissions: List<Permission?>?,
     private val pledged: Double,
     private val photo: Photo?,
@@ -99,6 +100,7 @@ class Project private constructor(
     fun launchedAt() = this.launchedAt
     fun location() = this.location
     fun name() = this.name
+    fun percentFunded() = this.percentFunded
     fun permissions() = this.permissions
     fun pledged() = this.pledged
     fun photo() = this.photo
@@ -161,6 +163,7 @@ class Project private constructor(
         private var launchedAt: DateTime? = null,
         private var location: Location? = null,
         private var name: String = "",
+        private var percentFunded: Int? = null,
         private var permissions: List<Permission?>? = null,
         private var pledged: Double = 0.0,
         private var photo: Photo? = null,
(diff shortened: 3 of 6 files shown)


app/src/main/graphql/fragments.graphql
```diff
@@ -292,6 +292,17 @@ fragment project on Project {
...
 
+fragment similarProject on Project {
+    id
+    pid
+    name
+    slug
+    isLaunched
+    deadlineAt
+    percentFunded
+    imageUrl(width: $imageUrlWidth)
```
tonyteate, line 303: Noticed `image: Photo` used elsewhere, can switch if preferred.

app/src/main/graphql/project.graphql
```diff
@@ -214,4 +214,19 @@ query BuildPaymentPlan($slug: String!, $amount: String!) {
       }
     }
   }
+}
+
+query FetchSimilarProjects($first: Int = 4, $similarToPid: String!, $excludePid: Int!, $recommended: Boolean, $seed: Int, $imageUrlWidth: Int = 1024) {
```
tonyteate, line 219: Can move this entire query to another file, e.g. `fetchSimilarProjects.graphql` if preferred.

app/src/main/java/com/kickstarter/models/Project.kt
```diff
@@ -40,6 +40,7 @@ class Project private constructor(
     private val launchedAt: DateTime?,
     private val location: Location?,
     private val name: String,
+    private val percentFunded: Int?,
```
tonyteate, line 43: `percentFunded` is available as a field in the API and was noted in the sample query. I see we already have `Project.percentageFunded()` which relies on `goal` and `pledged`. If necessary or preferred, I can either change the property name so as to not be so close, or include those fields in the GQL query and rely on the existing method.

app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
```diff
@@ -833,6 +834,32 @@ fun backingTransformer(backingGr: com.kickstarter.fragment.Backing?): Backing {
         .build()
 }
 
+/**
+ * Transform the Project GraphQL data structure into our own Project data model
+ * @param fragment.SimilarProject similarProjectFragment
+ * @return Project
+ */
+fun projectTransformer(similarProjectFragment: SimilarProject?): Project {
+    val id = decodeRelayId(similarProjectFragment?.id) ?: -1
```
tonyteate, line 843: It seems like we can use `pid` from the API now?

app/src/main/graphql/project.graphql
Arkariang: Love the suggestion!, we could have a `projects.graphql` where all the queries/mutations that affect projects (not a singular project) live. Could be done in a separate ticket so no need to change anything on this PR !

app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
Arkariang: For context:
`pid` -> will contain on graph just the project is (example: `1535688011`)
`id` -> on graph returned in a B64 format (example: `UHJvamVjdC0xNTM1Njg4MDEx` this example decoded will be `Project-1535688011`)

Using `id` to keep parity from V1  (the `Project` builder data model is inherited from V1 Rest API ), it has been extended according to needs to fit both API's.

Now a days the usage of V1 is marginal so it seems safe enough to use `pid` (without decodeRelay/encodeRelay) combos, I will vote for keeping consistency though so if we decide to go use `pid`  over `id`(with encode/decode) let's try to keep consistency across networking clients at least.

For this PR both fields seem good to me, whatever change we might wanna tackle can be done in separate ticket :)

app/src/main/java/com/kickstarter/models/Project.kt
Arkariang: This approach sounds good to me. 
Is another one of the examples in which our `Project` builder model is not reflecting latter additions to graphAPI 👍

Arkariang (approved):

app/src/main/graphql/project.graphql
tonyteate: Sounds like a great idea, will follow up later w/ a separate ticket 👍

app/src/main/java/com/kickstarter/services/transformers/GraphQLTransformers.kt
tonyteate: Makes sense, thanks for the context! Definitely agree that we should stay consistent. Will follow up later with a separate ticket for this as well 👍

app/src/main/java/com/kickstarter/models/Project.kt
tonyteate: Great, will add to mock factory in another PR. Thanks!

PR #2247, opened 2025-03-05
MBL-2135: SearchAndFilter viewmodel

Arkariang:
# 📲 What

- ViewModel now handles be able to search terms

# 🤔 Why

# 🛠 How

- on VM init debuncing searchterm will trigger the query with specific discovery params
- - if serarch term is empty will return sorting popular results
- - if search terms contains any term will return results for the terms and sorting popular.
ℹ️ (More sorting and filtering options will be extended in follow up tickets)
- Reftags added when pressing a project result of a search (either with or without terms, take a look at unit tests if more details required).
- Analytics events triggered with every search query execution ( 1CTA, 1 PAGE_VIEWED) with the discovery params.
ℹ️ (PAGE_VIEWD) even will be extended with pagination pages information on following tickets.
- Error handling via red snackbar with generic error message.

- `CancellationException` (within `KSApolloClient`) is not considered an error, when using `collectLatest` previous suspend functions/flow emissions will be canceled, this is the expected behaviour as we want to execute networking and update UI only with the most recent params and should not be considered an error that requires user feedback.

# 👀 See

https://github.com/user-attachments/assets/5021a53d-ef80-43e8-b951-5318d08750fd

|  |  |

# 📋 QA

- Either in staging or production perform some terms searching, when there is no search term a collection of "popular" projects will be presented.
- Try to navigate to some of the project results live or in prelaunch state.

# Story 📖

[MBL-2135](https://kickstarter.atlassian.net/browse/MBL-2135)

[MBL-2135]: https://kickstarter.atlassian.net/browse/MBL-2135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
@@ -68,7 +68,7 @@ class SearchAndFilterActivity : ComponentActivity() {
                 viewModel.provideErrorAction { message ->
                     lifecycleScope.launch {
                         snackbarHostState.showSnackbar(
-                            message = message ?: "Something went wrong",
+                            message = message ?: getString(R.string.Something_went_wrong_please_try_again),
                             actionLabel = KSSnackbarTypes.KS_ERROR.name,
                             duration = SnackbarDuration.Long
                         )
--- a/app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
@@ -7,6 +7,8 @@ import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.viewModels
 import androidx.compose.foundation.lazy.rememberLazyListState
+import androidx.compose.material.SnackbarDuration
+import androidx.compose.material.SnackbarHostState
 import androidx.compose.material.rememberScaffoldState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
@@ -14,6 +16,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.lifecycle.lifecycleScope
 import com.kickstarter.R
 import com.kickstarter.features.search.viewmodel.SearchAndFilterViewModel
 import com.kickstarter.libs.RefTag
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
```diff
@@ -1,90 +1,175 @@
...
+
+    /**
+     * Returns a project and its appropriate ref tag given its location in a list of popular projects or search results.
+     *
+     * @param searchTerm        The search term entered to determine list of search results.
+     * @param projects          The list of popular or search result projects.
+     * @param selectedProject   The project selected by the user.
+     * @return The project and its appropriate ref tag.
+     */
+    private fun projectAndRefTag(
```
Arkariang, line 144: This method has been ported directly from the old search experience.

app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
```diff
@@ -1,90 +1,175 @@
...
+    init {
+        scope.launch {
+            searchTerm
+                .debounce(debouncePeriod)
+                .onEach { debouncedTerm ->
+                    // - Reset to initial state in case of empty search term
+                    if (debouncedTerm.isEmpty() || debouncedTerm.isBlank()) {
+                        _params.emit(popularDiscoveryParam)
+                    } else
+                        _params.emit(
```
Arkariang, line 78: In future additions, with category selection or project sort, there will be inputs to this VM (similar to `updateSearchTerm`) that will add new emissions to  _params with each user selection.
In the future will likely not collect `searchTerm` chain but rather `params`

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
```diff
@@ -1911,6 +1912,9 @@ class KSApolloClientV2(val service: ApolloClient, val gson: Gson) : ApolloClient
     private suspend fun <T> executeForResult(block: suspend () -> T): Result<T> =
         try {
             Result.success(block())
+        } catch (cancellationException: CancellationException) {
+            // - When using try catch blocks with suspending functions always rethrow CancellationExceptions and not threat them as an error
+            throw cancellationException
```
tonyteate, line 1917: Nice 👌

app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
```diff
@@ -43,28 +48,40 @@ class SearchAndFilterActivity : ComponentActivity() {
...
 
                 val isTyping by remember { mutableStateOf(false) }
 
                 val lazyListState = rememberLazyListState()
 
+                val snackbarHostState = remember { SnackbarHostState() }
+
+                viewModel.provideErrorAction { message ->
+                    lifecycleScope.launch {
+                        snackbarHostState.showSnackbar(
```
tonyteate, line 70: Might enqueue multiple snackbars while searching

tonyteate (approved):
Lgtm! 👍

PR #2244, opened 2025-03-01
MBL-2132 FilterMenuBottomSheet For category UI

jlplks:
# 📲 What

Added CategorySelectionSheet to display a bottom sheet menu when a category pill is pressed. This implementation includes UI elements.

# 🤔 Why

To improve the user experience by providing a structured and interactive way to select categories through a bottom sheet. This UI component is essential for the upcoming full integration of category-based filtering in SearchActivity.

# 🛠 How

Implemented CategorySelectionSheet, which triggers when a category pill is pressed.
The bottom sheet includes:
A list of categories with toggles for selection.
A loading state to indicate data fetching.
Two buttons (Apply and Reset) with enabled/disabled states.
Since this ticket is only for UI, mock data is used for the categories.
Ensured that the category pill reflects the selected state correctly.

# 👀 See

[categories_bottom_sheet.webm](https://github.com/user-attachments/assets/4750f784-b933-4b6b-9001-105d70600086)

# 📋 QA

Steps to Test:
Tap a category pill.
Verify that the bottom sheet appears.
Check that categories are listed with toggles.
Toggle a category and ensure Apply/Reset buttons update accordingly.
Run UI tests to verify expected behavior.
# Story 📖

[MBL-2132 FilterMenuBottomSheet For category UI](https://kickstarter.atlassian.net/browse/MBL-2132)

PR #2254, opened 2025-03-11
MBL-2130 Pillbar UI

ycheng-kickstarter:
# 📲 What

Pillbar UI for new Search and Discover experience

# 🤔 Why

# 🛠 How

Updated colors and padding from https://github.com/kickstarter/android-oss/pull/2244 to spec and added Sort pill

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screenshot_20250311_155617 | Screenshot_20250311_163356 |

# 📋 QA

Go to search screen to see UI. Currently does not have VM logic to update state of buttons.

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2130


app/src/main/java/com/kickstarter/ui/views/compose/search/SearchTopBar.kt
```diff
@@ -59,17 +53,20 @@ fun SearchTopBarPreview() {
         SearchTopBar(
             onBackPressed = {},
             onValueChanged = {},
+            selectedFilterCounts = mapOf(),
```
Arkariang, line 56: You might wanna initialize the map here with values similar to the Screen 
mapOf(
        FilterRowPillType.SORT.name to 0,
        FilterRowPillType.CATEGORY.name to 0
    )

the compose preview is failing and the design system screen crashes due to missing key SORT
Screenshot 2025-03-12 at 10 48 02 AM

app/src/main/java/com/kickstarter/ui/views/compose/search/SearchTopBar.kt
```diff
@@ -165,38 +162,41 @@ fun SearchTopBar(
...
+    onCategoryPressed: () -> Unit,
+) {
+    Row(
+        modifier = Modifier
+            .fillMaxWidth()
+            .padding(dimensions.paddingMedium, dimensions.none, dimensions.none, dimensions.none),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        IconPillButton(
+            selectedFilterCounts.getValue(FilterRowPillType.SORT.name) > 0,
```
Arkariang, line 187: Could use here `getOrDefault` instead of `getValue` to avoid crashes if empty map

ycheng-kickstarter: Oo nice!

Arkariang (approved):
🔥

PR #2255, opened 2025-03-13
MBL-2137: ViewModel for FilterMenuBottomSheet

Arkariang:
# 📲 What

- ViewModel specific for the BottomSheet Filter Menu, for Phase 1 it will contain only the query to obtain rootCategories, will be increasing functionality on following phases.

# 🤔 Why

- We need to get the categories from somewhere without ending with 1 only massive viewmodel once all phases completed.

# 🛠 How

- New VM, calling new query
- categories list available in compose to be feed to the UI

# 👀 See

- No user facing changes but take a look on logcat you can see logs with the results from the query loading the search screen
Screenshot 2025-03-12 at 6 51 10 PM

- Can take a look at the network profiler 
Screenshot 2025-03-12 at 6 51 42 PM

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

# 📋 QA

# Story 📖

### DIFF
--- a/app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
@@ -10,6 +10,7 @@ import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material.SnackbarDuration
 import androidx.compose.material.SnackbarHostState
 import androidx.compose.material.rememberScaffoldState
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
@@ -18,6 +19,7 @@ import androidx.compose.runtime.setValue
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.lifecycle.lifecycleScope
 import com.kickstarter.R
+import com.kickstarter.features.search.viewmodel.FilterMenuViewModel
 import com.kickstarter.features.search.viewmodel.SearchAndFilterViewModel
 import com.kickstarter.libs.RefTag
 import com.kickstarter.libs.utils.ThirdPartyEventValues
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
```diff
@@ -37,46 +40,41 @@ import com.kickstarter.ui.activities.compose.search.SearchScreen
...
                 val searchedProjects = searchUIState.searchList
-
                 val isLoading = searchUIState.isLoading
 
-                val isTyping by remember { mutableStateOf(false) }
-
-                val lazyListState = rememberLazyListState()
+                val categoriesState by filterMenuViewModel.filterMenuUIState.collectAsStateWithLifecycle()
+                // TODO: send the list of categories to the BottomSheet coordinate with MBL-2171
+                val categories = categoriesState.categoriesList
```
Arkariang, line 75: @ycheng-kickstarter ready to use! but if you don't make it to do it on MBL-2171 let me know and I'll create another ticket :)

app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
```diff
@@ -117,10 +116,40 @@ class SearchAndFilterActivity : ComponentActivity() {
...
+                    }
+                }
             }
         }
     }
 
+    @Composable
+    private fun SetUpErrorActions(snackbarHostState: SnackbarHostState) {
+        val errorAction = { message: String? ->
+            lifecycleScope.launch { // TODO: store SnackbarResult on VM to consult it before showing new  one, in case there is multiple enqueue snackbars.
```
Arkariang, line 136: Will handle on follow up ticket once the two bottomSheets are wired up!

tonyteate:
Lgtm 🤙

tonyteate (approved):
Lgtm 🤙

app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
ycheng-kickstarter: Thanks! Used it in https://github.com/kickstarter/android-oss/pull/2257

PR #2260, opened 2025-03-19
MBL-2131: SortMenuBottomSheet UI component

leighdouglas:
# 📲 What

Sort button on search screen opens bottom sheet that shows sort options

# 🤔 Why

sort and filter for project search

# 🛠 How

- Call the appropriate bottom sheet layout depending on which button was pressed
- Persist the sort and filter values the user has selected, for updating the sort/filter params when one of them changes, and also for when the user reopens the bottom sheets
- Dismiss button on both sheets does not clear the sort/filter values

# 👀 See

https://github.com/user-attachments/assets/9447e886-9ef8-4cbf-9b9a-c15e5337c225

# 📋 QA

Test sort and filter selections are persisted when one changes, that the correct values are being sent, and that changing one does not reset the other

# Story 📖

[MBL-2131: SortMenuBottomSheet UI component](https://kickstarter.atlassian.net/browse/MBL-2131)


app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
```diff
@@ -187,35 +187,70 @@ fun SearchScreen(
...
+    )
+
+    val sortSheetState = rememberModalBottomSheetState(
+        initialValue = Hidden,
+        skipHalfExpanded = false
+    )
 
     ModalBottomSheetLayout(
-        sheetState = sheetState,
+        sheetState = if (activeBottomSheet.value == "categories") categorySheetState else sortSheetState,
```
Arkariang, line 208: instead of String comparison with categories you can use now `activeBottomSheet == FilterRowPillType.CATEGORY`

app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
```diff
@@ -187,35 +187,70 @@ fun SearchScreen(
...
+                                } else {
+                                    selectedFilterCounts[FilterRowPillType.CATEGORY.name] = 1
+                                }
+                            }
+                        },
+                        isLoading = false
+                    )
+                }
+
+                "sort" -> {
```
Arkariang, line 242: Now you can use here `FilterRowPillType.SORT` instead of the string "sort"

app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
```diff
@@ -248,12 +283,15 @@ fun SearchScreen(
                         },
                         selectedFilterCounts = selectedFilterCounts,
                         onSortPressed = {
-                            // Add logic here
+                            activeBottomSheet.value = "sort"
```
Arkariang, line 286: Now `activeBottomSheet.value = FilterRowPillType.SORT` and similar change on line 292

Arkariang (dismissed):

Arkariang (approved):

PR #2275, opened 2025-03-24
MBL-2238 && MBL-2213: Pagination polish

Arkariang:
# 📲 What

- MBL-2213: changed context property value for discover_search_results_count, now instead of sending the page, we send the total number of results the query returns.
- MBL-2238: Bug raised during QA, was unable to reproduce but found that we called viewModel.loadMore() on initial state (that method should only be called when fetching following pages, so we were attempting to fetch initial page twice). To solve this issue before calling on the UI viewModel.loadMore() we make sure is not loading already.
- Additional QA: search for really long sentences, and try to obtain no results from the query to show the "no results" screen, now  loadMore is not triggered in this scenario,  but before the fix "no results" screen was triggering loadMore . The fix for this scenario is to check the query has no more results to show.

# 🛠 How

 - For MBL-2213 added `totalCount` as field to to `SearchEnvelope`, now we send  that field as value for the property `discover_search_results_count` . tThe analytic events are sent to segment **only** on the first page load.

- For MBL-2238 added  `!isLoading` check before calling `loadMore()`

# 👀 See

This first screenshot has been taken with Feature flag off (or go to `ActivityExt:87` and  send false as the ff value)
OldSearchExperience

This second screenshot has been taken after the change were the property `discover_search_results_count` has now the total value of projects instead of the page, matching the values sent in the old experience)
NewSearchExperience

|||||||||

Pagination

https://github.com/user-attachments/assets/954d2eab-ea65-431d-9d37-3cbdd1c0096f

|||||||||||

Search for a really long sentence that will show the empty view several times.

https://github.com/user-attachments/assets/d3a6c2bb-adb5-4fed-9aaa-e7d88e572c2b

|  |  |

# 📋 QA

- For MBL-MBL-2213:
- - -  Go to Segment, you should see a CTA and PageViewed event generated with the default values, scroll and load several pages, no new events should be generated when loading new pages.
- - -  Now search for a term you should see a CTA and PageViewed event generated with the default values, scroll and load several pages, no new events should be generated when loading new pages.
- - - Now filter by category or Sort, you should see a CTA and PageViewed event generated with the default values, scroll and load several pages, no new events should be generated when loading new pages.

- For MBL-2238 was unable to reproduce as the video attached shows but try to break it, after scrolling to fetch new pages come back to the begining of the list you should see the first card being the big version. Any time the search gets updated either with a term or with new sorting/category options do execute the same. 

- **Additional bonus QA** search for a really long sentence likely something that will return no results to see hoe the empty results view behaves.

# Story 📖

[MBL-2238](https://kickstarter.atlassian.net/browse/MBL-2238) -> BUG
[MBL-2213](https://kickstarter.atlassian.net/browse/MBL-2213) -> Analytic update for property `discover_search_results_count`

[MBL-2238]: https://kickstarter.atlassian.net/browse/MBL-2238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
```diff
@@ -125,8 +125,8 @@ class SearchAndFilterActivity : ComponentActivity() {
                 }
 
                 val lifecycleOwner = LocalLifecycleOwner.current
-                LaunchedEffect(shouldLoadMore, lifecycleOwner.lifecycle.currentState) {
-                    if (shouldLoadMore && lifecycleOwner.lifecycle.currentState == Lifecycle.State.RESUMED) {
+                LaunchedEffect(shouldLoadMore, lifecycleOwner.lifecycle.currentState, isLoading) {
+                    if (shouldLoadMore && lifecycleOwner.lifecycle.currentState == Lifecycle.State.RESUMED && !isLoading) {
                         viewModel.loadMore()
```
Arkariang, line 130: `viewModel`.loadMore() was called on initial first load, now we are checking as well if it was not previously loading

Arkariang: Checking as well `hasMorePages` due to when having empty results from the query and showing the empty view `.loadMore` was detected to be triggered again.

ycheng-kickstarter (approved):
Bug no longer reproducible

leighdouglas (approved):

PR #2286, opened 2025-04-01
MBL-2248 [Pledge] The system does not allow pledging when  adding add-ons

jlplks:
# 📲 What

There was an error on some checkouts when selecting add-ons. This was cause previously when this was implemented the Rewards didn't have the AllowedAddOns.
Fixes the issue where users were able to select add-ons that were not allowed for the selected reward, which caused checkout errors later in the pledge flow.

# 🤔 Why

Previously, when a user selected a reward, all available add-ons were loaded, regardless of whether they were permitted for that specific reward. This caused situations where a user could select an invalid reward + add-on combination.

When proceeding to the final pledge step (createCheckout mutation), the backend rejected the request due to the mismatch, resulting in an error that was hard to trace.

# 🛠 How

- Ensured the app only loads and displays allowed add-ons for the selected reward.

- Updated logic to correctly filter allowedAddOns when reward is selected.

- Verified that only valid add-ons are passed downstream into the checkout flow.

- Updated any affected transformer or GraphQL response processing logic to use allowedAddOns instead of allAddOns.

# 👀 See

SS of the different 

|  Old query: FetchProjectRewards 🐛 | New Query 🦋 |
| --- | --- |
| Screenshot 2025-04-01 at 2 47 58 p m | Screenshot 2025-04-01 at 1 27 59 p m |

# 📋 QA

Launch any project with add-ons enabled.

Select a reward with specific allowed add-ons.

Confirm that:

Only the allowed add-ons for that reward are visible.

Complete the pledge and confirm checkout works without errors.

# Story 📖

[MBL-2248 [Pledge] The system does not allow pledging when  adding add-ons](https://kickstarter.atlassian.net/browse/MBL-2248)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
@@ -2,7 +2,6 @@ package com.kickstarter.viewmodels.projectpage
 
 import android.annotation.SuppressLint
 import android.os.Bundle
-import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.viewModelScope
@@ -10,7 +9,6 @@ import com.kickstarter.libs.Environment
 import com.kickstarter.libs.utils.RewardUtils
 import com.kickstarter.libs.utils.extensions.isNotNull
 import com.kickstarter.libs.utils.extensions.pledgeAmountTotalPlusBonus
-import com.kickstarter.mock.factories.LocationFactory
 import com.kickstarter.mock.factories.RewardFactory
 import com.kickstarter.models.Backing
 import com.kickstarter.models.Project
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
```diff
@@ -187,14 +187,15 @@ class AddOnsViewModel(val environment: Environment, bundle: Bundle? = null) : Vi
         }
     }
 
+    @SuppressLint("LogNotTimber")
     private fun getAddOns(selectedShippingRule: ShippingRule) {
```
Arkariang, line 191: selectedShippingRule us no longer required I assume

Arkariang:
Was able to test by using in staging the project with late pledges enabled -> Partial refunds - late pledge https://staging.kickstarter.com/projects/816433143/artisanal-kale-tarot. 
Was able to pledge to the reward "BACK THIS ONE" plus a digital reward AddOn

 
Screenshot 2025-04-02 at 1 15 50 PM

Have not being able to find a project with shipping addOns

Arkariang (approved):

PR #2287, opened 2025-04-02
MBL-2276 Adopt order object

ycheng-kickstarter:
# 📲 What

Graphql and model changes to introduce a barebones Order object to our codebase with just a couple of fields. Can add more relevant fields once the backend updates it for net new backers. For this PR we're just grabbing it off the Backing object for now.

# 🤔 Why

PM transactions are captured by the Order object, so we need to implement the Order object on mobile to support PM.

# 🛠 How

# 👀 See

Nothing to see visually.

# 📋 QA

Find a project and backing that is in the right state for PM (i.e. has a non-null Order). Put breakpoint at BackingFragmentViewModel#278 and verify that the backing object contains the order object with fields populated. 

# Story 📖

MBL-2276


tonyteate:
Lgtm, but could not QA w/ the project or user from the PM ticket. Can someone point out what I'm missing?

leighdouglas (approved):
lgtm!

PR #2296, opened 2025-04-09
MBL-2307 [QA] The system does not allow pledging when  adding add-ons

jlplks:
# 📲 What

This PR updates the logic used to fetch and display add-ons during the pledge flow, ensuring only valid shipping options are presented based on the selected reward and shipping rule.

# 🤔 Why

Previously, when a user selected a "Reward shipping everywhere" reward, in some cases, the system incorrectly included restricted shipping add-ons in the available options. This behavior did not match the web or iOS platforms and led to users selecting incompatible add-ons, resulting in a checkout error ("Sorry, something went wrong").
# 🛠 How

Updated getRewardAllowedAddOns() to include shipping location (Location) in the GraphQL query.

Modified getAddOnsFromAllowedAddOns() to extract and retain relevant shipping rule data per add-on.

Refactored getUpdatedList() to:

Filter add-ons that do not include the currently selected shipping location in their shipping rules.

Ensure that backed add-ons retain their quantity if still eligible.
# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| [error_shipping_addons.webm](https://github.com/user-attachments/assets/78a94a4f-50bb-4881-be56-a75aa0a2eef6) | [fix_shipping_addons.webm](https://github.com/user-attachments/assets/19930d8e-76c7-4c3b-9047-bfb93d01f4d4) |

# 📋 QA

Check [This project ](https://staging.kickstarter.com/projects/1768690592/purple-watch-3-d?ref=nav_search&result=project&term=purple%20watch&total_hits=5)

Before when you selected Reward shipping everywhere it returns 3 add-ons:
1-Digital reward
2-reward shipping everywhere
3-restricted rewards shipping 

The third option doesn't appear on Web or iOS and when we selected it it end up causing an error on checkout because it wasn't suppose to be there.

Now it should only show:
1-Digital reward
2-reward shipping everywhere

# Story 📖 
[MBL-2307 [QA] The system does not allow pledging when  adding add-ons](https://kickstarter.atlassian.net/browse/MBL-2307)

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
@@ -8,8 +8,10 @@ import com.kickstarter.libs.Environment
 import com.kickstarter.libs.utils.RewardUtils
 import com.kickstarter.libs.utils.extensions.isNotNull
 import com.kickstarter.libs.utils.extensions.pledgeAmountTotalPlusBonus
+import com.kickstarter.mock.factories.LocationFactory
 import com.kickstarter.mock.factories.RewardFactory
 import com.kickstarter.models.Backing
+import com.kickstarter.models.Location
 import com.kickstarter.models.Project
 import com.kickstarter.models.Reward
 import com.kickstarter.models.ShippingRule
@@ -154,7 +156,7 @@ class AddOnsViewModel(val environment: Environment, bundle: Bundle? = null) : Vi
                 }
             }
 
-            getAddOns()
+            getAddOns(shippingRule)
         }
     }
 
@@ -182,11 +184,12 @@ class AddOnsViewModel(val environment: Environment, bundle: Bundle? = null) : Vi
     fun provideSelectedShippingRule(shippingRule: ShippingRule) {
         if (this.shippingRule != shippingRule) {
             this.shippingRule = shippingRule
-            getAddOns()
+            getAddOns(shippingRule)
         }
     }
 
-    private fun getAddOns() {
+    private fun getAddOns(selectedShippingRule: ShippingRule) {
+
         // - Do not execute call unless reward has addOns
         if (currentUserReward.hasAddons()) {
             scope.launch(dispatcher) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
```diff
@@ -182,26 +184,28 @@ class AddOnsViewModel(val environment: Environment, bundle: Bundle? = null) : Vi
...
+    private fun getAddOns(selectedShippingRule: ShippingRule) {
+
         // - Do not execute call unless reward has addOns
         if (currentUserReward.hasAddons()) {
             scope.launch(dispatcher) {
                 apolloClient
                     .getRewardAllowedAddOns(
                         slug = project.slug() ?: "",
                         rewardId = currentUserReward.id(),
+                        locationId = selectedShippingRule.location() ?: LocationFactory.empty()
```
tonyteate, line 200: Minor thing: We can use the builder itself here & on 208 `Location.builder().build()`, or with `.id(-1)` if that's important, so the factories are kept for tests (not sure if we're consistent w/ this throughout though).

tonyteate:
Not sure why, as the logic here from the previous PR makes sense, but when I check that same project ["Purple Watch 3 D"](https://staging.kickstarter.com/projects/1768690592/purple-watch-3-d), shipping location "United States", select "pick up yer stickers" as the Reward:

- On the web for Add-Ons I see: "Digital reward" and "pick up yer stickers". 
- On Android I see: "Digital reward"

jlplks:
> Not sure why, as the logic here from the previous PR makes sense, but when I check that same project ["Purple Watch 3 D"](https://staging.kickstarter.com/projects/1768690592/purple-watch-3-d), shipping location "United States", select "pick up yer stickers" as the Reward:
> 
> * On the web for Add-Ons I see: "Digital reward" and "pick up yer stickers".
> * On Android I see: "Digital reward"
> 
> Double check though 🙏

Just checked, the filter needed to add the 'local' on the  shippingPreference. Uploading the fix

app/src/main/java/com/kickstarter/viewmodels/projectpage/AddOnsViewModel.kt
jlplks: Changed and honestly I don't know if we're consistent with that

tonyteate (approved):
Incredible work @jlplks 🔥

PR #2298, opened 2025-04-10
MBL-2225 && MBL-2223: Navigation between Main Filter menu screen (includes project state filter), and the Category Selection screen

Arkariang:
# 📲 What

- Now the user can navigate between menu options within the same bottomSheet, 
- pressing the filter pillButton, can navigate to the categories screen.
- pressing the categories pillButton,  can navigate back to the main filter menu screen.
- pressing the project status pillButton, has the same behaviour as the filter button.

# 🤔 Why
- Improve filtering experience.

# 👀 See
- Feature working on Emulator (MBL-2225)

https://github.com/user-attachments/assets/b2e97fa0-7a31-4dbe-bd8b-f441ae01a084

- Compose preview interactive mode supports the navigation as well (MBL-2223)

https://github.com/user-attachments/assets/5485a33e-f979-4794-b98f-01777eb1b3f5

|  |  |

# 📋 QA

All related to category selection keeps working exactly as in Phase 1, except for reset button updates.
- Select 1 category
- Apply 1 category
- Reset 1 category (it does not dismiss bottomSheet automatically, but the query with results with default category is executed, and projects list updated accordingly)

Single project status selection keeps working as usual
- Select 1 project status
- Apply 1 project status
- Reset 1 project status (it does not dismiss bottomSheet automatically, but the query with results with default project state is executed, and projects list updated accordingly)

Double selection with navigation:
- Select category and project status (regardless the order of the selection, navigate back and forth and use the filter pill button and the project status pill button).
- Apply the category and project status
- Reset button return to the default state both category and project status, it does not dismiss bottomSheet.

# Story 📖

[MBL-2225](https://kickstarter.atlassian.net/browse/MBL-2225)
[MBL-2223](https://kickstarter.atlassian.net/browse/MBL-2223)

[MBL-2225]: https://kickstarter.atlassian.net/browse/MBL-2225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
@@ -491,6 +491,7 @@ fun FilterAndCategoryPagerSheet(
                 onApply = { selectedProjectState, applyAndDismiss ->
                     projectState.value = selectedProjectState
                     if (applyAndDismiss != null) {
+                        // - Reset to default values
                         if (applyAndDismiss.isFalse()) {
                             category.value = null
                         }
@@ -518,6 +519,7 @@ fun FilterAndCategoryPagerSheet(
                 onApply = { selectedCategory, applyAndDismiss ->
                     category.value = selectedCategory
                     if (applyAndDismiss != null) {
+                        // - Reset to default values
                         if (applyAndDismiss.isFalse()) {
                             projectState.value = null
                         }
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
```diff
@@ -399,14 +441,137 @@ fun SearchScreen(
...
+                },
+                currentCategory = currentCategory,
+                onDismiss = onDismiss,
+                categories = categories,
+                onApply = { selectedCategory, applyAndDismiss ->
+                    category.value = selectedCategory
+                    if (applyAndDismiss != null) {
+                        // - Reset to default values
+                        if (applyAndDismiss.isFalse()) {
+                            projectState.value = null
```
tonyteate, line 524: I know at some point "Reset" might become "Reset all", just feels a little surprising to affect Project Status while looking at Categories.

Arkariang: it's a product requirement, when hitting reset (or apply for the case), all user selection must be applied. Both categories and project statuses.

Arkariang:
> Might need additional state to track these updates. Here selecting 'Late Pledges' Status, then 'Comics' Category. Going Back, Status is reset. Category remains selected, but not reflected in the UI.
> 
>  Screen_recording_20250414_060624.mp4

Will take a look, when going back status should keep being the selected one. 👍

PR #2300, opened 2025-04-11
MBL-2321 Fix bug that caused project page to infinite load for PM backers

ycheng-kickstarter:
# 📲 What

Users with an order could not load project page

# 🤔 Why

Order id was incorrectly parsed as a Long instead of a String

# 🛠 How

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screenshot_20250410_094049 | Screenshot_20250411_093347 |

# 📋 QA

Obtain a user that has gone through PM checkout. View the project page of that PM project.
Users in test state can be found here: https://kickstarter.atlassian.net/browse/MBL-2313

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2321

### DIFF
--- a/app/src/main/java/com/kickstarter/models/Order.kt
+++ b/app/src/main/java/com/kickstarter/models/Order.kt
@@ -10,7 +10,7 @@ import kotlinx.parcelize.Parcelize
  */
 @Parcelize
 class Order private constructor(
-    private val id: Long,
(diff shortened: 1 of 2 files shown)


tonyteate (approved):
Will update if decoded relay id is needed 👍

PR #2302, opened 2025-04-16
MBL-2266  The "pledge" button is disabled after attempting to pay with the wrong card

jlplks:
# 📲 What

Fixes a bug where the "Pledge" button remains disabled after a failed pledge attempt in Late Pledge Checkout flow due to invalid card input or failed payment intent creation.

# 🤔 Why

This issue was causing a broken user experience during the Late Pledge flow: if a backer selected an invalid card (e.g., test card 4000000000000341) and the payment failed, the "Pledge" button stayed disabled. This prevented the user from retrying the pledge without restarting the entire flow.

# 🛠 How

Created a new method: resetPledgeButtonUIOnFailure(errorMessage: String?) in LatePledgeCheckoutViewModel.

This method ensures the pledge button is re-enabled and the UI loading state is reset when:

- A payment intent fails to create

- Card validation fails

- 3DS authentication is invalid

- Stripe returns a lastPaymentError

Updated all failure paths to use this method.

Added tests to confirm this behavior is consistent and that the button state resets correctly.

# 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| [Error](https://github.com/user-attachments/assets/a85b84fe-c52d-4970-8be8-62312fc32b80) | [FIX](https://github.com/user-attachments/assets/bc08094a-def4-48b6-baf4-485554d45041) |

# 📋 QA

Steps to Reproduce:
- Go to the app and log in.

- Search and select a project in Late Pledge mode.

- Tap Back this project and select a reward.

- Proceed to the checkout screen.

- Add or select an invalid card (e.g., 4000000000000341 or 4000002760003184).

- Click Pledge button.

- Observe the behavior after the failure.

 Expected
- The pledge button becomes enabled again after a failed attempt.

- User is free to correct or select a different card without restarting the flow.

# Story 📖

[[Regression- checkout-Pledge late] The "pledge" button is disabled after attempting to pay with the wrong card](https://kickstarter.atlassian.net/browse/MBL-2266)


Arkariang:
@jlplks Jira ticket on the PR title!

tonyteate (approved):

PR #2303, opened 2025-04-17
MBL-2319: Open a Backing Details webview for users who have gone through PM: Messages Activity

leighdouglas:
# 📲 What

Changed entry point in Messages for view pledge to backing details webview for backers that have completed pledge management. 

# 🤔 Why

Rerouting users to the mobile web while we build out pledge management natively in "View Your Pledge" screen. Backer's need to be able to see their receipts in the meantime. 

# 🛠 How

- Created new feature flag
- Check for order on `Backing` object, and check if checkout is complete
- Updated v1 REST api call `fetchBacking` to graphql `getProjectBacking` query for one use case. `Order` has only been updated on the `Backing` struct in graph, and not on v1
- Created a few helper methods to check the Backing object for an Order, and if the order is complete, as well as an Activity extension function to open BackingDetails web view. Will update PPO to use this ext function in subsequent PRs.

# 👀 See

# 📋 QA

See https://kickstarter.atlassian.net/browse/MBL-2313 for accounts you can log into for QA
Test with project Sustainable Cooler 2.0

# Story 📖

[MBL-2319: Open a Backing Details webview for users who have gone through PM: Messages Activity](https://kickstarter.atlassian.net/browse/MBL-2319)


app/src/main/java/com/kickstarter/libs/featureflag/FeatureFlagClient.kt
```diff
@@ -82,6 +82,7 @@ enum class FlagKey(val key: String) {
     ANDROID_PLEDGED_PROJECTS_OVERVIEW_V2("android_pledged_projects_overview_v2"),
     ANDROID_SEARCH_FILTER("android_search_filter"),
     ANDROID_REWARD_SHIPMENT_TRACKING("android_reward_shipment_tracking"),
+    ANDROID_NET_NEW_BACKERS_WEBVIEW("android_net_new_backers_webview")
```
ycheng-kickstarter, line 85: Would android_completed_pm_checkout_webview would be more inclusive?

app/src/main/java/com/kickstarter/libs/utils/extensions/BackingExt.kt
```diff
@@ -21,6 +22,10 @@ fun Backing.isShippable(): Boolean {
     return isShippable(reward)
 }
 
+fun Backing.isOrderPresentAndComplete(): Boolean {
```
ycheng-kickstarter, line 25: 💯

ycheng-kickstarter:
QA looked great 🥰 just a renaming suggestion to avoid confusion in future

app/src/main/java/com/kickstarter/libs/featureflag/FeatureFlagClient.kt
leighdouglas: updated to your naming suggestion, was keeping parity with ios but i feel this better reflects the bucket of users that fall into this case

PR #2313, opened 2025-04-25
[no-jira] Disable track shipment button when tracking url null

leighdouglas:
# 📲 What

Previously, we would hide the tracking card in activity feed when the project, tracking number, or tracking url were null. now, we only hide when project or tracking number are null, and just disable the track shipment button when tracking url null. 

Also added tests for the activity tracking modal

# 🤔 Why

Users should still be able to see the tracking number card if the tracking url is null

# 👀 See

| Enabled 🐛 | Disabled 🦋 |
| --- | --- |
| button enabled | disabled button |

# 📋 QA

Ping me if you need credentials for a test account with reward shipping cards available in the activity.  
If you set the tracking number to null in the `RewardShippedViewHolder` and run the app, you should see that the button is disabled. 

# Story 📖

no-jira


ycheng-kickstarter (approved):
Makes sense!

PR #2324, opened 2025-05-14
MBL-2265 [Regression- Manage your pledge]The information is not loading

jlplks:
# 📲 What

Refactored BackingFragmentViewModel to aggregate related outputs into a single BackingState model. This addressed a bug where the backing screen would sometimes remain in a perpetual loading state and fail to display data. 

# 🤔 Why

The UI showed a loading state (e.g., isRefreshing = true), but the actual data wasn't emitted or rendered, and the screen would appear stuck in a loading spinner indefinitely.

This issue was intermittent and difficult to reproduce, but we observed it happening occasionally during QA. To prevent this, we centralized the ViewModel's state using a BackingState data class and converted the internal state flow into a single reactive source of truth.

This change also solved test flakiness in testSwipeRefresherProgressIsVisible, which used to fail due to uncoordinated state emissions.

# 🛠 How

- Introduced BackingState, a unified state model containing all relevant UI properties (loading, refreshing, errors, etc.).

- Consolidated logic inside the ViewModel to only update BackingState, then derived all UI outputs (Observables) via map from that state.

- Updated BackingFragment to reflect this architecture and simplified its subscriptions.

- Rewrote testSwipeRefresherProgressIsVisible to assert against a predictable [true, false] emission pattern after a refresh.

- Reduced the chances of state desynchronization, ensuring data always accompanies visible UI loading states.

# 📋 QA

Launch the app and open a backed project.

Confirm that project details and rewards load properly.

Pull-to-refresh the screen and confirm the swipe indicator appears and disappears as expected.

Observe that the app does not remain stuck in a loading state.

# Story 📖

[MBL-2265 [Regression- Manage your pledge]The information is not loading](https://kickstarter.atlassian.net/browse/MBL-2265)


app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
tonyteate, line 1: 🔑

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
```diff
@@ -535,11 +574,21 @@ interface BackingFragmentViewModel {
...
+                backing != null -> Observable.just(BackingState.Success(backing))
+                !slug.isNullOrBlank() -> Observable.concat(
+                    Observable.just(BackingState.Loading),
+                    apolloClient.getProjectBacking(slug)
+                        .timeout(15, TimeUnit.SECONDS)
+                        .retry(1)
+                        .map<BackingState> { BackingState.Success(it) }
+                        .onErrorReturn { BackingState.Error(it) }
+                )
+                else -> Observable.just(BackingState.Error(IllegalArgumentException("Missing backing and slug")))
```
tonyteate, line 591: Would we show this string without translation?

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
```diff
@@ -535,11 +574,21 @@ interface BackingFragmentViewModel {
...
+        private fun getBackingInfo(projectData: ProjectData): Observable<BackingState> {
+            val backing = projectData.backing()
+            val slug = projectData.project().slug()
+
+            return when {
+                backing != null -> Observable.just(BackingState.Success(backing))
+                !slug.isNullOrBlank() -> Observable.concat(
+                    Observable.just(BackingState.Loading),
+                    apolloClient.getProjectBacking(slug)
+                        .timeout(15, TimeUnit.SECONDS)
```
tonyteate, line 586: Not sure if we have a reference point, but perhaps drop this down to the 5-10 second range.

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
tonyteate, line 1: With these changes, when the user does a pull-to-refresh, the loading indicator may be hidden while `getProjectBacking()` is in flight, and a Toast message containing any error appears to be delayed. This is due to the code in lines 529–541 related to `refreshProject` and `notifyDelegateToRefreshProject`. These changes likely make that code obsolete, but this will require further testing/QA.

app/src/main/java/com/kickstarter/ui/fragments/BackingFragment.kt
```diff
@@ -271,6 +271,23 @@ class BackingFragment : Fragment() {
             }
             .addToDisposable(disposables)
 
+        this.viewModel.outputs.backingError()
```
tonyteate, line 274: Previously, it seems that we did not show any errors from this VM to the user. Is that because of the overlap with all of the other VMs in `ProjectPageActivity`?

`KSApolloClientV2.getProjectBacking()` has been migrated to throw `KSApolloClientV2Exception`. So if we want to show users error messages, we should be able show _only_ messages from the server (instead of default JVM error messages) by filtering for `KSApolloClientV2Exception.ApiError`.

tonyteate (dismissed):
There is a lot to consider here 😄. The simplest change to get this fix in might be to remove the hardcoded string `"Missing backing and slug"` (to be safe we might just use `Observable.empty()` in that case), and not show errors from `backingError()`. Should get more eyes on perhaps.

jlplks:
> backingError

Removed the hardcoded string, and the time out time reduced. We should have a further discussion with everyone about what's the best way to manage the errors

tonyteate:
> Removed the hardcoded string, and the time out time reduced. We should have a further discussion with everyone about what's the best way to manage the errors

Agree that we should discuss the errors here. Will ping the channel. What do you think about the video from the other [comment](https://github.com/kickstarter/android-oss/pull/2324#discussion_r2096510989)? I'm pretty sure we can just remove the code in lines 529–541 related to `refreshProject` and `notifyDelegateToRefreshProject`. It works as expected but haven't done nor written a thorough test.

app/src/main/java/com/kickstarter/viewmodels/BackingFragmentViewModel.kt
Arkariang: additional qa done by updating a pledge with a new reward or a new payment method, it all seems to work properly

Arkariang (approved):

PR #2330, opened 2025-05-28
MBL-2471: [phase 4 -% raised] definitive strings + active pill text for 

Arkariang:
# 📲 What

- wrapping up phase 4
- Active pill state text
- Definite strings

# 👀 See

https://github.com/user-attachments/assets/b73d8c12-6cbe-4cd3-b982-f070600a3ee2

| --- | --- |
|  |  |

# 📋 QA

- Reset button on first load of percentage raised screen should be disabled.
- If user press "See results" button without selecting any bucket, no error is shown 
- Apply only %raised filter
- Reset
- Apply any combination of the 2 other filter
- Reset

# Story 📖

[MBL-2471](https://kickstarter.atlassian.net/browse/MBL-2471)

[MBL-2471]: https://kickstarter.atlassian.net/browse/MBL-2471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/services/DiscoveryParams.kt
```diff
@@ -560,20 +560,19 @@ class DiscoveryParams private constructor(
     enum class RaisedBuckets {
         BUCKET_0,
         BUCKET_1,
-        BUCKET_2,
-        UNKNOWN;
```
Arkariang, line 564: Was generating a query error if the user pressed "See Results" on percentage raised screen

app/src/main/assets/json/server-config.json
```diff
@@ -19157,6 +19185,6 @@
       ]
     }
   },
-  "country_code": "MX",
+  "country_code": "CA",
```
tonyteate, line 19188: Do we mind changing the country code here?

tonyteate (approved):
Fix + active pill text LGTM. We see translation strings just no translations for percentage raised, so assuming another ticket or just pulled before release 👍

jlplks (approved):

PR #2339, opened 2025-06-02
[DO NOT MERGE] Test out video using Media3 ExoPlayer in Internal Tools menu

ycheng-kickstarter:
# 📲 What

Playing around with Media3 ExoPlayer in our existing native app for mobile visioning.

Currently using videos from https://gist.github.com/jsturgis/3b19447b304616f18657 rather than real Project videos.

# 🤔 Why

# 🛠 How

# 👀 See

https://github.com/user-attachments/assets/52002e20-c46a-4068-935f-cb2328701fe1

# 📋 QA

Access via:
Internal Tools > Mobile Visioning

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2475

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -193,8 +193,16 @@
             android:configChanges="screenSize|orientation"
             android:windowSoftInputMode="adjustPan"
             tools:ignore="LockedOrientationActivity" >
-
         </activity>
+        <activity
+          android:name=".ui.activities.NewProjectPageActivity"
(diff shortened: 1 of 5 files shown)


Arkariang:
Question! on Google IO (attached screenshot), seems we can use full compose and not rely any longer on `AndroidView`, if the objective is to compare the performance level between what can be achieved with different technologies, could we explore any of this topics presented? It could very well mean that what is shown in the video will be available only on Android 16  

Screenshot 2025-06-02 at 3 44 57 PM

Here the video -> https://youtu.be/IaNpcrCSDiI

PR #2343, opened 2025-06-09
MBL-2490 Secret Rewards - limited:The secret reward is not available and the system shows it as available

jlplks:
# 📲 What

Ensure the reward selection CTA button is disabled for unavailable rewards, except for the “No Reward” option which is always enabled.

# 🤔 Why

Previously, some edge cases allowed users to interact with rewards that were marked as unavailable, which led to confusing UX and potential errors during pledge flow. Also, the “No Reward” option could be incorrectly disabled depending on availability status.

# 🛠 How

Updated the ctaButtonEnabled logic to:
- Always enable the button for No Reward .
- Ensure all other rewards are disabled when reward.isAvailable() returns false, regardless of other conditions.
- Reordered the when branches to prioritize No Reward before availability checks.

# 👀 See

[Screen_recording_20250609_132438.webm](https://github.com/user-attachments/assets/ddc299f1-4291-425e-860e-fa7b2dfcfe2b)

# 📋 QA

Open a project in staging that has rewards in different availability states (including a “No Reward”).
- Confirm that:
- Rewards with isAvailable = false have a disabled CTA.
- “No Reward” is always selectable, regardless of its availability status.
- Example deeplink:
https://staging.kickstarter.com/projects/1768690592/plot-secret-rewards?secret_reward_token=d57addc8

# Story 📖

[MBL-2490](https://kickstarter.atlassian.net/browse/MBL-2490)

[MBL-2490]: https://kickstarter.atlassian.net/browse/MBL-2490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/RewardCarouselScreen.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/compose/projectpage/RewardCarouselScreen.kt
@@ -193,13 +193,23 @@ fun RewardCarouselScreen(
 
                     val ctaButtonEnabled = when {
                         RewardUtils.isNoReward(reward) -> true
-                        !reward.hasAddons() && backing?.isBacked(reward) != true -> true
-                        backing?.rewardId() != reward.id() && RewardUtils.isAvailable(
-                            project,
-                            reward
-                        ) && reward.isAvailable() -> true
 
-                        reward.hasAddons() && backing?.rewardId() == reward.id() && (project.isLive || (project.postCampaignPledgingEnabled() ?: false && project.isInPostCampaignPledgingPhase() ?: false)) && reward.isAvailable() -> true
+                        !reward.isAvailable() -> false
+
+                        !reward.hasAddons() && backing?.isBacked(reward) != true -> true
+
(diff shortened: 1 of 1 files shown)


Arkariang (approved):
Potential improvement here for the future, we do need to extract the logic of that button to unit test it as much as possible(potentially to an utils if not wanting to test super deep the compose screen itself). It is quite challenging as it is now to make sure no regression is added when touching anything here 😮‍💨 .

Changes look good to me, manually tested it as well 👍

jlplks:
> Potential improvement here for the future, we do need to extract the logic of that button to unit test it as much as possible(potentially to an utils if not wanting to test super deep the compose screen itself). It is quite challenging as it is now to make sure no regression is added when touching anything here 😮‍💨 .
> 
> Changes look good to me, manually tested it as well 👍

If i have time left on the sprint we could add that as a task to create test for this compose components and have it like a rule from now to create tests for each new component or change on the components

PR #2344, opened 2025-06-09
MBL- 2489 Late pledges Regression for Android 3.32.0

jlplks:
# 📲 What

 Refactored the onContinueClicked method in CheckoutFlowViewModel to support a simplified use case where only a continueCallback is needed. This is now used in ProjectPageActivity when login flow is already guaranteed to be completed.

# 🤔 Why

The original onContinueClicked required both logInCallback and continueCallback, even in cases where the user is already logged in and the login check was redundant. This change adds a simpler, cleaner path for flows where login gating is handled earlier in the UI flow, improving readability and responsibility separation.

# 🛠 How

Implemented the onContinueClicked(continueCallback: () -> Unit) method to:
- Emit the next UI state (currentPage = 4, expanded = true)
- Immediately invoke the provided continueCallback
- Updated ProjectPageActivity to use this simplified version when login is already guaranteed.

# 👀 See

[latepledgefix.webm](https://github.com/user-attachments/assets/30db794a-5dce-44d5-a31c-cd944782b273)

# 📋 QA

1.	Launch the app and navigate to a late pledge project.
	2.	Do the checkout flow
	3.	Select a reward and proceed to add addons.
	4.	Addons selection should then send you the final checkout step
	5.	Verify pledge submission still works correctly.

# Story 📖

[MBL- 2489 Late pledges Regression for Android 3.32.0](https://kickstarter.atlassian.net/browse/MBL-2489)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/ProjectPageActivity.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/ProjectPageActivity.kt
@@ -669,7 +669,10 @@ class ProjectPageActivity :
 
                     ProjectPledgeButtonAndFragmentContainer(
                         expanded = expanded,
-                        onContinueClicked = { checkoutFlowViewModel.onBackThisProjectClicked() },
+                        onContinueClicked = {     checkoutFlowViewModel.onContinueClicked(
+                            logInCallback = { startLoginToutActivity() },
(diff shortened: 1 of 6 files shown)


jlplks:
just need to fix the tets  that  are not  working now

tonyteate (approved):
👍

PR #2353, opened 2025-06-24
MBL-2504: Refactor menu options on manage pledge and added Edit Pledge option

jlplks:
# 📲 What

Refactors the way menu visibility is managed on the Project Page by introducing a centralized ManagePledgeMenuOptions model and helper to determine which options should be shown in the “Manage Pledge” menu.

Also introduces a new Edit pledge option that is conditionally shown based on the ANDROID_PLOT_EDIT_PLEDGE feature flag.

# 🤔 Why

Previously, the logic for determining which items to show in the manage pledge menu was scattered across multiple locations and heavily reliant on inflating different XML menu files depending on backing status. This made the code harder to maintain and extend, especially as new flags and states were introduced for Pledge Over Time (PLOT).

The new Edit pledge menu option needed to be introduced behind a feature flag, and this was a good opportunity to refactor and centralize the logic.
# 🛠 How

Created a ManagePledgeMenuOptions data class that defines booleans for each menu item visibility.
	•	Created a helper function createManagePledgeMenuOptions(Project, FeatureFlagClientType) to encapsulate the logic for determining which menu options are visible.
	•	Updated ProjectPageActivity to always inflate a single manage_pledge menu and dynamically show/hide items based on ManagePledgeMenuOptions.
	•	Replaced XML-based conditional menu inflation with programmatic control.
	•	Introduced a new Edit pledge menu item, visible only when:
	•	The project is live.
	•	The backing is not in PREAUTH status.
	•	The project is Pledge Over Time enabled.
	•	The feature flag ANDROID_PLOT_EDIT_PLEDGE is enabled.
# 👀 See

[editpledge.webm](https://github.com/user-attachments/assets/c4dd0e33-1b12-4825-ad96-433f5739e3d7)

# 📋 QA

1.	Navigate to a project you are backing.
2.	Tap the “Manage Pledge” overflow menu.
3.	Confirm that the options shown (Edit pledge, Change payment, Cancel, etc.) match the expected state based on the project’s backing status and feature flag configuration.
4.	Validate specifically that:
       •	The new Edit pledge option is only shown when ANDROID_PLOT_EDIT_PLEDGE is enabled.
       •	The item is hidden if the feature flag is off, the project is ended, or the backing is in PREAUTH.
# Story 📖

[MBL-2504](https://kickstarter.atlassian.net/browse/MBL-2504)

[MBL-2504]: https://kickstarter.atlassian.net/browse/MBL-2504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Arkariang:
@jlplks reminder, Jira ticket number on PR title

tonyteate:
Cool!

app/src/main/assets/json/server-config.json
```diff
@@ -19185,6 +19311,6 @@
       ]
     }
   },
-  "country_code": "US",
+  "country_code": "MX",
```
tonyteate, line 19314: I think we want this default to be "US"

app/src/main/java/com/kickstarter/ui/activities/ProjectPageActivity.kt
```diff
@@ -739,9 +747,9 @@ class ProjectPageActivity :
                             latePledgeCheckoutViewModel.onAddNewCardClicked(project = projectData.project())
                         },
                         onDisclaimerItemClicked = { disclaimerItem ->
-                            getEnvironment()?.let { environment ->
+                            if (::environment.isInitialized) {
```
tonyteate, line 750: Opinion: Since `environment` is now not-null (even though it it initialized late) I would argue that we should not run this check, especially since `onDisclaimerItemClicked` can certainly only be executed after `onCreate()`

app/src/main/java/com/kickstarter/ui/activities/ProjectPageActivity.kt
```diff
@@ -784,15 +792,16 @@ class ProjectPageActivity :
     }
 
     fun showAccountabilityPage() {
-        getEnvironment()?.webEndpoint()?.let { endpoint ->
+        if (::environment.isInitialized) {
```
tonyteate, line 795: Opinion: Since `environment` is now not-null (even though it it initialized late) I would argue that we should not run this check, especially since `onAccountabilityLinkClicked` can certainly only be executed after `onCreate()`

tonyteate (dismissed):
Lgtm if we can fix `server-config.json`!

app/src/test/java/com/kickstarter/viewmodels/ProjectPageViewModelTest.kt
tonyteate, line 1: We might also move these tests to a different file since they are no longer ViewModel tests?

Arkariang (dismissed):
Amazing job @jlplks

tonyteate (approved):

PR #2362, opened 2025-06-26
MBL-2505 & MBL-2503 Edit pledge flow updateBackingMutation updated.

jlplks:
# 📲 What

Support for the Edit Pledge feature where backers can now switch between Pledge Over Time (PLOT) and paying in full.
This update adds support for the incremental flag in the updateBacking mutation and adjusts the CheckoutScreen UI accordingly.

# 🤔 Why

Backers needed the ability to change their payment plan after pledging — whether switching from PLOT to full payment or vice versa.
To support this, we had to:
 1.	Send the incremental boolean via the updateBacking mutation.
 2.	Hide the payment method UI when a backer is only changing the payment plan, to avoid confusion and prevent altering the stored payment method unintentionally.

# 🛠 How

- Updated the GraphQL schema: UpdateBackingInput now includes incremental: Boolean.
- Hooked up this value in the ViewModel and passed it down to the mutation call.
- Adjusted CheckoutScreen to conditionally hide the payment method section for PledgeReason.UPDATE_PLEDGE.
- Ensured onPledgeCtaClicked lambda accepts the isIncremental param safely.

# 👀 See

[Edit pledge](https://github.com/user-attachments/assets/add760b8-2a6c-4f71-bcba-c2b53218f3d7)

# 📋 QA

1.	Go to a project with PLOT enabled. You can use [this project](https://staging.kickstarter.com/projects/649537918/creator-credit-card-plot-admin-refunds) for example.
	2.	Pledge using Pledge Over Time or Pay in Full.
	3.	Go to Edit Pledge — confirm:
	•	You don’t see the payment method options.
	•	You can switch plans between PLOT and Pay in Full.
	4.	Save changes and ensure mutation includes incremental: true or false depending on selection.
	5.	Check that in the manage pledge change on showing or not the collection plan depending on pledge selection.
	
# Story 📖

[Name of Trello Story](https://kickstarter.atlassian.net/browse/MBL-2505)


app/src/test/java/com/kickstarter/viewmodels/CrowdfundCheckoutViewModelTest.kt
```diff
@@ -359,8 +361,10 @@ class CrowdfundCheckoutViewModelTest : KSRobolectricTestCase() {
         viewModel.userChangedPaymentMethodSelected(cards.first())
         backgroundScope.launch(dispatcher) {
             viewModel.provideScopeAndDispatcher(this, dispatcher)
-            viewModel.pledgeOrUpdatePledge()
-
+            viewModel.pledgeOrUpdatePledge(
```
Arkariang, line 364: @jlplks question here, this test has nothing to do with plot (or seems not configured for it), why is here `isIncremental=true`

app/src/test/java/com/kickstarter/viewmodels/CrowdfundCheckoutViewModelTest.kt
```diff
@@ -788,7 +792,10 @@ class CrowdfundCheckoutViewModelTest : KSRobolectricTestCase() {
             viewModel.provideScopeAndDispatcher(this, dispatcher)
             viewModel.provideBundle(bundle)
             viewModel.userChangedPaymentMethodSelected(cards.last())
-            viewModel.pledgeOrUpdatePledge()
+            viewModel.pledgeOrUpdatePledge(
+                selectedCard = null,
+                isIncremental = true
```
Arkariang, line 797: @jlplks similar question, if the test is `plot is not allowed on project` ? should `isIncremental` be true ?

Arkariang (dismissed):
@jlplks let's take a closed look to the tests 🙏

Arkariang:
@jlplks Another qa step to check, if the project is not plot enabled, when calling update mutation, newly added incremental parameter on `UpdateBackingData` should be null. Or that is what I understand by the parameter being optional on the mutation and these piece over here in KSApolloClient:  
Screenshot 2025-06-26 at 10 56 19 AM

app/src/main/java/com/kickstarter/viewmodels/projectpage/CrowdfundCheckoutViewModel.kt
```diff
@@ -505,7 +548,8 @@ class CrowdfundCheckoutViewModel(val environment: Environment, bundle: Bundle? =
                         pledgeData?.checkoutTotalAmount().toString(),
                         locationId = locationIdOrNull,
                         rwListOrEmpty,
-                        selectedPaymentMethod
+                        selectedPaymentMethod,
```
tonyteate, line 551: Unless we pass `null` for `pData` here, we will never get the correct `UpdateBackingData` for updating `amount`, `location`, `rewards` or `incremental`.

Since `UPDATE_PAYMENT` and `FIX_PLEDGE` are their own flows, it looks like we _can_ explicitly pass `null` here. But this may also be related to L438 - why are we now accepting `selectedCard` and modifying `selectedPaymentMethod` in `pledgeOrUpdatePledge()`?

tonyteate (dismissed):

app/src/main/java/com/kickstarter/services/mutations/UpdateBackingData.kt
```diff
@@ -11,21 +11,35 @@ data class UpdateBackingData(
...
+ * @return An [UpdateBackingData] object configured with the provided parameters.
  */
 fun getUpdateBackingData(
     backing: Backing,
     amount: String? = null,
     locationId: String? = null,
     rewardsList: List<Reward>? = null,
-    pMethod: StoredCard? = null
+    pMethod: StoredCard? = null,
+    incremental: Boolean? = null
```
Arkariang, line 42: 💖

tonyteate (approved):
Looks great! Now on to QA 😄 Nice work @jlplks!

PR #2369, opened 2025-07-07
MBL-2555: Show only toggles UI

Arkariang:
# 📲 What

- Added new row `OthersRow` that contains the UI toggles for `recomended`, `Projects we love`, `saved`, and `following`
- Feature flag logic moved to gate keep now phase 6.

# 🛠 How

# 👀 See
- With feature flag enabled, the new row with all the toggles is visible

https://github.com/user-attachments/assets/0fcbc600-d460-4f68-a715-3f15896a82c3

- With Feature flag disabled, the functionality remains exactly the same as when released 

https://github.com/user-attachments/assets/a61ab76d-3524-44a2-a0f2-6a202fe427c1

- Compose preview, of the new row added

Screenshot 2025-07-07 at 2 52 58 PM

|  |  |

# 📋 QA

- No need to qa anything else yet appart from the ff on/off, that can be done by "hardcoding" the value on `SearchAndFilterActivity:70` or turning on/off on remote config

# Story 📖

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/compose/search/FilterMenuSheet.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/compose/search/FilterMenuSheet.kt
@@ -39,6 +39,8 @@ import com.kickstarter.mock.factories.LocationFactory
 import com.kickstarter.models.Category
 import com.kickstarter.models.Location
 import com.kickstarter.services.DiscoveryParams
+import com.kickstarter.ui.activities.compose.search.FilterMenuTestTags.OTHERS_ROW
+import com.kickstarter.ui.activities.compose.search.FilterMenuTestTags.switchTag
 import com.kickstarter.ui.compose.designsystem.KSDimensions
 import com.kickstarter.ui.compose.designsystem.KSIconButton
 import com.kickstarter.ui.compose.designsystem.KSPillButton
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
```diff
@@ -586,7 +586,7 @@ fun FilterPagerSheet(
                     }
                 },
                 availableFilters = if (shouldShowPhase) FilterType.values().asList()
-                else FilterType.values().asList().filter { it != FilterType.LOCATION && it != FilterType.AMOUNT_RAISED }
+                else FilterType.values().asList().filter { it != FilterType.OTHERS }
```
Arkariang, line 589: @jlplks ack for this point, on your side you will have to do the same for phase 7 goal

app/src/main/java/com/kickstarter/ui/activities/compose/search/FilterMenuSheet.kt
```diff
@@ -157,6 +165,147 @@ fun FilterMenuSheet(
...
+            modifier = Modifier.testTag(OTHERS_ROW)
+        ) {
+            Text(
+                text = text,
+                style = typographyV2.headingLG,
+                color = colors.textPrimary
+            )
+
+            Row(
+                modifier = Modifier.testTag(DiscoveryParams::recommended.name),
```
Arkariang, line 218: Notice here every row has as testtag the name of the field from `DiscoveryParams` it represents

app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
jlplks: Got it, thanks for the context!

jlplks (approved):
LGTM

jlplks (approved):

PR #2372, opened 2025-07-09
MBL-2408 Native onboarding UI

ycheng-kickstarter:
# 📲 What

Just the UI portion of a new onboarding flow with animated transitions and Lottie animations that can be accessed through Internal Tools menu. (Internal Tools menu could be just a temporary place for it if we can think of a better way to easily trigger the flow in the future for QA.)

Main diff to review is **OnboardingFlowScreen.kt** The rest of the files are just changes to the Internal Tools menu and new asset files (they're translated, so there's a lot).

****Navigation logic and business logic around when to actually display the flow coming in another PR.****

# 🤔 Why

We're replacing our braze onboarding flow with a better native one.

# 🛠 How

Used Lottie Animations 
Used AnimatedContent from Compose Animation API: https://developer.android.com/develop/ui/compose/animation/choose-api

# 👀 See

https://github.com/user-attachments/assets/591d961a-ab66-4d2c-b6ae-1ad3c5cb5227

# 📋 QA

Hamburger menu -> Internal tools -> Onboarding flow

Figma specs: https://www.figma.com/design/cCLSxffRzm3jCVkQd0F4fw/Native-Onboarding-Flow?node-id=49-4159&m=dev

Lottie animations should show translated versions if you change device language!

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2408


app/src/main/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreen.kt
```diff
@@ -0,0 +1,248 @@
...
+                        modifier = Modifier.fillMaxWidth(),
+                        type = KSButtonType.BORDERLESS
+                    )
+                }
+            }
+        }
+    }
+}
+
+@OptIn(ExperimentalAnimationApi::class)
```
ycheng-kickstarter, line 180: Used for AnimatedContent to achieve crossfade and slide transitions

Arkariang (approved):
Looks great Yun! 
Just a question, not a request, `OnboardingScreen` does not need any test? Could see as useful small ones like the number of pages for example 🤔  and the the action buttons on each page, but definitely those can come latter on with the real interactions

PR #2380, opened 2025-07-17
MBL-2409 Onboarding flow logic

ycheng-kickstarter:
# 📲 What

New onboarding flow. The flow only shows one time (even if you exit partway), the buttons now do what you expect them to do, and the consent management and notifications permission dialogs will now display as part of the onboarding flow.

# 🤔 Why

# 🛠 How

Exciting technical detail! I had to add in-app default value for the **android_native_onboarding_flow** feature flag in order for the flow to display immediately on the discovery page (otherwise remote config doesn't get initialized by the time I need them). I used a map of values for now, but the proper way might be to use an xml file, which I'll leave for https://kickstarter.atlassian.net/browse/MBL-2649

# 👀 See

If you open the app on a fresh install (and stay logged out) you should be able to see it! Otherwise onboarding flow can still be triggered from the Internal Tools menu.

# 📋 QA

QA consists of 2 parts: when to show the onboarding flow, and exceptions to the onboarding flow

**QA Part 1: QA when to show the onboarding flow** 

Only show the onboarding flow if:
1) the **android_native_onboarding_flow** FF is true
2) the onboarding flow hasn't been shown before
3) and the user is a "new user", using the following new user heuristic: 
user has not seen consent management or notifications permission dialog

If **android_native_onboarding_flow** FF is false, ensure the behavior of old consent management and notifications permission dialogs on the Discovery page does not change.

======

**QA Part 2: QA exceptions to the onboarding flow**

Flow: WELCOME -> SAVE PROJECTS -> NOTIFICATIONS -> ALLOW TRACKING -> SIGNUP/LOGIN
Exceptions to flow:
- If the user is logged in, skip the signup/logic page
- If the user's device is running on Android version <13 (TIRAMISU) we are granted permission by default, so we should skip the notification primer screen

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2409


app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.kt
```diff
@@ -423,6 +416,51 @@ interface DiscoveryViewModel {
...
+                .filter { !sharedPreferences.getBoolean(HAS_SEEN_ONBOARDING, false) }
+                .filter { newUserHeuristic }
+                .subscribe { showOnboardingFlow.onNext(Unit) }
+                .addToDisposable(disposables)
+
+            hasSeenOnboarding
+                .subscribe { sharedPreferences.edit().putBoolean(HAS_SEEN_ONBOARDING, it).apply() }
+                .addToDisposable(disposables)
+
+            // If native onboarding feature switch is false, show the existing alert dialogs
```
ycheng-kickstarter, line 441: Code below is copy pasta'ed from before, with the exception of the onboarding flow FF checks!

app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.kt
```diff
@@ -259,23 +268,7 @@ interface DiscoveryViewModel {
                 .subscribe { messageError.onNext(it) }
                 .addToDisposable(disposables)
 
-            currentUserType.isLoggedIn
```
ycheng-kickstarter, line 262: Moved this block of code controlling consent management dialog and notifications permissions dialog into setupUserOnboarding()

app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.kt
```diff
@@ -423,6 +416,51 @@ interface DiscoveryViewModel {
...
+                .filter { it }
+                .filter { !sharedPreferences.getBoolean(HAS_SEEN_ONBOARDING, false) }
+                .filter { newUserHeuristic }
+                .subscribe { showOnboardingFlow.onNext(Unit) }
+                .addToDisposable(disposables)
+
+            hasSeenOnboarding
+                .subscribe { sharedPreferences.edit().putBoolean(HAS_SEEN_ONBOARDING, it).apply() }
+                .addToDisposable(disposables)
+
```
ycheng-kickstarter, line 440: Code below is the same logic as before, moved from init block into setupUserOnboarding(). Only change I added is to also check that the user hasn't seen onboarding flow.

app/src/main/java/com/kickstarter/libs/featureflag/FeatureFlagClient.kt
```diff
@@ -104,9 +105,13 @@ class FeatureFlagClient(
             minimumFetchIntervalInSeconds = getFetchInterval()
         }
 
-        // - For the MVP no in-app defaults, will add them later on
         remoteConfig?.setConfigSettingsAsync(configSettings)
```
ycheng-kickstarter, line 109: Added this new defaults map for now!

app/src/main/java/com/kickstarter/libs/featureflag/FeatureFlagClient.kt
```diff
@@ -104,9 +105,13 @@ class FeatureFlagClient(
...
 
+        val defaultsMap: MutableMap<String, Any> = mutableMapOf(
```
ycheng-kickstarter, line 110: Added this map for in app defaults for now!

app/src/main/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreen.kt
```diff
@@ -48,60 +56,128 @@ import com.kickstarter.ui.compose.designsystem.KSPrimaryBlackButton
...
             buttonText = stringResource(R.string.Sign_up_or_log_in),
             secondaryButtonText = stringResource(R.string.Explore_the_app),
         )
     )
+    // If user is logged in, skip the last page prompting login/signup
+    val loggedInPages: List<OnboardingPageData> = loggedOutPages.toMutableList().dropLast(1)
+    val pages: List<OnboardingPageData> = if (isUserLoggedIn) loggedInPages.toList() else loggedOutPages.toList()
 
     var currentPage by remember { mutableStateOf(0) }
+    val context = LocalContext.current
```
Arkariang, line 149: super nitpick (no need to change anything), I feel this entire block could stay on the activity, passed a function to the compose screen to be executed. In that way we keep the activity/fragment manager references outside the compose screen.

app/src/main/java/com/kickstarter/libs/featureflag/FeatureFlagClient.kt
Arkariang: FYI: we are thinking on bringing for mobile visioning a SplashScreen, that screen will handle deeplinks and initialize dependencies, if we achieve that we can move how we operate in the loading strategy for Remote Config going from the current [Strategy 3: activate new values on next start_up](https://firebase.google.com/docs/remote-config/loading#strategy_3_load_new_values_for_next_startup)  to [Strategy2: Activate behind loading screen](https://firebase.google.com/docs/remote-config/loading#strategy_2_activate_behind_loading_screen)

Arkariang (approved):
💖 looks amazing!

app/src/main/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreen.kt
ycheng-kickstarter: That's a good point, as that way it also makes it easier to handle analytics in the activity/vm rather than the compose screen. I can make this change as part of the analytics ticket.

app/src/main/java/com/kickstarter/viewmodels/DiscoveryViewModel.kt
```diff
@@ -423,6 +416,49 @@ interface DiscoveryViewModel {
...
+            // Old logic for consent management and notification permissions dialogs
```
ycheng-kickstarter, line 441: No changes below! Just copy pasta'ed from before.

app/src/main/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreen.kt
ycheng-kickstarter: Oh now I remember why I did it like this. It was because I wanted to remain on the given page the entire duration the dialog was being displayed. That means I had to wait for a response from the user before calling currentPage++. I'm not sure how I'd be able to do that with this block moved to the activity.

PR #2381, opened 2025-07-21
MBL-2617: Empty states

Arkariang:
# 📲 What

Search & filter UI empty states polishes

# 👀 See

https://github.com/user-attachments/assets/3263e215-d612-44eb-833a-18e058d158f4

| --- | --- |
|  |  |

# 📋 QA

- Search for a query that will yield no results without applying any filter
- Search for queries and apply filters that will yield no results.

# Story 📖

[MBL-2617](https://kickstarter.atlassian.net/browse/MBL-2617)

[MBL-2617]: https://kickstarter.atlassian.net/browse/MBL-2617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/compose/designsystem/KSButtons.kt
+++ b/app/src/main/java/com/kickstarter/ui/compose/designsystem/KSButtons.kt
@@ -15,6 +15,7 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.wrapContentWidth
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.Button
 import androidx.compose.material.ButtonDefaults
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/kickstarter/ui/compose/designsystem/KSButtons.kt
```diff
@@ -732,10 +733,13 @@ fun KSButton(
     textStyle: TextStyle = typographyV2.body,
     textColor: Color,
     shape: RoundedCornerShape? = null,
+    shouldWrapContentWidth: Boolean = false
 ) {
+    val mod = if (shouldWrapContentWidth)
+        modifier.wrapContentWidth()
+    else modifier.fillMaxWidth()
```
Arkariang, line 740: Quite probably for our design system health we should not set width for buttons within the button itself, but on the modifier outside of the button itself, (fyi: modifiers in compose do follow the builder pattern, if we setup with within the button itself it will be the latest one applied)
Will add a follow up design system refactor ticket for this specific point to not set  `fillMaxWidth()` nor `wrapContentWidth()` within the button component. Adding this solution temporarily to not hold any release

leighdouglas (dismissed):

jlplks (approved):
LGTM!

PR #2383, opened 2025-07-23
MBL-2611: Update PPO query to include list of allow-listed ppo card types, send default list with query

leighdouglas:
# 📲 What

Updating the ppo query to include a list of allowed tier types, based on the v1 and v2 feature flags. 

# 🤔 Why

Backend is updating the PPO query to include a list of allow-listed ppo card types. This list will tell the backend what card types we want included in the response. The purpose of this is to give us more flexibility in releasing cards as they're being built, and allows us to decouple from iOS as development parity is inconsistent. 

# 🛠 How

Query is backwards compatible with older versions - if the list is null, the backend defaults to tier 1 card types and pledge management. 
Otherwise, we filter card types out of the list based on feature flags. 

# 👀 See

Nothing to really see, we just want to make sure if the feature flags are on or off, that the appropriate cards are being shown in the list. 

# 📋 QA

Tier 1: 
        PledgeTierType.FAILED_PAYMENT,
        PledgeTierType.ADDRESS_LOCK,
        PledgeTierType.SURVEY_OPEN,
        PledgeTierType.PAYMENT_AUTHENTICATION 

Tier 2:
        PledgeTierType.ADDRESS_CONFIRMED,
        PledgeTierType.PLEDGE_COLLECTED,
        PledgeTierType.SURVEY_SUBMITTED,
        PledgeTierType.AWAITING_REWARD,
        PledgeTierType.PLEDGE_MANAGEMENT,
        PledgeTierType.REWARD_RECEIVED

Turn off `ANDROID_PLEDGED_PROJECTS_OVERVIEW_V2` and observe that only the tier 1 cards populate

# Story 📖

[MBL-2611: Update PPO query to include list of allow-listed ppo card types, send default list with query](https://kickstarter.atlassian.net/browse/MBL-2611)


app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/viewmodel/PledgedProjectsOverviewViewModel.kt
```diff
@@ -245,6 +257,21 @@ class PledgedProjectsOverviewViewModel(
         )
     }
 
+    private fun getAllowedTierTypes(): List<PledgeTierType> {
+        val list = PledgeTierType.values().toMutableList()
+        list.remove(PledgeTierType.REWARD_RECEIVED)
+
+        isV1Enabled.isFalse().let {
```
tonyteate, line 264: Sorry if I'm misunderstanding the intent here. It seems like we want to remove all Tier 1 types from the list if the v1 ff is off, but keep them if it's on. Since `isFalse()` will not return `null`, the `let` here will run whether the feature flag is on or off. Similarly for the v2 ff. If this is the case, it might be better to use a regular `if` statement for both.

tonyteate (dismissed):
Lgtm!

tonyteate (approved):

PR #2398, opened 2025-07-31
Imartin/mbl 2691 polish UI empty states

Arkariang:
# 📲 What

- when translations received key to substitute variable missed.
- aligned text

# 🤔 Why

- more polished UI

# 👀 See
| Before 🐛 | 

ko

|After 🦋 |

ok

| --- | --- |
|  |  |

# 📋 QA

- Type something that will yield no results, then apply a filter

# Story 📖

[MBL-2691](https://kickstarter.atlassian.net/browse/MBL-2691)


jlplks (approved):
LGTM ! 👍

PR #2411, opened 2025-08-26
MBL-2745: Setup remote config initialization functions in Deeplink viewmodel

leighdouglas:
# 📲 What

Setup the deeplink view model to have functions that start initialization process for remote config. Also, modify the existing `fetchAndActivate` function on the feature flag client to be a suspend function. 

# 🤔 Why

Deeplink viewmodel will be the launching activity and handle all initializations. 

# 🛠 How

Firebase provides a function `fetchAndActivate` that we can call that will handle both the fetching and activation operations for us. In turn, we can convert the `fetchAndActivate` function in our `FeatureFlagClient` wrapper class to a suspend function and call `await`, which will return a boolean value for us if the call succeeds for fails.

# 👀 See

no user facing changes

# 📋 QA

nothing to QA yet. however i did confirm that these functions are all working:
Screenshot 2025-08-26 at 12 57 12

# Story 📖

[MBL-2745: Setup remote config initialization functions in Deeplink viewmodel](https://kickstarter.atlassian.net/browse/MBL-2745)


Arkariang (approved):
lgtm! 👍

tonyteate (approved):
🎉

PR #2412, opened 2025-08-27
no-jira: PPO timeout errors, paging crash, and total alerts bug for hotfix

leighdouglas:
# 📲 What

1. The PPO query was taking a long time, and would timeout after 10 seconds, causing the page to show an error. 
2. [Firebase showing a crash for the paging 3 library](https://console.firebase.google.com/u/1/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/5526eaecfe0ff800288b2cd9275387a8?time=last-seven-days&types=crash&sessionEventKey=68AF33BA02150001485D4A17B253CE50_2121752419153019001)
3. Total alerts in the header was displaying the incorrect number

# 🛠 How

1. Changed the ppo page size from 25 to 10, making the call run much faster (<half the time)
2. Added the `cachedIn(viewModelScope)` to the pledged projects overview paging flow
3. Changed the value we display from `totalAlerts` on the ppo query, to `backingActionsCount` on the user object. hide count when user has no tier 1 alerts

# 👀 See

1 Alert:
Screenshot_20250827_133418

No Alerts: 
Screenshot_20250827_130752

# 📋 QA

1. Open backings dashboard. Paginate, pull to refresh, reload the page, scroll down a few pages. Make sure there are no errors or timeouts
2. Couldnt reproduce the crash, but if you could, then please make sure the crash is no longer happening
3. Make sure the "Alerts()" displayed at the top of the screen is reflective of all tier 1 alerts

# Story 📖

no-jira

### DIFF
--- a/app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/pledgedprojectsoverview/ui/PledgedProjectsOverviewActivity.kt
@@ -80,7 +80,7 @@ class PledgedProjectsOverviewActivity : AppCompatActivity() {
                 val ppoUIState by viewModel.ppoUIState.collectAsStateWithLifecycle()
 
                 val lazyListState = rememberLazyListState()
-                val totalAlerts = viewModel.totalAlertsState.collectAsStateWithLifecycle().value
+                val totalAlerts = ppoUIState.totalAlerts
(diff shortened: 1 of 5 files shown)


Arkariang (approved):
Not getting the something went wrong message any longer :) 

https://github.com/user-attachments/assets/681bfe7b-edee-44a0-9060-6bf84dfae163

PR #2413, opened 2025-08-27
MBL-2699: V1 Rest API using coroutines

Arkariang:
# 📲 What

In preparation for mobile visioning

# 🤔 Why

- New networking client for the old Rest API using Retrofit, now with suspend functions integration instead of RXJava.
For now it has only 1 active endpoint to obtain `Configuration`, but serves as example moving forward.
- Added error handling keeping still `ErrorEnvelope` -> `ApiException` formats.
- Added tests with `mockwebserver`, potentially we could create new Exceptions formats - error reporting based on http code errors if we see need for it.

# 👀 See
- Added in a random VM a call to this endpoint

networkInspector

| --- | --- |
|  |  |

# 📋 QA
- You can get this snippet code into any VM that is already using coroutines:
```
 val config: Config? = environment.apiClient()?.getConfig()?.let { response ->
                if (response.isSuccess)
                    return@let response.getOrNull()
                else null
            }
```
Then debug, or use the network profiled to double check the call is successful.

- As for the error handling take a look at the new test class.

# Story 📖

[MBL-2699](https://kickstarter.atlassian.net/browse/MBL-2699)

[MBL-2699]: https://kickstarter.atlassian.net/browse/MBL-2699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ycheng-kickstarter (approved):
Great start!

app/src/main/java/com/kickstarter/services/ApiClient.kt
```diff
@@ -0,0 +1,72 @@
...
+import com.kickstarter.libs.Config
+import com.kickstarter.services.apiresponses.ErrorEnvelope
+import retrofit2.Response
+import retrofit2.http.GET
+import java.io.EOFException
+import kotlin.coroutines.cancellation.CancellationException
+
+interface ApiService {
+    @GET("/v1/app/android/config")
+    suspend fun getConfigResponse(): Response<Config>
```
Arkariang, line 14: @leighdouglas let me know if any other endpoint is required for the SplashScreen initialization project and we can add it right away

tonyteate:
🔥

PR #2414, opened 2025-09-02
MBL-2746: Create logic for when remote config init completes, then call initialization function from activity

leighdouglas:
# 📲 What

We want to call the init process from the deeplink activity, and when we finish fetching from remote config, we want to process the intent that carries the deeplink. 

# 🤔 Why

Unified splash screen

# 🛠 How

If initialization of remote config is successful, we call a function that essentially processes the intent and opens the deeplink. In subsequent PRs, when we are launching the deeplink activity as the main activity, we will also port the user to the discovery screen in the case that it is not a deeplink.

# 👀 See

https://github.com/user-attachments/assets/9f9932e5-3b01-4fa0-a8cb-f9508a031a12

Screenshot 2025-09-02 at 19 47 39

# 📋 QA

This only works for deeplinks. next pr will make this activity the launching activity, and will port user to discovery upon app launch

Open a deeplink and verify that it opens the correct page. Can also verify in logcat that the fetch to remote config worked

# Story 📖

[MBL-2746: Create logic for when remote config init completes, then call initialization function from activity](https://kickstarter.atlassian.net/browse/MBL-2746)


app/src/main/java/com/kickstarter/libs/FirebaseHelper.kt
```diff
@@ -27,21 +33,22 @@ class FirebaseHelper(context: Context, ffClient: FeatureFlagClientType, callback
...
 
     init {
-        if (context.isKSApplication()) {
-            if (FirebaseApp.getApps(context).isEmpty()) {
-                FirebaseApp.initializeApp(context)
-                FirebaseCrashlytics.getInstance().setCrashlyticsCollectionEnabled(true)
-                FirebaseAnalytics.getInstance(context).setAnalyticsCollectionEnabled(true)
-            }
+        // not sure if runBlocking is the correct coroutine launcher we want here, but it seems to work on normal app bootstrap
+        runBlocking {
```
Arkariang, line 37: @leighdouglas what happens if you try to start the app without internet? 
Fwi: if you want an alternative to `runBlocking` you can use from `KSAplication` application scope, you just need to send it to FirebaseHelper.initialize
Screenshot 2025-09-03 at 12 47 45 PM

Screenshot 2025-09-03 at 12 48 31 PM

app/src/main/java/com/kickstarter/libs/featureflag/FeatureFlagClient.kt
```diff
@@ -137,7 +137,7 @@ class FeatureFlagClient(
     }
 
     override suspend fun fetchAndActivate(): Boolean {
-        val isInitialized = remoteConfig?.fetchAndActivate()?.await().isTrue()
+        val isInitialized = remoteConfig?.fetchAndActivate()?.await().isNotNull()
 
         log("${this.javaClass} fetchAndActivate completed: $isInitialized")
```
Arkariang, line 142: nitpick: We can change this log and variable name really, as we discover the other day taking a closer look to the documentation, for `fectchAndActivate`will only be true in those instances where there has been any configuration change applied, for example ff value has changed or new feature flag added.

Screenshot 2025-09-03 at 12 57 39 PM

Arkariang (approved):
lgtm, left some comments that can addressed in follow up tickets if required follow up

app/src/main/java/com/kickstarter/ui/activities/DeepLinkActivity.kt
```diff
@@ -65,6 +71,16 @@ class DeepLinkActivity : AppCompatActivity() {
             }
         }
 
+        lifecycleScope.launch {
+            lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
+                FirebaseHelper.identifier
+                    .filter { !it.isBlank() }
+                    .collect {
+                        viewModel.runInitializations()
```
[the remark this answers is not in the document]
leighdouglas: Okay, i agree that the this logic should be moved to the viewmodel. i moved this here because testing was becoming difficult with this logic in the VM, but i have a path forward with this

app/src/main/java/com/kickstarter/libs/FirebaseHelper.kt
leighdouglas: I've reverted some of these changes to FirebaseHelper. 
I will have a better idea of what happens when the app starts without internet once I make the deeplink activity the launching activity. The plan is to emulate the behavior that we have right now and bring them into the app anyways. We can address no internet connection in the future

tonyteate (approved):
🎉

PR #2417, opened 2025-09-04
MBL-2727 Refactor onboarding flow tests to avoid calling Thread.sleep()

ycheng-kickstarter:
# 📲 What

Refactor the onboarding flow logic and tests to use an injectable Clock, with a FakeClock implementation that can be manually advanced in time so that we don't have to call Thread.sleep() in tests.

# 🤔 Why

The button debounce in the onboarding flow was implemented using the real clock System.currentTimeMillis(), which required us to call Thread.sleep() to pass real time. But Thread.sleep() slows down our tests because it uses real time and can sometimes cause flakey behavior. 

# 🛠 How

Create an injectable clock interface with two implementations - SystemClock and FakeClock.
 - The real SystemClock is used in code to avoid hardcoding System.currentTimeMillis().
 - The FakeClock is injected in tests and can be used to advance the time manually using the advanceBy() method. Use this FakeClock instead of calling Thread.sleep() which slows tests down.

# 👀 See

No user facing changes.

# 📋 QA

Tests in OnboardingFlowScreenTest.kt should continue to pass as they did before.

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2727

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreen.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreen.kt
@@ -234,7 +234,6 @@ fun OnboardingScreen(
                 KSPrimaryBlackButton(
                     text = filteredPages[currentPage].buttonText,
                     onClickAction = {
-
                         val currentTime = clock.now()
                         if (currentTime - lastClickTime > debounceInterval) {
                             lastClickTime = currentTime
--- a/app/src/main/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreen.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreen.kt
@@ -52,9 +52,11 @@ import com.airbnb.lottie.compose.animateLottieCompositionAsState
 import com.airbnb.lottie.compose.rememberLottieComposition
 import com.kickstarter.R
 import com.kickstarter.libs.AnalyticEvents
+import com.kickstarter.libs.utils.Clock
 import com.kickstarter.libs.utils.EventContextValues.ContextSectionName.ACTIVITY_TRACKING_PROMPT
 import com.kickstarter.libs.utils.EventContextValues.CtaContextName.ALLOW
 import com.kickstarter.libs.utils.EventContextValues.CtaContextName.DENY
+import com.kickstarter.libs.utils.SystemClock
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreen.kt
```diff
@@ -231,7 +234,8 @@ fun OnboardingScreen(
                 KSPrimaryBlackButton(
                     text = filteredPages[currentPage].buttonText,
                     onClickAction = {
-                        val currentTime = System.currentTimeMillis()
+
+                        val currentTime = clock.now()
```
ycheng-kickstarter, line 237: clock.now() is just a wrapper for System.currentTimeMillis(), but having the clock injected this way allows us to then pass in a fake clock for tests

app/src/test/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreenTest.kt
```diff
@@ -121,9 +125,9 @@ class OnboardingFlowScreenTest : KSRobolectricTestCase() {
         setupOnboardingScreen()
 
         primaryButton.performClick() // Welcome -> Save
-        Thread.sleep(500)
+        fakeClock.advanceBy(600)
```
ycheng-kickstarter, line 128: The debounce interval is 500 milliseconds so we advance time by 600 milliseconds

app/src/test/java/com/kickstarter/ui/activities/compose/OnboardingFlowScreenTest.kt
```diff
@@ -216,15 +220,14 @@ class OnboardingFlowScreenTest : KSRobolectricTestCase() {
     fun `Test activity tracking page secondary button click navigates to Discovery page if user is already logged in`() {
         setupOnboardingScreen(isUserLoggedIn = true)
 
-        // Navigate to Notifications Page (Welcome -> Save -> Notifications)
         primaryButton.performClick() // Welcome -> Save
-        Thread.sleep(500)
-        primaryButton.performClick() // Save -> Notifications
```
ycheng-kickstarter, line 222: Fixed some erroneous comments

Arkariang (approved):
Amazing!!! You've made my day with this addtion yun! 💖

PR #2418, opened 2025-09-10
MBL-2747: Make deeplink activity launching activity, install the splash screen, create logic for splash screen dismissal

leighdouglas:
# 📲 What

We are at the point where the splash screen is ready to be the launching activity, replacing discovery as the first entry point into the app. In this PR, we handle updating the manifest to reflect this new launching activity, handle processing the intent if it is not a deeplink (bringing the user into the discovery screen), and keeping track of processes to know when to dismiss the splash screen. 

# 🤔 Why

# 🛠 How

- Run initialization is called from deeplink activity
- We wait for processes to finish, then process the intent that opened the app
- if the intent is launching, we take the user to the discovery screen, if not, we bring them to wherever the deeplink is pointing
- Currently, if there is an exception caused by firebase remote config for whatever reason, we still bring the user into the discovery screen (as we would do today) -> in the future we will want to handle this more robustly.

# 👀 See

Screenshot 2025-09-10 at 12 38 34

# 📋 QA

- Test that feature flags are downloaded correctly before loading into discovery by turning the feature flag on or off in staging, and confirming via logs that the correct value is being loaded upon app launch
- test that deeplinks are opening the app correctly
- test that going offline doesnt crash the app, and still brings the user into the discovery screen

# Story 📖

[MBL-2747: Make deeplink activity launching activity, install the splash screen, create logic for splash screen dismissal](https://kickstarter.atlassian.net/browse/MBL-2747)


app/src/main/java/com/kickstarter/viewmodels/DeepLinkViewModel.kt
```diff
@@ -116,35 +116,50 @@ interface DeepLinkViewModel {
...
+                            initializationsProcessing = false
+                            processIntent(externalCall = externalCall)
+                        }
                     }
             }
         }
 
         private fun processIntent(intent: Observable<Intent> = intent(), externalCall: CustomNetworkClient) {
+            intent()
+                .filter { it.action == Intent.ACTION_MAIN || it.categories.contains(Intent.CATEGORY_LAUNCHER) }
```
tonyteate, line 149: Let's filter for the combination (`&&`) of `ACTION_MAIN` and `CATEGORY_LAUNCHER`, that will specifically distinguish that the app was opened from a launcher.

app/src/main/java/com/kickstarter/ui/activities/DeepLinkActivity.kt
```diff
@@ -41,30 +37,21 @@ class DeepLinkActivity : AppCompatActivity() {
...
             }
         }
 
+        installSplashScreen().setKeepOnScreenCondition {
+            // replace with consuming vm state in subsequent work -> if (state == SplashState.Finished) false ?: true
+            viewModel.initializationsProcessing
+        }
+
+        super.onCreate(savedInstanceState)
+        setUpConnectivityStatusCheck(lifecycle)
```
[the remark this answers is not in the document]
leighdouglas: The plan is to have the internet connection check to be apart of the initialization logic - in fact, i think it is the first thing that we should check before running the other processes, and throw an exception/show some ui to the user letting them know that they dont have internet connect. However, i dont think we necessarily need to do this right now, as we havent determined what the recourse is if the device doesnt have internet connection, and we dont have any ui to show them either.
If you see some benefit adding this logic sooner than later before , then i am more than happy to include that in a subsequent PR or even this one, but i think it's out of scope for this ticket to be frank.

app/src/main/java/com/kickstarter/viewmodels/DeepLinkViewModel.kt
```diff
@@ -275,13 +297,17 @@ interface DeepLinkViewModel {
                 }.addToDisposable(disposables)
 
             uriFromIntent
-                .filter { it.isSettingsUrl() }
```
tonyteate, line 278: Did we want the null-check here?

tonyteate (approved):
🎉

PR #2419, opened 2025-09-16
MBL-2748: Create splash screen ui/nav states and update ui accordingly when ui state is Finished

leighdouglas:
# 📲 What

# 🤔 Why

Now that the deeplink activity is the launching activity, we want to update to consume state flows instead of observables. We will migrate the deeplinks to coroutines next

# 🛠 How

Created ui state and navigation target state classes for deeplink viewmodel
Replaced the boolean tracking when the initializations have finished processing with state flow updates
Consume those updates in the activity, dismiss the splash screen accordingly

# 👀 See

No userfacing changes

# 📋 QA

QA launching the kickstarter app from scracth, as well as deeplinks -> make sure the splash screen dismisses properly and that they are ported to the correct place in the app
Test that the app still ports the user into the discovery screen when in airplane mode
For all scenarios, make sure `fetchAndActivate` is being called via logcat

# Story 📖

[MBL-2748: Create splash screen ui/nav states and update ui accordingly when ui state is Finished](https://kickstarter.atlassian.net/browse/MBL-2748)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -231,7 +231,7 @@
             android:exported="false"
             android:theme="@style/KSTheme" />
 
-        <activity android:name=".ui.activities.DeepLinkActivity"
+        <activity android:name=".ui.activities.SplashScreenActivity"
             android:theme="@style/SplashTheme"
             android:label="@string/app_name"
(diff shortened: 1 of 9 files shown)


tonyteate (approved):
🔥

PR #2428, opened 2025-10-01
MBL-2833 Parse pledge redemption messages in MessageService for push notifs

ycheng-kickstarter:
# 📲 What

In MessageService.kt, add logic to parse messages related to Pledge Redemption.

# 🤔 Why

We forgot to actually add logic to parse pledge redemption messages in MessageService when we first worked on PM push notifs in https://github.com/kickstarter/android-oss/pull/2172 ...so users haven't been able to receive PM push notifs

# 🛠 How

Spent some time with @a-lavis and we determined the backend sends us a payload like this, so we'd need to parse the `order` json block:
```
{
  "GCM": {
    "data": {
      "gcm": {
        "title": "Finalize your pledge",
        "alert": "Response needed! Get your reward for backing India Street Lettering: A Book."
      },
      "type": "complete_pledge_redemption",
      "ref_tag": "android_complete_pledge_redemption_push",
      "order": {
        "id": 327151,
        "backing_id": 186055186,
        "project_id": 1464449823,
        "pledge_manager_path": "/projects/blaft/india-street-lettering-a-book/backing/redeem"
      }
    }
  },
  "default": "Response needed! Get your reward for backing India Street Lettering: A Book."
}
```

# 👀 See

# 📋 QA

# Story 📖

https://kickstarter.atlassian.net/browse/MBL-2833


app/src/main/java/com/kickstarter/services/firebase/MessageService.kt
```diff
@@ -115,6 +115,11 @@ class MessageService : FirebaseMessagingService() {
                     )
                 )
                 .survey(gson.fromJson(data["survey"], Survey::class.java))
+                .pledgeRedemption(
+                    gson.fromJson(
+                        data["order"], PushNotificationEnvelope.PledgeRedemption::class.java
```
ycheng-kickstarter, line 120: See https://github.com/kickstarter/android-oss/blob/master/app/src/main/java/com/kickstarter/services/apiresponses/PushNotificationEnvelope.kt#L274

Arkariang: @ycheng-kickstarter any chance we can add as reference the ruby controller counterpart? I really wanna double check the data structure we should receive, I've been out of the loop for PM pushes for quite a long time so catching up :).

Arkariang (approved):
Confirmed with @ycheng-kickstarter there will be QA session alongside @a-lavis (team-b) in Prod.

PR #2496, opened 2026-04-08
DISC-143: Discovery Entry point card behind a feature gate + StatsigClient initialization improvements

Arkariang:
# 📲 What

This PR adds a new promotional banner to the Discovery screen that serves as an entry point for the new Video Feed feature, a Statsig feature gate will enable/disable access to users to this feature.
Key UI/UX additions:
- Discovery Banner: Integrated a new banner into the Discovery feed using `DiscoveryVideoFeedBannerViewHolder`.
- Design System: Added KSBanners.kt to our Compose design system for standardized banner components.

⚠️  **Statsig Integration updates:** Refactored StatsigClient to safely handle feature gate and experiment checks after finding race conditions on fresh app install. The check of the gate or experiment needs to happen once the StatsigSDK has completed initialization or will return false.

📥 The banner is visible on top of the `Sign Up or LogIn card` for **logged out** users (see video on #see section). For **Logged In** users it appears on top of the `Since you've been gone` card (attached screenshot), or alone on top of the projects shown (attached screenshot).

Screenshot 2026-04-08 at 3 48 15 PM

Screenshot 2026-04-08 at 3 45 03 PM

# 🤔 Why
The primary goal is to surface the Video Feed feature to users through the Discovery screen by adding a banner behind a feature gate, while doing so we had to upgrade StatsigClient integration.

The changes made for `StatsigClient` and `StatsigClientTest` address race conditions that occurred when checking a gate while the StatsigSDK initialization was not yet completed. By providing an StateFlow `isReady`, we ensure that feature flags are only evaluated once the client is fully prepared.

Worth mentioning, in the future we aim to move Statsig initialization call from `KSApplication` to `SplashScreenViewModel.runInitializations()`. This work is yet to be planned/scheduled alongside Segment/Braze SDK's initialization once SplashScreen polishing happens to further streamline the startup process.

# 🛠 How

**Discovery Feed:** Updated DiscoveryFragment and its adapter to include the new DiscoveryVideoFeedBannerViewHolder which renders the entry point banner.

**StatsigClient Refactor**:
- Added an isReady StateFlow to signal when initialization is complete.
- Updated initialize to handle errors and success states more explicitly.

**Testing Infrastructure**: 
* Enhanced StatsigClientTest to verify that gates are checked correctly relative to the initialization state.
* Created MockStatsigClient, a specialized mock object for use in ViewModel unit tests. It allows developers to provide a predefined gateMap and starts in a "ready" state immediately.

# 👀 See

https://github.com/user-attachments/assets/2bcc8e50-652d-47af-9107-273b3fde38d9

|  |  |

# 📋 QA

Both Stating and Production Statsig Dashboards have a feature gate targeting android version "3.39.0":
- Modify the file 'internal/external'_version_name.txt to see the banner and access VideoFeedActivity.
- - - Make sure on fresh install you see the banner without having to kill the app and opening it again, changes made on `StatsigClient` do account for this specific scenario.
- Current files 'internal/external'_version_name.txt are set to 3.38.0 so won't pass the flag check.

# Story 📖

[DISC-140](https://kickstarter.atlassian.net/browse/DISC-143)

[DISC-140]: https://kickstarter.atlassian.net/browse/DISC-143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -98,28 +113,4 @@ open class StatsigClient(
             }
         }
     }
-
-    fun updateExperimentUser(dispatcher: CoroutineDispatcher = Dispatchers.IO, errorCallback: (Exception) -> Unit = {}) {
```
Arkariang, line 102: ⚠️ Removed as this piece was tied to the assumption back in 2025 the client will need to update the statsig userID, apparently Segment -> Statsig source + some action still unkown on the backend will handle that. 

`currentUser` observable is still a part of the constructor + DI as there is a chance we might either way need to do something about the userID, waiting for more information in this part.

app/src/main/java/com/kickstarter/ui/activities/DiscoveryActivity.kt
```diff
@@ -281,8 +281,6 @@ class DiscoveryActivity : AppCompatActivity(), SharedPreferences.OnSharedPrefere
             .compose(Transformers.observeForUIV2())
             .subscribe { this@DiscoveryActivity.showErrorSnackBar(binding.discoveryAnchorView, it ?: "") }
             .addToDisposable(disposables)
-
-        statsigClient.updateExperimentUser()
```
Arkariang, line 285: This experiment was set up back in february 2025, stale since then no longer needed, as the project was kicked from the roadmap.

app/src/main/java/com/kickstarter/ui/compose/designsystem/KSColors.kt
Arkariang, line 1: As mentioned on documentation on both data classes, runned into a JVM limitation of 255 parameters when adding new semantic color names for the banner. Refactored VideoFeed + Banner into data classes to overpass the limitation

app/src/main/res/drawable/entrypoint_banner.png
Arkariang, line 1: Figma SVGs, they'll come as:

```
<image> tags (embedded bitmaps), or
url(#pattern...) fills
```

Android vector drawables don't support either of those — they only support paths, solid colors, and gradients. So the separate SVG with embedded photos would give AAPT error. For now using this PNG, will explore alternative solutions on follow up non-blocking tickets.

app/src/test/java/com/kickstarter/libs/MockStatsigClient.kt
```diff
@@ -0,0 +1,61 @@
...
+    init {
+        _isReady.value = true
+    }
+
+    override fun isInitialized(): Boolean = true
+
+    override fun checkGate(gateName: String): Boolean =
+        gateMap[gateName] ?: false
+
+    override fun getFeatureGate(gateName: String): FeatureGate =
```
Arkariang, line 53: @tonyteate notice that the mock object does not override as of now`getExperiment` you will likely wanna add additional private val `experimentMap: Map<String, Boolean> = emptyMap()` to the constructor and override the method when doing the experiment on Checkout.

tonyteate (approved):
Amazing 💯 🙏

PR #2500, opened 2026-04-14
PF-121: Upgrade Compose testing libraries + swap in createComposeRule() v2

tonyteate:
# 📲 What

Upgrade the Compose libraries for local UI tests to `1.11.0-rc01`, overriding the version provided by the BOM.

# 🤔 Why

The v1 versions of the core Compose testing APIs ([createComposeRule](https://developer.android.com/reference/kotlin/androidx/compose/ui/test/junit4/v2/package-summary#createComposeRule(kotlin.coroutines.CoroutineContext)), [createAndroidComposeRule](https://developer.android.com/reference/kotlin/androidx/compose/ui/test/junit4/v2/package-summary#createAndroidComposeRule(java.lang.Class,kotlin.coroutines.CoroutineContext)), [runComposeUiTest](https://developer.android.com/reference/kotlin/androidx/compose/ui/test/v2/package-summary#runComposeUiTest%28kotlin.coroutines.CoroutineContext,kotlin.coroutines.CoroutineContext,kotlin.time.Duration,kotlin.coroutines.SuspendFunction1%29), [runAndroidComposeUiTest](https://developer.android.com/reference/kotlin/androidx/compose/ui/test/v2/package-summary#runAndroidComposeUiTest%28kotlin.coroutines.CoroutineContext,kotlin.coroutines.CoroutineContext,kotlin.time.Duration,kotlin.coroutines.SuspendFunction1%29), etc) are deprecated.

The v2 APIs, which improve control over coroutine execution, are now past beta.

From the documentation:

> [I]t's strongly recommended to migrate to the new APIs. Migrating verifies your tests align with standard coroutine behavior and avoids future compatibility issues.

Reference: [Migrate to v2 testing APIs  |  Jetpack Compose  |  Android Developers](https://developer.android.com/develop/ui/compose/testing/migrate-v2) (Please read)

# 🛠 How

The v1 APIs remain in the same packages, so upgrading these libraries does not automatically upgrade usages. The v2 APIs are generally available under a `.v2` sub-package. 

Virtually all local UI tests rely on the `composeTestRule` property of the abstract class `KSRobolectricTestCase`, defined by a call to `createComposeRule()`. We take advantage of this by simply updating the import declaration in that file to bring `createComposeRule` from `.v2`, thus upgrading all of the tests at the same time.

In addition, we leverage the new version of `runComposeUiTest` to test `AsyncImagePainter` with a custom `ImageLoader`, but avoid the "two clock problem" introduced by using the v1 `createComposeRule` with `runTest` (as [recommended](https://developer.android.com/develop/ui/compose/testing/migrate-v2#previous-approach-2)).

# 👀 See

N/A

# 📋 QA

(CircleCI checks pass)

# Story 📖

[\[PF-121\] Upgrade Compose testing libraries & use createComposeRule() v2 - Jira](https://kickstarter.atlassian.net/browse/PF-121)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -212,7 +212,7 @@ dependencies {
     /* Compose BOM */
     Dependency composeBom = platform('androidx.compose:compose-bom:2026.03.01')
     // BOM version override for `androidx.compose.ui:ui-test-junit4` and `androidx.compose.ui:ui-test-manifest`
-    def androidxComposeUiTestVersion = "1.11.0-beta02"
+    def androidxComposeUiTestVersion = "1.11.0-rc01"
     implementation composeBom
     testImplementation composeBom
     androidTestImplementation composeBom
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -198,12 +198,6 @@ repositories {
 dependencies {
     implementation fileTree(include: ['*.jar'], dir: 'libs')
 
-    // Compose BOM
-    Dependency composeBom = platform('androidx.compose:compose-bom:2026.03.01')
-    implementation composeBom
-    testImplementation composeBom
-    androidTestImplementation composeBom
-
     implementation 'androidx.core:core-ktx:1.16.0'
     implementation 'androidx.annotation:annotation:1.9.1'
     implementation 'androidx.appcompat:appcompat:1.7.1'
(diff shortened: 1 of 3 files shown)


app/src/test/java/com/kickstarter/features/projecstory/ui/ProjectStoryCaptionedImageTest.kt
tonyteate, line 1: Happy to delete the redundant tests in a later PR, but I do think it's notable and educational that they are both correct in their own approaches. Despite the "two clock problem" technically existing in `` `test image loading with caption (mainClock)` ``, it never attempts to manually advance Compose's virtual clock, and instead relies on the normal synchronization triggered by assertions.

app/build.gradle
```diff
@@ -215,6 +209,14 @@ dependencies {
     // Chrome Tabs library
     implementation 'androidx.browser:browser:1.8.0'
 
+    /* Compose BOM */
+    Dependency composeBom = platform('androidx.compose:compose-bom:2026.03.01')
+    // BOM version override for `androidx.compose.ui:ui-test-junit4` and `androidx.compose.ui:ui-test-manifest`
+    def androidxComposeUiTestVersion = "1.11.0-beta02"
```
Arkariang: `[1.11.0-rc01](https://mvnrepository.com/artifact/androidx.compose.ui/ui-test-junit4/1.11.0-rc01)` was released last week have you tried it?. If there is already release candidates likely will be an official release soon 🤞and hopefully the override of the BOM version is just temporal

Arkariang (dismissed):

app/build.gradle
tonyteate: Great timing! Going to bump it.

Arkariang (approved):

PR #2503, opened 2026-04-20
use ssh git endpoint for remotes

warrenwegs:
# 📲 What

Uses SSH endpoint for makefile.

# 🤔 Why

Our circle CI pipeline uses git ssh keys for auth rather than name and password


Arkariang (approved):
lgtm! thanks @warrenwegs !!

PR #2504, opened 2026-04-27
DISC-199: Android 3.39.0 regression testing crashes

Arkariang:
# 📲 What

Fix two fatal crashes affecting `SearchAndFilterActivity` and `ReportProjectActivity`.

# 🤔 Why

Both crashes were triggered by unsafe assumptions at ViewModel initialization time:

`SearchAndFilterViewModel` called `statsigClient.checkGate()` synchronously in init before the Statsig SDK had finished its async initialization, causing an IllegalStateException.
`ReportProjectViewModel` passed intent.extras (which can be null) directly into Observable.just(), which rejects null items immediately with a NullPointerException — before any null-filtering operators had a chance to run.

# 🛠 How

**Crash 1 — Statsig not initialized**:

* Moved the checkGate call out of the synchronous init and into a coroutine that collects `statsigClient.isReady: StateFlow<Boolean>`. The banner value now stays false until Statsig signals readiness, then evaluates the gate. Two separate `scope.launch` blocks are used so the Statsig collector and the search params collector run concurrently — a single coroutine would have blocked on collect forever and prevented search from working.
* Updated `MockStatsigClient` with a `startReady: Boolean = true` parameter and a `triggerReady()` method to support deferred-initialization test scenarios.

* Added a test covering the new behavior: banner is false while not ready, flips to the gate value once triggerReady() is called.

**Crash 2 — Null intent extras**:

`Observable.just(null)` throws before any operators run in RxJava 2. Replaced arguments() with a null-safe Kotlin expression: `this.arguments?.let { Observable.just(it) } ?: Observable.empty()`, which emits the bundle when present and completes silently when not.
Updated `setUpEnvironment` in the test to accept `Bundle?` and added `testNullBundle_doesNotCrash` which instantiates the VM with null and asserts no crash, no project-dependent emissions, and that email still loads independently.

# 👀 See
- No user facing changes

# 📋 QA

- Verify normal Search & Filter functionality (search, filters, pagination) still works end to end.
- Verify the Report Project flow works normally when reached through the standard in-app path (project page → report project).
- Crashes were likely generated by opening both activities using CLI: `adb shell am start -n com.kickstarter.dev/com.kickstarter.ui.activities.ReportProjectActivity`, found no way of reproducing with user navigation nor deeplinks associated.

# Story 📖

[DISC-199](https://kickstarter.atlassian.net/browse/DISC-199)

[DISC-199]: https://kickstarter.atlassian.net/browse/DISC-199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/viewmodels/ReportProjectViewModel.kt
+++ b/app/src/main/java/com/kickstarter/viewmodels/ReportProjectViewModel.kt
@@ -59,7 +59,7 @@ interface ReportProjectViewModel {
         private val flaggingKind = PublishSubject.create<String>()
         private val urlTag = PublishSubject.create<String>()
 
-        private fun arguments() = Observable.just(this.arguments).filter { it.isNotNull() }.map { requireNotNull(it) }
+        private fun arguments() = this.arguments?.let { Observable.just(it) } ?: Observable.empty()
         private val disposables = CompositeDisposable()
 
         init {
--- a/app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
+++ b/app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
@@ -81,7 +81,11 @@ class SearchAndFilterViewModel(
     private var isLoadingMore = false
 
     init {
-        _isVideoFeedBannerVisible.value = statsigClient.checkGate(StatsigGateKey.ANDROID_VIDEO_FEED.key)
+        scope.launch {
+            statsigClient.isReady.collect { isReady ->
+                _isVideoFeedBannerVisible.value = isReady && statsigClient.checkGate(StatsigGateKey.ANDROID_VIDEO_FEED.key)
+            }
+        }
 
         scope.launch {
             val debounced = _searchTerm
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/kickstarter/viewmodels/ReportProjectViewModel.kt
```diff
@@ -59,7 +59,7 @@ interface ReportProjectViewModel {
         private val flaggingKind = PublishSubject.create<String>()
         private val urlTag = PublishSubject.create<String>()
 
-        private fun arguments() = Observable.just(this.arguments).filter { it.isNotNull() }.map { requireNotNull(it) }
+        private fun arguments() = this.arguments?.let { Observable.just(it) } ?: Observable.empty()
```
Arkariang, line 62: 🤔 when your back @tonyteate and able to discuss something, thinking on extracting this `arguments()` function as a `ViewModel()` extension method and apply to all legacy RXJava based VM's, (this should be added to PF roadmap most likely but open to handle on Discovery!)

PR #2509, opened 2026-05-06
CHECK-47: Re-fetch Statsig configs when Segment and User IDs become available

tonyteate:
# 📲 What

Launch a `Job` that re-fetches feature configurations from Statsig when the Segment Anonymous and KSR User IDs become available.

# 🤔 Why

As described in the docs: "Statsig evaluates gates and experiment buckets based only on the information you provide when you check something." ([User (StatsigUser) Object - Statsig Documentation](https://docs.statsig.com/sdks/user#i-passed-that-attribute-before-why-do-i-need-to-pass-it-again))

In order to target and bucket both anonymous and authenticated users, as well as to ensure a consistent user experience, the app needs to pass all available IDs (and in the future, other User properties) to Statsig when fetching configurations. ([Build your first Device-level Experiment - Statsig Documentation](https://docs.statsig.com/guides/first-device-level-experiment#step-3-optional--update-user-info-for-logged-in-users), [Client SDKs - Statsig Documentation](https://docs.statsig.com/client/introduction#updating-experiment-configuration))

# 🛠 How

Add a method `observeUserAndFetchConfigs()` to our `StatsigClient` that launches a long-lived `Job` for observing the Segment client state and current User. When the Segment client is ready and an Anonymous ID is available, and/or when a user signs in, a call is made to `Statsig.updateUser()` with an updated `StatsigUser` data object.

# 👀 See

# 📋 QA

**Anonymous User without tracking -> Authenticated User**
- Precondition: Fresh install of the app
- Launch the app from the Home Screen or App Drawer (or Android Studio)
- Result:
  - The system navigates to Discover and immediately displays Onboarding
  - The key `"1839732653"` in the JSON map logged under `Statsig.initialize(): - JSON: {...}` includes:
    - `is_user_in_experiment` : `false`
    - `secondary_exposures[0].gateValue` : `false`
- Step through Onboarding to **DECLINE Personalization**, then click through and **log into the app**
- Result:
  - After authentication, the system navigates back to Discover
  - The key `"1839732653"` in the JSON map logged under `Statsig.updateUser: - JSON: {...}` includes:
    - `is_user_in_experiment` : `true`
    - `secondary_exposures[0].gateValue` : `true`
- Terminate the app
- Re-install the app, or Clear storage from the app's system Settings.

**Anonymous User with tracking**
- Precondition: Fresh install of the app
- Launch the app from the Home Screen or App Drawer (or Android Studio)
- Result:
  - The system navigates to Discover and immediately displays Onboarding
  - The key `"1839732653"` in the JSON map logged under `Statsig.initialize(): - JSON: {...}` includes:
    - `is_user_in_experiment` : `false`
    - `secondary_exposures[0].gateValue` : `false`
- Step through Onboarding to **ALLOW Personalization**, then click through to Explore the app
- Navigate away from the app, then return
- Result:
  - The key `"1839732653"` in the JSON map logged under `Statsig.updateUser: - JSON: {...}` includes:
    - `is_user_in_experiment` : `true`
    - `secondary_exposures[0].gateValue` : `true`

# Story 📖

[\[CHECK-47\] [Android] Verify logged-in and logged-out users are correctly attributed in Statsig - Jira](https://kickstarter.atlassian.net/browse/CHECK-47)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
@@ -4,12 +4,15 @@ import android.content.Context
 import com.kickstarter.KSApplication
 import com.kickstarter.libs.Build
 import com.kickstarter.libs.CurrentUserTypeV2
+import com.kickstarter.libs.SegmentTrackingClient
 import com.kickstarter.libs.utils.Secrets
 import com.kickstarter.libs.utils.extensions.isFalse
 import com.kickstarter.libs.utils.extensions.isTrue
+import com.segment.analytics.Analytics
 import com.statsig.androidsdk.FeatureGate
 import com.statsig.androidsdk.InitializationDetails
 import com.statsig.androidsdk.Statsig
+import com.statsig.androidsdk.Statsig.getInitializeResponseJson
 import com.statsig.androidsdk.StatsigOptions
 import com.statsig.androidsdk.StatsigUser
 import com.statsig.androidsdk.Tier
@@ -19,7 +22,12 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.plus
+import kotlinx.coroutines.rx2.asFlow
+import timber.log.Timber
 
 enum class StatsigGateKey(val key: String) {
     ANDROID_VIDEO_FEED("android_video_feed")
@@ -33,7 +41,7 @@ interface StatsigClientType {
 
     /**
      * Returns the full [FeatureGate] object for [gateName], including the evaluated boolean value,
-     * the matched [com.statsig.androidsdk.FeatureGate.getRuleID] (which rule in the Statsig
+     * the matched [FeatureGate.getRuleID] (which rule in the Statsig
      * console was responsible for the result), and [com.statsig.androidsdk.EvaluationDetails]
      * describing how the value was resolved (network, cache, etc.).
      *
@@ -42,11 +50,13 @@ interface StatsigClientType {
      */
     fun getFeatureGate(gateName: String): FeatureGate = Statsig.getFeatureGate(gateName)
     fun getExperiment(experimentName: String) = Statsig.getExperiment(experimentName) // TODO: for test MOCK statsig DynamicConfig type or expose to the rest of the app just the json values, will explore in the future
-    suspend fun updateUser(id: String) = Statsig.updateUser(StatsigUser(id))
-    fun getStableId() = Statsig.getStableID()
+    suspend fun updateUser(user: StatsigUser) = Statsig.updateUser(user)
+    fun getStableId() = Statsig.getStatsigMetadata().stableID
 
     /** Emits true once the SDK has been successfully initialized. */
     val isReady: StateFlow<Boolean>
+
+    suspend fun handleObservedUserData(stableId: String?, segmentAnonymousId: String?, userId: String?)
 }
 
 /**
@@ -62,17 +72,30 @@ open class StatsigClient @JvmOverloads constructor(
     internal val build: Build,
     private val context: Context,
     private val currentUser: CurrentUserTypeV2,
+    private val segmentTrackingClient: SegmentTrackingClient,
+    private val getAnonymousId: () -> String? = {
+        Analytics.with(context).analyticsContext.traits().anonymousId()
+    },
     private val sdkInitializer: suspend StatsigClient.() -> InitializationDetails? = {
         val isProduction = build.isRelease && Build.isExternal()
         val options = StatsigOptions().apply {
             setTier(if (isProduction) Tier.PRODUCTION else Tier.STAGING)
         }
-        Statsig.initialize(
+        val user = StatsigUser()
+        val details = Statsig.initialize(
             application = context as KSApplication,
             if (isProduction) Secrets.Statsig.PRODUCTION else Secrets.Statsig.STAGING,
-            StatsigUser(),
+            user,
             options = options
         )
+        Timber.d("Statsig.initialize():")
+        Timber.d("- Stable ID: ${getStableId()}")
+        val initializeResponseJson = Statsig.runCatching { getInitializeResponseJson() }.getOrNull()
+        initializeResponseJson?.let {
+            Timber.d("- EvaluationDetails: ${it.getEvalDetails()}")
+            Timber.d("- JSON: ${it.getInitializeResponseJSON()}")
+        }
+        details
     }
 ) : StatsigClientType {
 
@@ -98,8 +121,8 @@ open class StatsigClient @JvmOverloads constructor(
      * @param errorCallback invoked with the error when initialization fails or throws
      */
     fun initialize(scope: CoroutineScope, dispatcher: CoroutineDispatcher = Dispatchers.IO, errorCallback: (Throwable) -> Unit) {
-        this.scope = scope
-        scope.launch(context = dispatcher) {
+        this.scope = (scope + dispatcher)
+        scope.launch {
             try {
                 val details = sdkInitializer()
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -62,17 +72,30 @@ open class StatsigClient @JvmOverloads constructor(
...
+    private val getAnonymousId: () -> String? = {
+        Analytics.with(context).analyticsContext.traits().anonymousId()
+    },
     private val sdkInitializer: suspend StatsigClient.() -> InitializationDetails? = {
         val isProduction = build.isRelease && Build.isExternal()
         val options = StatsigOptions().apply {
             setTier(if (isProduction) Tier.PRODUCTION else Tier.STAGING)
         }
-        Statsig.initialize(
+        val user = StatsigUser()
```
tonyteate, line 84: In a follow-up PR, planning to keep this `StatsigUser` object in a class-level `MutableStateFlow`.

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -98,8 +121,8 @@ open class StatsigClient @JvmOverloads constructor(
      * @param errorCallback invoked with the error when initialization fails or throws
      */
     fun initialize(scope: CoroutineScope, dispatcher: CoroutineDispatcher = Dispatchers.IO, errorCallback: (Throwable) -> Unit) {
-        this.scope = scope
-        scope.launch(context = dispatcher) {
+        this.scope = (scope + dispatcher)
```
tonyteate, line 124: @Arkariang For clarity of intent, combining these for now. Removed the test that checked for `scope` equivalence. Thinking about refactoring this such that e.g.  `scope` is passed into the constructor and available for all methods. Maybe `KSApplication.applicationScope` can be available from `Environment`?

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -113,4 +136,57 @@ open class StatsigClient @JvmOverloads constructor(
...
+        segmentAnonymousId?.let {
+            statsigUser.customIDs = mapOf("segmentAnonymousId" to it)
+        }
+        userId?.let {
+            statsigUser.userID = userId
+        }
+
+        try {
+            Timber.d("Statsig.updateUser($statsigUser):")
+            updateUser(statsigUser)
```
tonyteate, line 178: In a follow-up PR, when this call is successful, we can update the `MutableStateFlow<StatsigUser>` afterwards.

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -62,17 +72,30 @@ open class StatsigClient @JvmOverloads constructor(
...
+        val details = Statsig.initialize(
             application = context as KSApplication,
             if (isProduction) Secrets.Statsig.PRODUCTION else Secrets.Statsig.STAGING,
-            StatsigUser(),
+            user,
             options = options
         )
+        Timber.d("Statsig.initialize():")
+        Timber.d("- Stable ID: ${getStableId()}")
+        val initializeResponseJson = Statsig.runCatching { getInitializeResponseJson() }.getOrNull()
```
tonyteate, line 93: Tidbit here since it's not well documented on Statsig's website for the Android SDK: [`getInitializeResponseJson()`](https://github.com/statsig-io/android-sdk/blob/7824053f59ebd2ee537e3d64910fe2d0eb739c97/src/main/java/com/statsig/androidsdk/StatsigClient.kt#L811) returns a JSON string of the current feature configurations which is really helpful for debugging without logging or relying on exposure events.

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -62,17 +72,30 @@ open class StatsigClient @JvmOverloads constructor(
     internal val build: Build,
     private val context: Context,
     private val currentUser: CurrentUserTypeV2,
+    private val segmentTrackingClient: SegmentTrackingClient,
+    private val getAnonymousId: () -> String? = {
+        Analytics.with(context).analyticsContext.traits().anonymousId()
```
Arkariang, line 77: Are we accesign directly here Segment without using the client wrapper `SegmentTrackingClient` what is the reason for that? 
Have you thought in adding to our currentUser a new field for the AnonymousID from Segment?

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
Arkariang: Sounds good,  back at the time when added applicationScope to KSApplication we were trying to test the approach of not using DI. But feels perfectly reasonable to have it on Environment

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -113,4 +136,57 @@ open class StatsigClient @JvmOverloads constructor(
             }
         }
     }
+
+    fun observeUserAndFetchConfigs(scope: CoroutineScope) {
```
Arkariang, line 140: Question here, why do we not extract all of this orchestration work into `KSApplication` via a helper like `FirebaseHelper` or even the utils `ApplicationLifecycleUtil` . Once that is done you can keep updating `handleObservedUserData` within the StatsigClient.

tonyteate: In general, I agree this would ideally live in a place where we handled orchestration at the application level. But we currently do not have a consistent approach to this that makes enough sense to separate it from `StatsigClient` without continuing to refactor the client itself (which I hope to do). 

For example, there are mixed concerns spanning the `init` block of `SegmentTrackingClient` itself, `LoginUseCase`, `ApplicationLifecycleUtil` as you noted, etc.

A `StatsigClientHelper` is a good idea as a parallel to the Firebase client, but using the `scope` in this client is more consistent with the purpose of having one, otherwise I'd suggest taking the `scope` out of this client and moving the methods to `suspend` functions, but I think we chatted about that at the time. Moreover, we would still not be consolidating business logic to a place that makes more sense that this client for the intent here.

(Edit: `KSApplication` would actually be more clear, but again only if we moved the scope)

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
tonyteate: Yes, this particular access will change in a subsequent PR. Actually thinking about exposing it through a StateFlow like what we did in `FirebaseHelper`, but still thinking through other refactors to the Segment client. Was more interested in whether or not this was the right overall pattern that you would expect given our chats about how this relates to the Mock.

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
Arkariang: I think there are two separate concerns getting bundled in this thread, and they shouldn't constrain each other:

** 1 - Where the orchestration lives** — StatsigClient vs. a StatsigClientHelper vs. KSApplication.
**2 - How concurrency is managed** — the client owning a stored scope vs. exposing suspend functions and letting the caller bring its own scope.

(1) is a business-logic / discoverability question; (2) is an implementation detail of the client. A helper could own the orchestration regardless of whether the client stores a scope, and conversely keeping the scope in the client doesn't force the orchestration to live there too.

I'm aware we are trying to find a good architecture framework for the mentioned orchestration, and will take some attempts until we are satisfied, so no issues with the proposal of this PR.

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -113,4 +136,57 @@ open class StatsigClient @JvmOverloads constructor(
...
+                Triple(statsigIsReady, segmentIsReady, optionalUser)
+            }
+                .map { (statsigIsReady, segmentIsReady, optionalUser) ->
+                    val stableId = if (statsigIsReady) getStableId() else null
+                    val segmentAnonymousId = if (segmentIsReady) getAnonymousId() else null
+                    val userId = if (optionalUser.isPresent()) optionalUser.getValue()?.id().toString() else null
+                    Triple(stableId, segmentAnonymousId, userId)
+                }
+                // TODO: consider debounce
+                .collect {
```
Arkariang, line 155: Could we change `.collect` over `collectLatest`?
with `combine` every time any of the three upstream flows changes, we get a new tuple, and each emission triggers a full Statsig.updateUser() round-trip.
`collectLatest` cancels the in-flight handler whenever a newer value arrives.

I assume scenarions like the above is why the suspend functions are the preferred option to handle concurrency, much easier to track!.

Arkariang (approved):
Tested the described steps on QA, works fine 👍

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
tonyteate: Yes, these are different concerns. But are they necessarily separate?

> conversely keeping the scope in the client doesn't force the orchestration to live there too.

No, it does not force the orchestration to live there, but it does imply that orchestration can and very well may live there. That entirely depends on the structure and purpose of the class, as a service object is a valid architectural component. And we've already started down that path with the initialization block, which is why this new method is consistent.

This is also an interesting case, as this is particular flow is arguably not something we might consider to be orchestration. Here we are simply collecting data from the "environment", and are not calling the method of any other client, or interacting across any other domain. We have this similarly in other clients and helpers.

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
tonyteate: Yes great callout

PR #2510, opened 2026-05-06
DISC-146: Social Share updated experience

Arkariang:
# 📲 What

Introduces the Social Share bottom sheet — a full sharing experience that lets users share a Kickstarter project to Instagram, Facebook, WhatsApp, X, Messages, Email, and more, or copy the project link to the clipboard.

# 🤔 Why
The social share sheet is built as a standalone, reusable feature so it can be dropped into any screen that needs sharing (Video Feed and is just the first integration point but PreLaunchPage and Project Page will use it as well).

New Social Sheet is able to share image + text with those social media that allows it:

Screenshot 2026-05-06 at 9 09 29 AM

# 🛠 How
- 📄 [Social Share README](app/src/main/java/com/kickstarter/features/socialshare/README.md) -> Added full context of the feature here, including architecture pieces usage samples.

The feature is structured as a self-contained package under `features/socialshare/`:

- **`SocialShareService` / `AndroidSocialShareService`** — interface that abstracts all Android framework calls (PackageManager, ClipboardManager, FileProvider, intent construction) so the ViewModel never holds a `Context`.
- **`SocialShareViewModel`** — on creation it detects which platform apps are installed and begins caching the project image in the background. Exposes `onPlatformSelected`, `onCopyLinkClicked`, and `onCopiedToastShown`.
- **`SocialShareSheet`** — a `ModalBottomSheet` composable. It reads the ViewModel through `LocalSocialShareViewModel` (a `CompositionLocalProvider`) so the ViewModel is always created by the caller with the correct `shareData` rather than inside the composable itself.
- **`ShareImageCache`** — downloads the project image and writes it to the app cache as a `content://` URI via `FileProvider`, which is required to grant read permission to third-party apps.
- **`SocialShareIntentBuilder`** — constructs the correct `Intent` per platform.

# 👀 See

- Share icon in VideoFeed triggers the new social experience. But heads up here query not returning project image, so a grey placeholder visible instead.
- Go to PreLaunchPage, add the next code snippet to substitute the current share experience for the new one
```
// declare in within compoe
var shareData by remember { mutableStateOf<SocialShareData?>(null) }
 // Before PreLaunchProjectPageScreen body completation
 shareData?.let { data ->
                    val shareViewModel = remember(data) {
                        SocialShareViewModel(
                            shareService = AndroidSocialShareService(applicationContext),
                            shareData = data
                        )
                    }
                    CompositionLocalProvider(LocalSocialShareViewModel provides shareViewModel) {
                        SocialShareSheet(
                            shareData = data,
                            isVisible = true,
                            onDismiss = { shareData = null },
                            onIntentReady = { startActivity(it) }
                        )
                    }
                }

// PreLaunchScreen callback 
middleRightClickAction = {
                        projectState.value?.let { project ->
                            shareData = SocialShareData(
                                projectName = project.name() ?: "",
                                projectUrl = project.urls()?.web()?.project() ?: "",
                                imageUrl = project.photo()?.full() ?: "",
                                creatorName = project.creator()?.name() ?: ""
                            )
                        }
                    },

```

https://github.com/user-attachments/assets/04880496-4a27-4bb2-8294-6adedd2bb881

|  |  |

# 📋 QA
- Open the screen, it should show only social media installed on the devide
- Try to share on some social media. Tap a platform (e.g. WhatsApp) → the native share sheet / app should open pre-filled with the project URL. Tap Copy link → a "Link copied!" snackbar should appear; paste to verify the project URL is on the clipboard.

# Story 📖

[DISC-146](https://kickstarter.atlassian.net/browse/DISC-146)

[DISC-146]: https://kickstarter.atlassian.net/browse/DISC-146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/features/socialshare/ui/icons/SocialShareIcons.kt
Arkariang, line 1: Icons generated with Valkyrie extension (From figma exported SVG's to compose ImageVector), no need to review deeply!

app/src/main/java/com/kickstarter/features/socialshare/ShareImageCache.kt
Arkariang, line 1: This piece is not tested as of now, still some work to do here on follow up ticket -> https://kickstarter.atlassian.net/browse/DISC-208

app/src/main/java/com/kickstarter/features/socialshare/AndroidSocialShareService.kt
```diff
@@ -0,0 +1,89 @@
...
+        shareData: SocialShareData,
+        imageUri: Uri?
+    ): Intent?
+}
+
+/**
+ * Production implementation of [SocialShareService] that delegates to real Android APIs.
+ *
+ * **Important:** always pass [Context.getApplicationContext] — never an Activity context.
+ * Application context is safe to hold indefinitely and cannot cause a memory leak.
```
tonyteate, line 54: In this case would it be possible to do:

```
class AndroidSocialShareService(
    context: Context
) : SocialShareService  {
    private val context = context.applicationContext
    ...
}
```

?

app/src/main/java/com/kickstarter/features/socialshare/viewmodel/SocialShareViewModel.kt
```diff
@@ -0,0 +1,128 @@
...
+            errorAction.invoke("Could not open ${platform.name}")
+            return
+        }
+
+        intentLaunchAction.invoke(intent)
+    }
+
+    fun onCopyLinkClicked() {
+        shareService.copyToClipboard("Kickstarter project link", shareData.projectUrl)
+        scope.launch {
```
tonyteate, line 88: Do we need to call scope.launch to emit new state in these?

Edit: mostly a nit since we could update the `value` directly in these cases

app/src/main/java/com/kickstarter/features/socialshare/SocialShareIntentBuilder.kt
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+
+    /**
+     * Email — text always included; image attached as an inline attachment when available.
+     * ACTION_SEND with image/jpeg opens email clients in compose mode with the image
+     * attached and the body/subject pre-filled. Falls back to ACTION_SENDTO + mailto:
+     * (no attachment support) when no image is cached yet.
+     */
+    private fun emailIntent(shareData: SocialShareData, imageUri: Uri?): Intent {
```
tonyteate, line 177: Definitely seems like it's the best Android can do, but wish we could target email apps more specifically in the case w/ an image attachment 🤔

tonyteate (approved):
Brilliant ✨

app/src/main/java/com/kickstarter/features/socialshare/ui/icons/SocialShareIcons.kt
tonyteate: A Dark Mode version of the grey ones could be nice. Seems like we can possibly use the same paths and just update the `Color(0xFF3C3C3C)` in a lot of cases.

Arkariang: Will do a follow up ticket for it, or add it to the analytics remaining piece!

app/src/main/java/com/kickstarter/features/socialshare/viewmodel/SocialShareViewModel.kt
Arkariang: Will do when this piece is integrated within ProjectPage. For videoFeed/preLaunchProject not likely, but the integration with old XML/RXJava will be another story

PR #2511, opened 2026-05-12
DISC-141: Analytics

Arkariang:
# 📲 What

Adds Segment analytics instrumentation to the Video Feed feature. Implements four event types agreed upon in the DISC-141 spec: Video Impression, Video Swipe with watch progress, Progress Bar Tap, and Video CTA Clicked (save, creator_profile, share, play, pause).

# 🤔 Why

In order to have visibility into how users are engaging with the feed — which videos are being watched, how deep engagement is, and which CTAs are being used. This data is needed to guide product decisions on the feature.

# 🛠 How

Event architecture mirrors the existing Analytics model with two Segment event names (PAGE_VIEWED, CTA_CLICKED) with context_page = "video_feed" on all events.

Entry surface ("discovery" or "search") is propagated from the launch entry point through the Intent (PREVIOUS_SCREEN extra) into VideoFeedViewModel and attached to every impression and swipe event.

**Impression vs. Swipe**:
`trackVideoFeedImpression` → PAGE_VIEWED fires on every `pagerState.settledPage` change, including first load. This covers impression measurement independently.
`trackVideoFeedPageViewed` → PAGE_VIEWED fires only when previousSettledPage is valid (never on first load). It consolidates swipe navigation + watch depth (total_watch_time, total_video_duration, from_video_id) into one event.

**Watch time collection:** KSVideoPlayer gained an onBecameInactive callback that fires with (watchTimeMs, videoDurationMs) as the player deactivates during a swipe. VideoFeedScreen stores this in a watchTimeByPage map keyed by page index. When LaunchedEffect fires on the next settled page, it reads (and removes) the stored data for the previous page — ensuring the data is always ready regardless of animation timing.

CTAs wired: play/pause toggle, progress bar tap, bookmark (save), profile click, and share button all route through viewModel.onCTAClicked or viewModel.onProgressBarTapped.

# 👀 See
CTA_Clicked-progress-bar

CTA_Clicked-video-pause

VideoSwipe

VideoImpresion

|  |  |

# 📋 QA

- Open the app from Discovery/Search and tap the Video Feed banner.
- - Verify a PAGE_VIEWED event fires in Segment with context_page = "video_feed", video_feed_entry_surface = "discovery/Search", video_feed_position_in_session = 0.
- - Swipe to the next video.
- -Verify a second PAGE_VIEWED fires with video_feed_position_in_session = 1, video_feed_from_video_id matching the previous video, and non-zero video_feed_total_watch_time.
 

- Tap share, crator_profile, pause, or scrub the progress bar, they all should fire a CTA with the corresponding context_cta 

 

# Story 📖

[DISC-141](https://kickstarter.atlassian.net/browse/DISC-141)

[DISC-141]: https://kickstarter.atlassian.net/browse/DISC-141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/ui/extensions/ActivityExt.kt
```diff
@@ -172,9 +172,9 @@ fun Activity.startPledgeRedemption(project: Project) {
     }
 }
 
-fun Activity.startVideoFeed() {
+fun Activity.startVideoFeed(entrySurface: String) {
```
tonyteate, line 175: Since this is tied to analytic events, are you ok with the type here being  `String` as opposed to `ThirdPartyEventValues.ScreenName` for example? If we went with `ScreenName` we could add an empty/unknown string.

tonyteate:
Just double checking in general that all `*_video_id` properties should be set to the Project id.

app/src/main/java/com/kickstarter/ui/compose/designsystem/videoplayer/KSVideoPlayer.kt
```diff
@@ -149,6 +153,12 @@ fun KSVideoPlayer(
     var showControls by remember { mutableStateOf(false) }
     val hazeState = rememberHazeState()
 
+    // Keep references to the latest callback lambdas so lambdas inside remember(exoPlayer)
+    // blocks always invoke the current version even if the parent recomposes.
```
tonyteate, line 157: nit: maybe naming & declaration convention `currentOnPlayPauseToggle by rememberUpdatedState(onPlayPauseToggle)` and then be able to call `currentPlayPauseToggle()` for clarity.

tonyteate (approved):
lgtm 🎉

app/src/main/java/com/kickstarter/ui/extensions/ActivityExt.kt
Arkariang: Will tackle on follow up PR, this makes sense 👍

Arkariang:
> Just double checking in general that all *_video_id properties should be set to the Project id.


Each video within each project has it's own individual id obtained from the query, passing over to the analytic events the corresponding id's:
Screenshot 2026-05-14 at 1 31 55 PM

PR #2512, opened 2026-05-13
CHECK-52: Keep a StatsigUser in StatsigClient + support AA Experiments

tonyteate:
# 📲 What

- Add a stateful and reactive `StatsigUser` property to `StatsigClient`. 
- Create a new `ExperimentsViewModel` that logs exposure events for two A/A experiments for a given `StatisigUser`, and include this VM in Project Page.

# 🤔 Why

Checkout team is running an A/A test with Statsig - a null hypothesis test to validate that our configurations work correctly. There are two experiments as a part of this test -- one for anonymous users, and the other for authenticated users. (See [Mobile A/B Testing for Conversion with Statsig - Checkout - Confluence](https://kickstarter.atlassian.net/wiki/spaces/Checkout/pages/4438818822/Mobile+A+B+Testing+for+Conversion+with+Statsig))

Statsig's analysis of Metric impact depends on the app logging "exposure" events through the SDK ([Raw Events - Statsig Documentation](https://docs.statsig.com/metrics/raw-events). For experiments, an exposure event is logged for every non-redundant call to `Statsig.getExperiment()` for a given User. A given user is distinguished by the properties included in the [StatsigUser](https://docs.statsig.com/sdks/user) object passed to Statsig via `initialize()` and `updateUser()`.

To ensure that an experiment is retrieved for an up-to-date StatsigUser object, we can keep track of the latest one sent to Statsig, and make a call to `getExperiment()` whenever the object is updated.

# 🛠 How

- Add a `MutableStateFlow<StatsigUser>` to `StatsigClient` and set a new `value` upon a successful call to `Statsig.updateUser()`, particularly in reaction to newly observed user data (i.e. anonymous and user IDs) via `handleObservedUserData()`.
- Create a new `ExperimentsViewModel` that collects `StatsigClient`'s `MutableStateFlow<StatsigUser>`, and calls `getExperiment()` for both A/A experiments in order to properly log exposure events.
- Add an instance of `ExperimentsViewModel` to Project Page to only trigger such exposures for users who view the Project Page, and to validate this orchestration in preparation for further A/B tests on the Project Page.
- Re: https://github.com/kickstarter/android-oss/pull/2509:
  - Add a method `getAnonymousIdOrNull()` to `SegmentTrackingClient` that returns the Segment anonymous ID if possible; remove the `getAnonymousId` lambda from `StatsigClient`; and refactor tests in `StatsigClientTest` accordingly.

# 👀 See

**Note:** As this is just an A/A experiment with no change to the user experience, the app does not and should not behave any differently.

# 📋 QA

> [!IMPORTANT]
> Before starting these QA steps, in `StatsigClient`:
> - Set `NoOpAuthenticatedUsers.name` to `"2026-05-12_checkout_aa"`
> - Set `NoOpAnonymousUsers.name` to `"2026-05-12_checkout_aa_logged_out"`

- Precondition: Fresh install of the app (Internal build)
- Launch the app from the Home Screen, App Drawer, Android Studio)
- Result:
  - The system navigates to Discover and immediately displays Onboarding
- Step through Onboarding to **DECLINE Personalization**, then click through to **Explore the App**
- Open any Project Page for a Project that you are not backing
- Results Logged for `tag:ExperimentsViewModel`:
  - Under `statsigUser[stableID: ...]`
    - `userID` : `null`
  - Under `authenticatedExperiment: 2026-05-12_checkout_aa`
    - `getIsUserInExperiment()` : `false`
  - Under `anonymousExperiment: 2026-05-12_checkout_aa_logged_out`
    - `getIsUserInExperiment()` : `true`
- Click Back this Project to trigger the authentication flow, and sign into the app (you do not need to actually pledge).
- Results Logged for `tag:ExperimentsViewModel`:
  - Under `statsigUser[stableID: ...]`
    - `userID` : `<Your KSR User ID`
  - Under `authenticatedExperiment: 2026-05-12_checkout_aa`
    - `getIsUserInExperiment()` : `true`
  - Under `anonymousExperiment: 2026-05-12_checkout_aa_logged_out`
    - `getIsUserInExperiment()` : `true`

**Optional:** Exposure events can take a few minutes to show up in the Console, but you should also see at least entry for each distinct User object in the Exposure Stream under Diagnostics for each experiment.

# Story 📖

- [\[CHECK-52\] Android: Add checks for A/A experiments to Project page - Jira](https://kickstarter.atlassian.net/browse/CHECK-52)
- [\[CHECK-158\] [Android] Create A/A experiment for logged-out users with Segment Anonymous ID - Jira](https://kickstarter.atlassian.net/browse/CHECK-158)

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/SegmentTrackingClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/SegmentTrackingClient.kt
@@ -77,6 +77,12 @@ open class SegmentTrackingClient(
         }
     }
 
+    override fun getAnonymousIdOrNull() = try {
+        Analytics.with(context).analyticsContext.traits().anonymousId()
+    } catch (_: Exception) {
+        null
+    }
+
     private fun privateInitializer() {
         if (build.isDebug) {
             Timber.d("${type().tag} isEnabled: ${this.isEnabled()}")
--- a/app/src/main/java/com/kickstarter/libs/SegmentTrackingClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/SegmentTrackingClient.kt
@@ -78,7 +78,9 @@ open class SegmentTrackingClient(
     }
 
     override fun getAnonymousIdOrNull() = try {
-        Analytics.with(context).analyticsContext.traits().anonymousId()
+        if (isEnabled())
+            Analytics.with(context).analyticsContext.traits().anonymousId()
+        else null
     } catch (_: Exception) {
         null
     }
--- a/app/src/main/java/com/kickstarter/libs/TrackingClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/TrackingClient.kt
@@ -75,6 +75,9 @@ abstract class TrackingClient(
      */
     abstract fun trackingData(eventName: String, newProperties: Map<String, Any?>)
 
+    // TODO: re-consider `abstract` here vs `open` in `SegmentTrackingClient`
+    abstract fun getAnonymousIdOrNull(): String?
+
     // Default property values
     override fun brand(): String = android.os.Build.BRAND
 
--- a/app/src/main/java/com/kickstarter/libs/TrackingClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/TrackingClient.kt
@@ -75,7 +75,7 @@ abstract class TrackingClient(
      */
     abstract fun trackingData(eventName: String, newProperties: Map<String, Any?>)
 
-    // TODO: re-consider `abstract` here vs `open` in `SegmentTrackingClient`
+    // Alternatively `open` in `SegmentTrackingClient` pending any other tracking clients
     abstract fun getAnonymousIdOrNull(): String?
 
     // Default property values
--- a/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
@@ -41,15 +41,15 @@ object StatsigExperiments {
         val parameters: T
     }
 
-    object NoopAuthenticatedUsers : Experiment<NoopAuthenticatedUsers.Parameters> {
+    object NoOpAuthenticatedUsers : Experiment<NoOpAuthenticatedUsers.Parameters> {
         override val name = "logged_in_aa_experiment"
         override val parameters = Parameters
         object Parameters {
             const val TEST = "test_parameter"
         }
     }
-    object NoopAnonymousUsers : Experiment<NoopAnonymousUsers.Parameters> {
-        override val name = "logged_out_aa_experiment"
+    object NoOpAnonymousUsers : Experiment<NoOpAnonymousUsers.Parameters> {
+        override val name = "logged_out_aa_experiment_android"
         override val parameters = Parameters
         object Parameters {
             const val TEST = "test_parameter"
--- a/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
@@ -226,6 +226,6 @@ open class StatsigClient @JvmOverloads constructor(
     }
 
     companion object {
-        const val KEY_SEGMENT_ANONYMOUS_ID = "segmentAnonymousId"
+        const val KEY_SEGMENT_ANONYMOUS_ID = "segmentAnonymousID"
     }
 }
--- a/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
@@ -42,14 +42,14 @@ object StatsigExperiments {
     }
 
     object NoopAuthenticatedUsers : Experiment<NoopAuthenticatedUsers.Parameters> {
-        override val name = "2026-05-12_checkout_aa"
+        override val name = "logged_in_aa_experiment"
         override val parameters = Parameters
         object Parameters {
             const val TEST = "test_parameter"
         }
     }
     object NoopAnonymousUsers : Experiment<NoopAnonymousUsers.Parameters> {
-        override val name = "2026-05-12_checkout_aa_logged_out"
+        override val name = "logged_out_aa_experiment"
         override val parameters = Parameters
         object Parameters {
             const val TEST = "test_parameter"
--- a/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
@@ -1,6 +1,7 @@
 package com.kickstarter.libs.featureflag
 
 import android.content.Context
+import androidx.annotation.VisibleForTesting
 import com.kickstarter.KSApplication
 import com.kickstarter.libs.Build
 import com.kickstarter.libs.CurrentUserTypeV2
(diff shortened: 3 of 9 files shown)


app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -33,6 +35,28 @@ enum class StatsigGateKey(val key: String) {
     ANDROID_VIDEO_FEED("android_video_feed")
 }
 
+object StatsigExperiments {
```
tonyteate, line 38: Just experimenting with a way to manage names and parameter keys for different Statsig experiemnts. The name `Experiment` for the interface seems less than ideal, but this is all relatively easy to change with the IDE.

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
```diff
@@ -160,14 +190,13 @@ open class StatsigClient @JvmOverloads constructor(
     }
 
     override suspend fun handleObservedUserData(stableId: String?, segmentAnonymousId: String?, userId: String?) {
-        // `stableId` is automatically attached to `StatsigUser` internally, but included here for clarity, testing, and debugging
         Timber.d("handleObservedUserData(stableId: $stableId, segmentAnonymousId: $segmentAnonymousId, userId: $userId)")
 
-        if (stableId == null && segmentAnonymousId == null && userId == null) return
+        if (stableId == null) return
```
tonyteate, line 195: We want to allow `segmentAnonymousId` and `userId` to be `null` for when users log out. In the future we will also want to reset the Statsig Stable ID / User.

app/src/main/java/com/kickstarter/libs/SegmentTrackingClient.kt
```diff
@@ -77,6 +77,12 @@ open class SegmentTrackingClient(
         }
     }
 
+    override fun getAnonymousIdOrNull() = try {
```
tonyteate, line 80: Considering if this should perhaps be gated by `isEnabled()`. Thoughts @Arkariang ?

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
tonyteate, line 155: @Arkariang A thought for the future: after some in-depth QA, it's worth noting that we probably want `_isReady` (or another piece of state that can have more values) to represent that `Statsig.initialize()` was called and completed, regardless of whether or not the call was "successful" (e.g. it might fail due to the network). The SDK seems intentionally designed to allow us to continue to call `checkGate()`, `getExperiment()`, etc. and everything will just return cached values, or default values (or `null`). But depending on the scenario, if `_isReady` is never `true`, we might never check.

app/src/main/java/com/kickstarter/viewmodels/projectpage/ExperimentsViewModel.kt
tonyteate, line 1: Since there is no behavior change, a new ViewModel is easier to integrate into `ProjectPageActivity`, create tests for, and remove.

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
Arkariang: Sounds good, `_isReady` was tailored very specifically for the checkGates on Discovery Screen on First load, happy to extend the concept to an actual data class with more information and not just success initialization 👍

app/src/main/java/com/kickstarter/libs/SegmentTrackingClient.kt
Arkariang: Agree!

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
Arkariang: question though, when the user logs out, `segmentAnonymousId` seems to never be null on Segment SDK, it keeps the same value over the entire session, should we really allow `segmentAnonymousId` to be null? We risk a discrepancy then between Statsig information about Segment's ID.
For userID totally ok, is the default value when logged out

app/src/main/java/com/kickstarter/viewmodels/projectpage/ExperimentsViewModel.kt
```diff
@@ -0,0 +1,88 @@
...
+import kotlin.coroutines.EmptyCoroutineContext
+
+@OptIn(FlowPreview::class)
+class ExperimentsViewModel(
+    environment: Environment,
+    testDispatcher: CoroutineDispatcher? = null
+) : ViewModel() {
+
+    private val scope = viewModelScope + (testDispatcher ?: EmptyCoroutineContext)
+    private val statsigClient = environment.statsigClient()!!
```
Arkariang: super nitpick: could we use requireNotNull(environment.statsigClient()) instead of `!!`?

app/src/main/java/com/kickstarter/viewmodels/projectpage/ExperimentsViewModel.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    init {
+        scope.launch {
+            statsigClient.isReady.first { it }
+            statsigClient.statsigUser
+                .onEach {
```
Arkariang: nitpick: let's add a `catch` operator and log into Firebase, - our typical provideErrorAction struture - we should have absolutely no issues but we should know in case Project Page was affected

app/src/main/java/com/kickstarter/libs/TrackingClient.kt
```diff
@@ -75,6 +75,9 @@ abstract class TrackingClient(
      */
     abstract fun trackingData(eventName: String, newProperties: Map<String, Any?>)
 
+    // TODO: re-consider `abstract` here vs `open` in `SegmentTrackingClient`
```
Arkariang: By looking at the comment, more inclined to `abstract` to make the contract "unbreakable" 🤔 , but no strong opinions agains open either

app/src/test/java/com/kickstarter/libs/StatsigClientTest.kt
Arkariang, line 1: 💖

app/src/main/java/com/kickstarter/viewmodels/projectpage/ExperimentsViewModel.kt
tonyteate: Ha I actually prefer the double-bang, it has the same effect right? Maybe throws a different exception though. Do you prefer the function for readability? I don't mind switching over.

app/src/main/java/com/kickstarter/viewmodels/projectpage/ExperimentsViewModel.kt
tonyteate: Good idea - will add the `catch` and log to Firebase.

app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
tonyteate: Currently when a user logs out, `segmentAnonymousId` gets reset to a different value. We want to allow it to be `null` here in case a user has declined consent management. So the discrepancy won't be with the information Statsig has (though we eventually need to make a decision about Stable ID), but with the idea that a user that has intentionally logged out is still eligible for Experiments that target `segmentAnonymousId`. But this is because we don't re-prompt for consent management. If we do re-prompt in the future, gating the new method `getAnonymousIdOrNull()` will appropriately result in a `null` value here.

Arkariang (dismissed):
Nice job!

app/src/main/java/com/kickstarter/viewmodels/projectpage/ExperimentsViewModel.kt
Arkariang: readability and seems more intentional

Arkariang (approved):

PR #2515, opened 2026-05-19
DISC-142: Small fixes 2

Arkariang:
# 📲 What

- Fix an analytics bug where the `video_feed_from_video_id` property was incorrectly sending the project ID of the video the user swiped away from, instead of the video ID. 
- Improved lifecycle awareness in the video player so playback automatically stops when the app is backgrounded or the user navigates to a creator profile/project page
- Locks VideoFeedActivity to portrait orientation.

# 🤔 Why

- The `video_feed_from_video_id` analytic property was being populated using fromProject.id() — the Kickstarter project ID — instead of the VideoFeedItem.videoId. 

- Video playback was not pausing fast enough when navigating away (project page or creator bio), because it waited for Compose to signal inactivity rather than reacting immediately to lifecycle events, resulting on audio still playing for a few seconds when in other activities.

# 🛠 How

**Analytics fix:**
Changed the `fromProject: Project` parameter across the entire call chain (VideoFeedScreen → VideoFeedActivity → VideoFeedViewModel → AnalyticEvents → AnalyticEventsUtils) to `fromVideoFeedItem: VideoFeedItem` .

**Lifecycle awareness:**
Added a `LifecycleEventObserver` inside `KSVideoPlayer` via `DisposableEffect` that tracks ON_RESUME/ON_PAUSE events using `LocalLifecycleOwner`.The LaunchedEffect controlling exoPlayer.playWhenReady now depends on both isActive and isAppInForeground, so playback stops immediately when the lifecycle pauses.

**Portrait lock:**
Added `android:screenOrientation="portrait"` and `android:configChanges="orientation"` to VideoFeedActivity in AndroidManifest.xml.

# 👀 See
- No user facing channges

|  |  |

# 📋 QA

Open the Video Feed.
- Swipe through a few videos and verify in analytics/Segment that video_feed_from_video_id matches the video ID (not the project ID) of the video you swiped away from.
- While a video is playing -ideally a video with sound-, navigate to the project page — confirm the video stops immediately (no delay).
- Background the app while a video is playing — confirm it pauses. Return to the app — confirm it resumes.
- Rotate the device while in the Video Feed — confirm the screen stays locked in portrait.

# Story 📖

[DISC-142](https://kickstarter.atlassian.net/browse/DISC-142)

[DISC-142]: https://kickstarter.atlassian.net/browse/DISC-142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -228,8 +228,9 @@
             android:theme="@style/VideoPlayerActivity" />
         <activity
             android:name=".features.videofeed.ui.VideoFeedActivity"
-            android:configChanges="keyboardHidden|screenSize"
+            android:configChanges="keyboardHidden|screenSize|orientation"
             android:exported="false"
+            android:screenOrientation="portrait"
             android:theme="@style/KSTheme" />
         <activity
             android:name=".ui.activities.BackingActivity"
--- a/app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedActivity.kt
@@ -94,8 +94,8 @@ class VideoFeedActivity : ComponentActivity() {
                     onVideoImpression = { videoFeedItem, position ->
                         viewModel.onVideoImpression(videoFeedItem, position)
                     },
-                    onVideoPageSettled = { videoFeedItem, toPosition, fromProject, watchTimeMs, videoDurationMs ->
-                        viewModel.onVideoPageSettled(videoFeedItem, toPosition, fromProject, watchTimeMs, videoDurationMs)
+                    onVideoPageSettled = { videoFeedItem, toPosition, fromVideoFeedItem, watchTimeMs, videoDurationMs ->
+                        viewModel.onVideoPageSettled(videoFeedItem, toPosition, fromVideoFeedItem, watchTimeMs, videoDurationMs)
                     },
                     onPlayPauseTap = { project, isPlaying ->
                         val cta = if (isPlaying) CtaContextName.VIDEO_PLAY else CtaContextName.VIDEO_PAUSE
--- a/app/src/main/java/com/kickstarter/ui/compose/designsystem/videoplayer/KSVideoPlayer.kt
+++ b/app/src/main/java/com/kickstarter/ui/compose/designsystem/videoplayer/KSVideoPlayer.kt
@@ -156,6 +156,7 @@ fun KSVideoPlayer(
     var showControls by remember { mutableStateOf(false) }
     val hazeState = rememberHazeState()
     val lifecycleOwner = LocalLifecycleOwner.current
+    var isAppInForeground by remember { mutableStateOf(true) }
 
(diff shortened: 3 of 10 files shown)


app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedActivity.kt
Arkariang, line 1: Exchanging `Project` over to `VideoFeedItem` that contains both project and videoId information

app/src/main/java/com/kickstarter/ui/compose/designsystem/videoplayer/KSVideoPlayer.kt
```diff
@@ -152,17 +155,34 @@ fun KSVideoPlayer(
...
+        }
+    }
+
     // - Updated progress bar only when active and not scrubbing
-    LaunchedEffect(isActive) {
-        exoPlayer.playWhenReady = isActive
-        if (isActive) {
+    LaunchedEffect(isActive, isAppInForeground) {
+        val shouldPlay = isActive && isAppInForeground
+        exoPlayer.playWhenReady = shouldPlay
```
tonyteate, line 184: Nice. Any concern about the ExoPlayer playing state not being tied directly to `isActive` anymore? For example, there's a `DisposableEffect(isActive) { ... }` later on. Does it still match expectations intuitively?

app/src/main/AndroidManifest.xml
```diff
@@ -228,8 +228,9 @@
             android:theme="@style/VideoPlayerActivity" />
         <activity
             android:name=".features.videofeed.ui.VideoFeedActivity"
-            android:configChanges="keyboardHidden|screenSize"
+            android:configChanges="keyboardHidden|screenSize|orientation"
             android:exported="false"
+            android:screenOrientation="portrait"
```
tonyteate, line 233: Makes complete sense and is definitely ideal. Just a note/reminder fwiw that this setting [_might_](https://developer.android.com/about/versions/16/behavior-changes-16#ignore-orientation) not work for tablets / large phones anymore. (Looks like we could opt out temporarily, but I wouldn't worry about it now.) Also using `configChanges` is a recommended path forward iirc anyways.

tonyteate (approved):
👍

app/src/main/AndroidManifest.xml
Arkariang: That's a good point, in our top most devices there are no tablets so confident (at least for phase 1 release) will take a look though to 'configChanges' not familiar with it

app/src/main/java/com/kickstarter/ui/compose/designsystem/videoplayer/KSVideoPlayer.kt
Arkariang: the DisposableEffect(isActive) is an analytics/callback concern, not a playback concern, so it's correct that it doesn't also key on `isAppInForeground`.

PR #2518, opened 2026-05-21
DISC-209: Reftags added for social sharing

Arkariang:
# 📲 What

- Adds platform-specific reftags and a `CTA Clicked` analytics event to the Social Share bottom sheet. 
- Also ships three bug fixes uncovered during implementation: 
- - empty project URL in share payloads for VideoFeed
- - the email intent incorrectly surfacing Instagram as a candidate
- - invisible monochrome icons in dark mode

# 🤔 Why

Share projects from the Video Feed were not attribution-tracked, making it impossible to measure how users discover projects via social sharing. Each platform needed its own reftag appended to the shared URL so downstream analytics can attribute traffic back to the specific channel (X, WhatsApp, email, etc.).

# 🛠 How

**Analytics & reftags**
- Added 10 platform-specific RefTag factory methods (android_project_share_copy_link, android_project_share_x, android_project_share_whatsapp, etc.).
- Added SHARE_PLATFORM_CLICKED to CtaContextName and a new SharePlatformContextType enum covering all 10 platforms.
- Added trackSharePlatformCTAClicked(platform, contextPage) to AnalyticEvents, firing CTA_CLICKED with context_cta, context_page, and context_type properties.
- `SocialShareViewModel` now receives Environment (contains analytics client) and ContextPageName (for the page property). Before building the share intent or copying to clipboard, it appends the platform's reftag to the URL via `UrlUtils.appendRefTag` and fires the analytics event.

**Bug: empty project URL in VideoFeed payloads**
The videoFeedProject GraphQL fragment was missing the url field, so project.urls()?.web()?.project() always returned "". Added url to the fragment and mapped it through VideoFeedTransformers into Urls → Web.project(). Originally Social Share feature was developed under PreLaunch page scope.

**Bug: email intent surfacing Instagram**
emailIntent with an image was using ACTION_SEND + type = "image/jpeg" with no package restriction. Android's intent resolver presented Instagram as a top candidate. Changed to type = "message/rfc822" (the standard email MIME type), which limits the chooser to email clients only while still allowing the image attachment.

**Bug: monochrome icons invisible in dark mode**
Platform icons (Copy Link, X, Email, Messages, More) use hardcoded dark-grey SVG paths (#3C3C3C / black). The icon container was using backgroundSurfaceRaised (dark in dark mode) and tint = Color.Unspecified (no override). Added monochromeIconTint to KSSocialShareColors (black in light, white in dark), switched the container background to socialShare.iconBackground (white in light, #222222 in dark), and applied the tint selectively — brand icons (Instagram, Facebook, WhatsApp) keep Color.Unspecified to preserve their gradients/colors.

# 👀 See

https://github.com/user-attachments/assets/151901ed-4b1f-4fc0-ac6e-b5f55d774086

|  |  |

# 📋 QA

1 - Open the Video Feed and tap the share icon on any project.
Reftags — tap each platform and verify the shared URL contains the expected ?ref= parameter:
- Copy Link → ?ref=android_project_share_copy_link
- Instagram Feed/Stories → android_project_share_instagram_feed / _stories
- X → android_project_share_x
- Facebook Feed/Stories → android_project_share_facebook_feed / _stories
- WhatsApp → android_project_share_whatsapp
- Messages → android_project_share_messages
- Email → android_project_share_email
- More → android_project_share_more

2 - Analytics — confirm a CTA Clicked event fires with `context_cta: share_platform_clicked`, the correct `context_page`, and `context_type` matching the tapped platform.

3 - Dark mode — switch device to dark mode, open the share sheet, and confirm all icons are visible (Copy Link, X, Email, Messages, More should appear white on dark circles).

4 - Email — tap the Email icon and confirm only email apps appear in the chooser (no Instagram, WhatsApp, etc.).
Copy Link — tap Copy Link and paste; confirm the pasted URL is a full Kickstarter URL with reftag, not just ?ref=....

# Story 📖

[DISC-209](https://kickstarter.atlassian.net/browse/DISC-209)

[DISC-209]: https://kickstarter.atlassian.net/browse/DISC-209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/features/socialshare/SocialShareIntentBuilder.kt
+++ b/app/src/main/java/com/kickstarter/features/socialshare/SocialShareIntentBuilder.kt
@@ -170,8 +170,9 @@ object SocialShareIntentBuilder {
 
     /**
      * Email — text always included; image attached as an inline attachment when available.
-     * ACTION_SEND with image/jpeg opens email clients in compose mode with the image
-     * attached and the body/subject pre-filled. Falls back to ACTION_SENDTO + mailto:
+     * ACTION_SEND with message/rfc822 restricts the chooser to email clients only (prevents
+     * social apps like Instagram from appearing) and opens them in compose mode with the
+     * image attached and the body/subject pre-filled. Falls back to ACTION_SENDTO + mailto:
      * (no attachment support) when no image is cached yet.
      */
     private fun emailIntent(shareData: SocialShareData, imageUri: Uri?): Intent {
--- a/app/src/main/java/com/kickstarter/features/socialshare/SocialShareIntentBuilder.kt
+++ b/app/src/main/java/com/kickstarter/features/socialshare/SocialShareIntentBuilder.kt
@@ -179,7 +179,7 @@ object SocialShareIntentBuilder {
         val body = "I thought you'd love this — ${shareData.projectName} by ${shareData.creatorName}:\n\n${shareData.projectUrl}"
         return if (imageUri != null) {
             Intent(Intent.ACTION_SEND).apply {
-                type = "image/jpeg"
+                type = "message/rfc822"
                 putExtra(Intent.EXTRA_SUBJECT, subject)
                 putExtra(Intent.EXTRA_TEXT, body)
                 putExtra(Intent.EXTRA_STREAM, imageUri)
--- a/app/src/main/java/com/kickstarter/features/socialshare/data/SocialShareModels.kt
+++ b/app/src/main/java/com/kickstarter/features/socialshare/data/SocialShareModels.kt
@@ -1,6 +1,8 @@
 package com.kickstarter.features.socialshare.data
 
 import android.net.Uri
+import com.kickstarter.libs.RefTag
+import com.kickstarter.libs.utils.EventContextValues.SharePlatformContextType
 
 data class SocialShareData(
     val projectName: String,
@@ -28,6 +30,32 @@ enum class SocialSharePlatform(val targetPackage: String?) {
         FACEBOOK_STORIES -> true
         else -> false
     }
+}
+
+fun SocialSharePlatform.refTag(): RefTag = when (this) {
+    SocialSharePlatform.COPY_LINK -> RefTag.projectShareCopyLink()
+    SocialSharePlatform.INSTAGRAM_FEED -> RefTag.projectShareInstagramFeed()
+    SocialSharePlatform.INSTAGRAM_STORIES -> RefTag.projectShareInstagramStories()
+    SocialSharePlatform.X -> RefTag.projectShareX()
+    SocialSharePlatform.FACEBOOK_FEED -> RefTag.projectShareFacebookFeed()
+    SocialSharePlatform.FACEBOOK_STORIES -> RefTag.projectShareFacebookStories()
+    SocialSharePlatform.WHATSAPP -> RefTag.projectShareWhatsApp()
+    SocialSharePlatform.MESSAGES -> RefTag.projectShareMessages()
+    SocialSharePlatform.EMAIL -> RefTag.projectShareEmail()
+    SocialSharePlatform.MORE -> RefTag.projectShareMore()
+}
+
+fun SocialSharePlatform.analyticsContextType(): String = when (this) {
+    SocialSharePlatform.COPY_LINK -> SharePlatformContextType.COPY_LINK.contextName
+    SocialSharePlatform.INSTAGRAM_FEED -> SharePlatformContextType.INSTAGRAM_FEED.contextName
+    SocialSharePlatform.INSTAGRAM_STORIES -> SharePlatformContextType.INSTAGRAM_STORIES.contextName
+    SocialSharePlatform.X -> SharePlatformContextType.X.contextName
+    SocialSharePlatform.FACEBOOK_FEED -> SharePlatformContextType.FACEBOOK_FEED.contextName
+    SocialSharePlatform.FACEBOOK_STORIES -> SharePlatformContextType.FACEBOOK_STORIES.contextName
+    SocialSharePlatform.WHATSAPP -> SharePlatformContextType.WHATSAPP.contextName
+    SocialSharePlatform.MESSAGES -> SharePlatformContextType.MESSAGES.contextName
+    SocialSharePlatform.EMAIL -> SharePlatformContextType.EMAIL.contextName
+    SocialSharePlatform.MORE -> SharePlatformContextType.MORE.contextName
 }
 
 data class SocialShareUIState(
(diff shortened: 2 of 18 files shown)


app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedScreen.kt
```diff
@@ -74,6 +76,7 @@ enum class VideoFeedScreenTestTag {
 @Composable
 fun VideoFeedScreen(
     items: List<VideoFeedItem>,
+    environment: Environment,
```
Arkariang, line 79: Passing environment here for the `CompositionLocalProvider` that injects the `SocialShareViewModel` within VideoFeed. Following the same pattern we applied on Search for `LocationSheet`-`FilterMenuViewModel` tandem. 
I could see  though passing instead of the entire environment just what is required, in this specific case just the analytics client. 

Not completely sold on this patter just yet, it works so far though and allows to test as much as needed 🤔.

The two pieces I find problematic for this approach are:
— any composable in the subtree can silently consume the VM, making data flow hard to trace. 
— CompositionLocal has no lifecycle awareness. The ViewModel's lifecycle is tied to the parent activity, if we get to a situation in a hierarchy with several levels of compose screens several using local composition will get complicated to identify the scope owner.

Ideally I think we should create the VM at the screen level (in Activity/Fragment that owns it), and pass only the state and event callbacks down to SocialShareSheet as plain parameters. The sheet doesn't need to know a VM exists — it just needs a SocialShareUIState and a set of lambdas. Let me know what you think about this topic @tonyteate and will try to refactor once I'm back! :)

tonyteate (approved):

app/src/main/java/com/kickstarter/features/socialshare/data/SocialShareModels.kt
```diff
@@ -30,6 +32,32 @@ enum class SocialSharePlatform(val targetPackage: String?) {
...
+    SocialSharePlatform.X -> RefTag.projectShareX()
+    SocialSharePlatform.FACEBOOK_FEED -> RefTag.projectShareFacebookFeed()
+    SocialSharePlatform.FACEBOOK_STORIES -> RefTag.projectShareFacebookStories()
+    SocialSharePlatform.WHATSAPP -> RefTag.projectShareWhatsApp()
+    SocialSharePlatform.MESSAGES -> RefTag.projectShareMessages()
+    SocialSharePlatform.EMAIL -> RefTag.projectShareEmail()
+    SocialSharePlatform.MORE -> RefTag.projectShareMore()
+}
+
+fun SocialSharePlatform.analyticsContextType(): String = when (this) {
```
tonyteate, line 48: nit: Might be clearer to just return `SharePlatformContextType` here and reference `platform.analyticsContextType().contextName` in `AnalyticEvents`.

app/src/main/java/com/kickstarter/features/socialshare/ui/components/SocialSharePlatformGrid.kt
```diff
@@ -107,11 +107,11 @@ private fun PlatformButton(
         Icon(
             imageVector = platform.icon(),
             contentDescription = null,
-            tint = Color.Unspecified,
+            tint = if (platform.isBrandIcon()) Color.Unspecified else colors.socialShare.monochromeIconTint,
```
tonyteate, line 110: Awesome 🕶️

app/src/main/java/com/kickstarter/features/socialshare/SocialShareIntentBuilder.kt
```diff
@@ -170,16 +170,17 @@ object SocialShareIntentBuilder {
...
+     * image attached and the body/subject pre-filled. Falls back to ACTION_SENDTO + mailto:
      * (no attachment support) when no image is cached yet.
      */
     private fun emailIntent(shareData: SocialShareData, imageUri: Uri?): Intent {
         val subject = "Check out this Kickstarter project: ${shareData.projectName}"
         val body = "I thought you'd love this — ${shareData.projectName} by ${shareData.creatorName}:\n\n${shareData.projectUrl}"
         return if (imageUri != null) {
             Intent(Intent.ACTION_SEND).apply {
-                type = "image/jpeg"
+                type = "message/rfc822"
```
tonyteate, line 183: Oh what a great idea! ❤️

app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedScreen.kt
tonyteate: Look forward to chatting more once your back! Early thoughts on this specific implementation:

Overall, I think this is fine. Completely agree that the main issue would be with a deeper hierarchy of screens, but we might not actually run into that problem. I also wonder if we can enforce by convention such VMs existing only at `*Screen`-level Composables.

Will write other comments in the relevant locations.

app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedScreen.kt
```diff
@@ -253,8 +256,10 @@ fun VideoFeedScreen(
             val context = LocalContext.current
             val shareViewModel = remember(data) {
                 SocialShareViewModel(
```
tonyteate, line 258: So yes, we can create and provide this here at the `Screen`. But even in this case, would it be better to just create and provide it from `Activity` level, like with Search?

app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedScreen.kt
tonyteate, line 189: Then, if these values only depend on `project`, `SocialShareData` could also be created within `onShareCTAClick`?

app/src/main/java/com/kickstarter/libs/AnalyticEvents.kt
tonyteate, line 994: Comment mismatch

app/src/main/java/com/kickstarter/features/socialshare/viewmodel/SocialShareViewModel.kt
```diff
@@ -33,14 +37,19 @@ import kotlin.coroutines.EmptyCoroutineContext
  *   - [SocialShareData.creatorName] — displayed in email body copy.
  *   This object is immutable for the lifetime of the ViewModel; one ViewModel instance
  *   corresponds to one sharing session for a specific project.
+ * @param contextPage The screen from which the share sheet was opened, used as the
+ *   analytics context_page property.
  */
 class SocialShareViewModel(
+    private val environment: Environment,
```
tonyteate, line 44: Or maybe we just don't call this a `ViewModel`, get rid of the `Factory` since we don't need it, and pass in `Context` here from which you can retrieve `Environment` 🤔

PR #2522, opened 2026-06-02
PF-188: Move PreLaunchProjectPageScreen's content into a `LazyColumn`

tonyteate:
# 📲 What

Move the main content of `PrelaunchProjectPageScreen` into a `LazyColumn`.

# 🤔 Why

This is a preliminary step to dynamically emitting SDUI-powered Project Story components into the screen, while also preserving the existing components. (Following the strategy from the PoC branch in #2474.)

# 🛠 How

Virtually all of the page content is contained in a secondary `ConstraintLayout`. Wrap this content in a `LazyColumn` by emitting the `ConstraintLayout` and its content via `LazyListScope.item()`. The Similar Projects Carousel is separately emitted so that Project Story components can be emitted above it.

# 👀 See

There should be no detectable visual changes.

# 📋 QA

Visit any Prelaunch Project to confirm that the behavior is the same. Existing tests pass.

# Story 📖

[\[PF-188\] Update layout in PrelaunchProjectPageScreen to support new Project Story component - Jira](https://kickstarter.atlassian.net/browse/PF-188)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/compose/PreLaunchProjectPageScreen.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/compose/PreLaunchProjectPageScreen.kt
@@ -9,9 +9,8 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.systemBarsPadding
-import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.shape.RoundedCornerShape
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kickstarter/ui/activities/compose/PreLaunchProjectPageScreen.kt
tonyteate, line 1: The changes here are now minimal. Originally I was working on pulling the rest of the composables out of the inner `ConstraintLayout`, as some of the constraints depend on properties of `Project` not being `null`. But it now seems likely that this screen will be redesigned or not used in the near future. That said, I may still move the overall component constrained as `buttonCardLayout` into the `Scaffold`'s bottom bar, which would allow us to get rid of the parent `ConstraintLayout`.

Arkariang (approved):

PR #2524, opened 2026-06-03
DISC-210: Trigger App Review Dialog when an user reaches the end of VideoFeed

Arkariang:
# 📲 What

Trigger the Google Play in-app review dialog when a user reaches the last video in the Video Feed and there are no more pages to load.

# 🤔 Why

We want to prompt engaged users for a review at a high-value moment — right after they've consumed all available videos in the feed — increasing the likelihood of a positive rating.

# 🛠 How

- `VideoFeedUIState` now carries a hasMore: Boolean flag, populated from the API's `pageInfo.hasNextPage` response.
- `VideoFeedScreen` accepts two new parameters: `hasMore` and `onReachedLastVideo`. A LaunchedEffect keyed on settledPage, items.size, and hasMore fires onReachedLastVideo exactly once per session when the user lands on the last page and hasMore is false. A local hasTriggeredReview guard prevents re-firing if the effect re-evaluates.
VideoFeedActivity passes hasMore from the UI state and calls showRatingDialogWidget() in the onReachedLastVideo callback.

- `showRatingDialogWidget (ActivityExt)` was improved:
In DEBUG builds, shows a native AlertDialog as a stand-in so the flow can be verified without a Play Store connection. In production, the review flow is unchanged but cleaned up.

|  |  |

# 📋 QA

- Build a debug from Android Studio.
- Open the Video Feed.
- Scroll through all videos until reaching the end of the feed (no more pages load).
- A dialog titled "In-App Review [DEBUG]" should appear exactly once. Dismiss and scroll up again, down again— the dialog should not re-appear.
For a production build: cannot really test the Production piece just yet! 

# Story 📖

[DISC-210](https://kickstarter.atlassian.net/browse/DISC-210)

[DISC-210]: https://kickstarter.atlassian.net/browse/DISC-210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/java/com/kickstarter/ui/extensions/ActivityExt.kt
```diff
@@ -133,26 +134,29 @@ fun Activity.setUpConnectivityStatusCheck(lifecycle: Lifecycle) {
 }
 
 fun Activity.showRatingDialogWidget() {
+    if (BuildConfig.DEBUG) {
```
Arkariang, line 137: This piece is purely to help QA, this dialog will trigger only on debug builds :)

app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedScreen.kt
```diff
@@ -110,6 +113,15 @@ fun VideoFeedScreen(
         }
     }
 
+    // - In-app review: fire once when the user settles on the last video and there are no more pages.
+    LaunchedEffect(pagerState.settledPage, items.size, hasMore) {
+        if (!hasTriggeredReview && items.isNotEmpty() && !hasMore && pagerState.settledPage == items.size - 1) {
+//            Timber.d("VideoFeedScreen: ratingDialog triggering — user reached last video (page=${pagerState.settledPage})")
+            hasTriggeredReview = true
```
tonyteate, line 120: partial nit: Would it be clearer if we elevated this state `hasTriggeredReview` to where `showRatingDialogWidget()` happens? In the current local context, a developer just sees `onReachedLastVideo()`, which makes complete sense, but they don't immediately know the effect of that function.

tonyteate (approved):

tonyteate:
Opinion: The UX feels a bit urgent and unexpected. I wonder if the dialog could maybe be delayed a little for example.

Arkariang:
> Opinion: The UX feels a bit urgent and unexpected. I wonder if the dialog could maybe be delayed a little for example.

The real dialog is a bottom nav -> https://developer.android.com/guide/playcore/in-app-review when appears live it has a nice animation etc, the alert dialog is indeed intrusive but is more for QA than anything else, I could add a few seconds delay though 🤔

tonyteate:
> The real dialog is a bottom nav -> https://developer.android.com/guide/playcore/in-app-review when appears live it has a nice animation etc, the alert dialog is indeed intrusive but is more for QA than anything else, I could add a few seconds delay though 🤔

That new experience does look much smoother for sure. I've only seen the 5-star review dialog that pops up after pledging to certain projects.

PR #2526, opened 2026-06-08
[No-jira]: SearchScreen Compose previews rendering errors fixed!

Arkariang:
# 📲 What

Fix rendering of Compose Previews for the Search screen and correct a layout issue that caused the full-screen loading indicator to not render properly.

# 🤔 Why

The PagerPreview, SearchScreenPreviewNonEmpty, and SearchScreenPreviewEmpty preview functions were crashing at render time because FilterMenuViewModel — required via LocalFilterMenuViewModel — was not being provided in the preview context. Additionally, the SearchScreen composable had a layout issue where the full-screen loading overlay was placed outside the inner Box, causing incorrect rendering relative to the scaffold padding.

# 🛠 How

- Preview fixes: Each preview now instantiates a minimal Environment backed by MockApolloClientV2 and an anonymous CurrentUserTypeV2, creates a FilterMenuViewModel from it, and wraps the preview content in CompositionLocalProvider(LocalFilterMenuViewModel provides fakeViewModel).

- Layout fix: Wrapped the LazyColumn/empty view and the loading overlay in a shared Box(modifier = Modifier.fillMaxSize().padding(padding)), making the loading spinner a sibling that correctly overlays the full content area. Padding is now applied once at the Box level rather than duplicated on the LazyColumn.

- Compatibility fix: Changed FilterPages.entries.size to FilterPages.values().size for the pager page count to use the stable API. 

- Preview mutableListOf state holders wrapped in remember { } to survive recomposition correctly.

# 👀 See
- No user facing changes
|  |  |

# 📋 QA

- Open `SearchScreen.kt` in Android Studio and verify all three @Preview functions (PagerPreview, SearchScreenPreviewNonEmpty, SearchScreenPreviewEmpty) render without errors in the Preview panel.
- Trigger a loading state (e.g. load next page or set airplane mode) and verify the loading spinner appears centered and fills the screen correctly.
- Verify that applying filters, viewing results, and the empty state all render as expected with no visual regressions.

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
@@ -120,7 +120,7 @@ fun PagerPreview() {
 
     KSTheme {
         CompositionLocalProvider(LocalFilterMenuViewModel provides fakeViewModel) {
-            val testPagerState = rememberPagerState(initialPage = 0, pageCount = { FilterPages.values().size })
+            val testPagerState = rememberPagerState(initialPage = 0, pageCount = { FilterPages.entries.size })
             val testSheetState = rememberModalBottomSheetState(
                 skipPartiallyExpanded = true
             )
@@ -470,7 +470,7 @@ fun SearchScreen(
     val currentRecommended = remember { mutableStateOf<Boolean>(false) }
 
     val currentGoal = remember { mutableStateOf<DiscoveryParams.GoalBuckets?>(null) }
-    val pagerState = rememberPagerState(initialPage = 0, pageCount = { FilterPages.values().size })
+    val pagerState = rememberPagerState(initialPage = 0, pageCount = { FilterPages.entries.size })
 
     val activeBottomSheet = remember {
         mutableStateOf<FilterRowPillType?>(null)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
```diff
@@ -413,7 +470,7 @@ fun SearchScreen(
     val currentRecommended = remember { mutableStateOf<Boolean>(false) }
 
     val currentGoal = remember { mutableStateOf<DiscoveryParams.GoalBuckets?>(null) }
-    val pagerState = rememberPagerState(initialPage = 0, pageCount = { FilterPages.entries.size })
+    val pagerState = rememberPagerState(initialPage = 0, pageCount = { FilterPages.values().size })
```
tonyteate: Do we want `values()` here instead of `entries`?

tonyteate (dismissed):

app/src/main/java/com/kickstarter/ui/activities/compose/search/SearchScreen.kt
Arkariang: nop! reverted!

tonyteate (approved):

PR #2528, opened 2026-06-15
DISC-236: Pagination loading state UI for the Video Feed

Arkariang:
# 📲 What

Surfaces the pagination loading state in the Video Feed, and prevents redundant feed loads.

- Adds a full-screen loading page (a centered spinner over the video background) that appears as a trailing page in the feed while the next page of videos is being fetched.
- Moves the initial `loadVideoFeed()` call into a `LaunchedEffect` so it no longer re-runs on every recomposition.
- `KSCircularProgressIndicator` now accepts an optional `color` parameter so the spinner can match the video player content color.

# 🤔 Why

The `isLoading` flag already existed in `VideoFeedUIState`, but it was never wired to any UI — so pagination gave the user no visual feedback while waiting for more videos to load.

Separately, `loadVideoFeed()` was being invoked directly in the composable body, which caused it to re-run on every recomposition (e.g. whenever items emit), triggering more network calls than necessary.

# 🛠 How

- **`VideoFeedScreen`**: Appends an extra trailing page to the pager when `isLoading && hasMore`. A sentinel key (`Long.MIN_VALUE`) keeps the pager key stable for the loading page without colliding with real (always positive) project ids. The new `VideoFeedLoadingPage` composable renders a centered `KSCircularProgressIndicator` over the player background. Thanks to prefetch + `beyondViewportPageCount = 1`, the next page usually loads before the user reaches the end, so the loading page is typically only seen when the user out-swipes the request, or on the very first load.
- **`VideoFeedActivity`**: Wraps the initial `loadVideoFeed()` in `LaunchedEffect(Unit)` and passes `uiState.isLoading` into `VideoFeedScreen`.
- **`KSCircularProgressIndicator`**: New optional `color` param defaulting to `kds_create_700`, so existing call sites are unaffected.
- **Tests**: Screen tests cover the loading-page visibility rules (shown when `isLoading && hasMore`, hidden when `isLoading` is false or there are no more pages). View-model tests assert `isLoading` toggles around the fetch and that a second `loadVideoFeed()` call while one is already in flight is a no-op.

# 📋 QA

1. Open the Video Feed.
2. Swipe through the videos to the end of the currently loaded page.
4. Confirm no spinner page appears once there are no more videos to load (`hasMore = false`).
5. Confirm the feed still loads correctly on first open, and that scrolling / recomposition does not trigger duplicate network loads.

Worth noticing the pagination loading state as of now is hard to trigger as the pre-fecth happens on item 7, preparing and rendering 3 items in advance, will be visible on very first load and on new page reached only on very poor quality networks

# Story 📖

[DISC-236]: (https://kickstarter.atlassian.net/browse/DISC-236)

### DIFF
--- a/app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedActivity.kt
@@ -8,6 +8,7 @@ import androidx.activity.compose.setContent
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.activity.viewModels
 import androidx.compose.material3.SnackbarHostState
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
@@ -55,7 +56,11 @@ class VideoFeedActivity : ComponentActivity() {
                 val errorAction = setUpVideoFeedErrorActions(snackbarHostState)
                 viewModel.provideErrorAction { message -> errorAction.invoke(message) }
 
-                viewModel.loadVideoFeed()
+                // Initial load must live within LaunchedEffect, calling it directly in the
+                // composable body re-runs it on every recomposition (e.g. when items emit)
+                LaunchedEffect(Unit) {
+                    viewModel.loadVideoFeed()
+                }
 
                 val uiState by viewModel.videoFeedUIState.collectAsStateWithLifecycle()
                 VideoFeedScreen(
@@ -63,7 +68,9 @@ class VideoFeedActivity : ComponentActivity() {
                     environment = env,
                     errorSnackBarHostState = snackbarHostState,
                     hasMore = uiState.hasMore,
-                    onLoadMore = { viewModel.loadVideoFeed() },
+                    onLoadMore = {
+                        viewModel.loadVideoFeed()
+                    },
                     onReachedLastVideo = {
                         showRatingDialogWidget()
                     },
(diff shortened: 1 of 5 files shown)


tonyteate:
Nice! Logic looks good. Couple of UI/UX nits:

- Do we want the surface and loading indicator color to be dark gray + white in both light and dark modes?
  - Do we want the gray to actually be gray or black?
- Is it possible to keep the Close button visible while the loading page is displayed? If a user had a slow internet connection or was going to run into a timeout exception after e.g. 10-30 seconds, it would give them a visual option to leave (they can always back out of course).
  - Separately from this PR, but visible in the video, is it possible to keep the Close button visible when there's just an error?

https://github.com/user-attachments/assets/4ea94629-447a-4e44-9360-0d94876a2b96

tonyteate (approved):
💫

Arkariang:
> Nice! Logic looks good. Couple of UI/UX nits:
> 
> * Do we want the surface and loading indicator color to be dark gray + white in both light and dark modes?
>   
>   * Do we want the gray to actually be gray or black?
> * Is it possible to keep the Close button visible while the loading page is displayed? If a user had a slow internet connection or was going to run into a timeout exception after e.g. 10-30 seconds, it would give them a visual option to leave (they can always back out of course).
>   
>   * Separately from this PR, but visible in the video, is it possible to keep the Close button visible when there's just an error?
> 
>  Screen_recording_20260615_171508.mp4

The user will not have time to interact with the X button if we add it to the loading indicator. As for the theme, VideoFeed loads the same set of colors on both light and dark, using the same palette for the loading indicator here. Also an fyi this will likely be reviewed on follow up phases once the real chances of that loading indicator to be shown are bigger.

PR #2529, opened 2026-06-15
DISC-238: Swiping quickly through the video feed showed several black screens

Arkariang:
# 📲 What

Two improvements to the Video Feed playback experience:

Poster / preview image on the video player. Each VideoFeedItem now carries a previewImageUrl. KSVideoPlayer renders that poster (center-cropped) over the video surface while the HLS stream buffers, then fades it out the moment the player draws its first frame. The query was already prepared for this change thought the UI was not consuming it. 
Initial feed load moved into the ViewModel. The first page is now loaded from VideoFeedViewModel.init {} instead of from a LaunchedEffect in VideoFeedActivity.

# 🤔 Why

When swiping between videos (especially on fast swipes or slow networks), the player surface showed a black flash while the next HLS stream buffered, which looked broken. Painting the poster over the surface until the first frame is ready makes the transition feel instant and seamless.
The initial loadVideoFeed() call lived in the composable layer. Owning it in the ViewModel guarantees it runs exactly once and isn't re-triggered when the Activity's Compose content is recreated, avoiding redundant feed fetches.

# 🛠 How

- `VideoFeedItem` gains a nullable `previewImageUrl`, populated in VideoFeedTransformers from `verticalVideo.previewImageUrl`.
- VideoFeedScreen passes item.previewImageUrl down to KSVideoPlayer.

KSVideoPlayer:
- Tracks a firstFrameRendered flag, set via a new Player.Listener.onRenderedFirstFrame() override and reset when the player instance changes.
- Shows the poster in an AnimatedVisibility (Coil AsyncImage, ContentScale.Crop, fadeOut exit) while !firstFrameRendered && !previewImageUrl.isNullOrEmpty(), tagged VIDEO_PLAYER_POSTER.

- VideoFeedActivity drops the LaunchedEffect(Unit) { viewModel.loadVideoFeed() }; VideoFeedViewModel calls loadVideoFeed() in its init {} block.

# 👀 See

https://github.com/user-attachments/assets/ee0e308f-9837-417d-b00d-58fa0749e91f

|  |  |

# 📋 QA

- Load videofeed, ideally with poor network conditions, and do scroll over the feed. Observe a lack of black screens before the video starts loading.

# Story 📖

[DISC-238](https://kickstarter.atlassian.net/browse/DISC-238)

[DISC-238]: https://kickstarter.atlassian.net/browse/DISC-238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/features/videofeed/data/VideoFeedModels.kt
+++ b/app/src/main/java/com/kickstarter/features/videofeed/data/VideoFeedModels.kt
@@ -21,6 +21,7 @@ data class VideoFeedItem(
     val badges: List<KSVideoBadgeType>,
     val project: Project,
     val hlsUrl: String?,
+    val previewImageUrl: String? = null,
     val videoId: Long = 0
 )
 
--- a/app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedActivity.kt
@@ -8,6 +8,7 @@ import androidx.activity.compose.setContent
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.activity.viewModels
 import androidx.compose.material3.SnackbarHostState
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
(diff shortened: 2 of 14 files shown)


tonyteate:
Would it be worth preloading them? Definitely not necessary, since the images load pretty quickly on normal internet connections, but on slower connections users might still black screens.

tonyteate (approved):
Cool!

Arkariang:
> Would it be worth preloading them? Definitely not necessary, since the images load pretty quickly on normal internet connections, but on slower connections users might still black screens.

You're totally right, with slow connections there still migth be black screens, but with the pager option beyondViewportPageCount = 1, 3 pages are preloaded each time (same as the player itself), so there is some level of preloading already in place, the expectation is as of now not to get rid completely of the black screens in absolutely all instances but to reduce it as much as possible without adding too much complexity

PR #2530, opened 2026-06-16
DISC-259: Blurry image on Entry card

Arkariang:
# 📲 What

Improves the image quality of the Discovery Video Feed entry-point banner (`entrypoint_banner`). Replaces the blurry, low-resolution PNG with density-specific WebP assets and lets the image render at its correct, higher intrinsic size.

# 🤔 Why

The banner image was a single 66×96px PNG living in the density-less `res/drawable/` folder. Android treats that as the mdpi (1×) baseline, so on
every higher-density device (xhdpi/xxhdpi/xxxhdpi — i.e. most phones)

# 🛠 How

- Re-exported the full banner artwork from Figma (composed graphic: fanned background cards + video thumbnail + play button) at high resolution.
- Generated density-specific WebP assets for every bucket to keep file sizes small while preserving the transparent card edges using Android Studio "Convert to WebP".

# 👀 See

VideoFeedCard

|  |  |

# 📋 QA

- Open Discovery and Search with several density devices to check none of them has a blurry image anymore.

# Story 📖

[DISC-259](https://kickstarter.atlassian.net/browse/DISC-259)

[DISC-259]: https://kickstarter.atlassian.net/browse/DISC-259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonyteate (approved):

app/src/main/java/com/kickstarter/ui/compose/designsystem/KSBanners.kt
```diff
@@ -120,7 +119,6 @@ fun KSVideoFeedBanner(
             painter = painterResource(id = R.drawable.entrypoint_banner),
             contentDescription = null,
             modifier = Modifier
-                .size(width = dimensions.discoveryBannerImageWidth, height = dimensions.discoveryBannerImageHeight)
```
tonyteate, line 123: You might still want something like this, or different spacing between the text, if you want the top and bottom of all of the content to be aligned

Arkariang: Yeah, it's giving me a bit of OCD XD, Figma is not entirely completely aligned either, it is 2px smaller top & bottom, will ping tae for feedback once the new build is ready
Screenshot 2026-06-16 at 12 57 30 PM

PR #2538, opened 2026-06-23
DISC-258: Inconsistencies on play button

Arkariang:
# 📲 What

Fixes the related glitches with the VideoFeed play/pause controls overlay, where the play-button icon got out of sync with the video's actual playback state:

- The play button lingering on screen over a video that is actually playing (specially when navigating away either snapping to another page or navigating to another activity).
- The play button appearing intermittently when the user only meant to tap or drag the progress bar.

# 🛠 How

- `KSVideoPlayer.kt`: In the LaunchedEffect(isActive, isAppInForeground) that resumes playback, reset showControls = false when the player auto-resumes. This covers both the swipe-away/back and the leave-screen/return cases, keeping the overlay in sync with the forced playback.
- KSProgressIndicators.kt (KSVideoScrubBar) — Consume the down event immediately on awaitFirstDown(). Since Compose's clickable uses awaitFirstDown(requireUnconsumed = true), claiming the gesture up front stops a tap or short drag on the scrub bar from propagating to the surface click handler and toggling the controls.
- Added tests for previous use cases

# 📋 QA

Open the VideoFeed.
- Swipe scenario: Pause a video (play button shows) → swipe up to the next video → swipe back. Play button should not appear
- Navigation scenario: Pause a video → tap into Creator Bio or the project page → return to the feed. Play button should not appear
- Progress bar scenario: While a video is playing, tap and also drag (fast-forward/rewind) on the progress bar several times. The play button never appears.

# Story 📖

[DISC-258](https://kickstarter.atlassian.net/browse/DISC-258)

[DISC-258]: https://kickstarter.atlassian.net/browse/DISC-258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/compose/designsystem/KSProgressIndicators.kt
+++ b/app/src/main/java/com/kickstarter/ui/compose/designsystem/KSProgressIndicators.kt
@@ -398,6 +398,7 @@ fun KSVideoScrubBar(
                 awaitEachGesture {
                     // - detect touch down for scale animation
                     val down = awaitFirstDown()
+                    down.consume()
                     isDragging = true
                     onScrubStart()
                     dragProgress = (down.position.x / size.width).coerceIn(0f, 1f)
--- a/app/src/main/java/com/kickstarter/ui/compose/designsystem/videoplayer/KSVideoPlayer.kt
+++ b/app/src/main/java/com/kickstarter/ui/compose/designsystem/videoplayer/KSVideoPlayer.kt
@@ -195,6 +195,11 @@ fun KSVideoPlayer(
         val shouldPlay = isActive && isAppInForeground
         exoPlayer.playWhenReady = shouldPlay
         if (shouldPlay) {
+            // The player auto-resumes when the page becomes active or the app returns to the
+            // foreground, overriding any earlier manual pause. Reset the controls overlay so the
+            // play button doesn't linger on screen while the video is actually playing
+            // (e.g. pause → swipe away → swipe back, or pause → leave the screen → return).
+            showControls = false
             while (true) {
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/kickstarter/ui/compose/designsystem/videoplayer/KSVideoPlayer.kt
```diff
@@ -195,6 +195,7 @@ fun KSVideoPlayer(
         val shouldPlay = isActive && isAppInForeground
         exoPlayer.playWhenReady = shouldPlay
         if (shouldPlay) {
+            showControls = false
```
Arkariang, line 198: The player auto-resumes when the page becomes active or the app returns to the
foreground, overriding any earlier manual pause. Reset the controls overlay so the play button doesn't linger on screen while the video is actually playing

app/src/main/java/com/kickstarter/ui/compose/designsystem/KSProgressIndicators.kt
```diff
@@ -398,6 +398,7 @@ fun KSVideoScrubBar(
                 awaitEachGesture {
                     // - detect touch down for scale animation
                     val down = awaitFirstDown()
+                    down.consume()
```
Arkariang, line 401: `awaitFirstDown()` detects the initial touch.
`down.consume()` ensures the event is only used and consumed within the container being dragged

tonyteate (approved):
Nice

PR #2540, opened 2026-06-24
[No-jira]: playback error telemetry + retry

Arkariang:
# 📲 What

Adds observability for **playback failures**, plus **bounded self-recovery** for the on-screen video.

# 🤔 Why

We have no visibility into how reliably feed videos actually play, on which devices, or whether the hardware-decoder budget is a real problem on lower-end hardware.

This telemetry gives us sata so future playback-performance decisions can be data-driven instead of speculative. 

The recovery piece turns the most common failures (the OS reclaiming our decoder under pressure) from a permanently frozen frame into an automatic resume for the user.

# 🛠 How

- `KSVideoPlayer` now reports `ExoPlayer` `onPlayerError`s via a new optional `onPlaybackError(error, isActive)` callback, threaded through `VideoFeedScreen` → `VideoFeedActivity` → `VideoFeedViewModel`.
- `VideoFeedViewModel` records a **Crashlytics non-fatal** for each failure, tagged with `video_feed_video_id`, `video_feed_error_code`, `video_feed_is_active`, and `video_feed_position`.
- When the error is **recoverable** (a reclaimed hardware decoder or a transient network failure) **and** it's the video currently on screen, the player re-prepares to resume — bounded so it can never loop.

# 👀 See

No UI change; the observable surface is Crashlytics (non-fatals filterable by the new custom keys).

|  |  |

# 📋 QA
1. - Everything should be working as usual in terms of user experience.
2. **Telemetry:** in a debug build, force an error (e.g. point a feed item's HLS URL at an invalid/expired source) and confirm a Crashlytics non-fatal is recorded with keys `video_feed_error_code`, `video_feed_is_active`, `video_feed_video_id`, `video_feed_position` (or the Timber fallback warning in logcat).

### DIFF
--- a/app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/videofeed/ui/VideoFeedActivity.kt
@@ -113,6 +113,9 @@ class VideoFeedActivity : ComponentActivity() {
                     },
                     onShareCTAClick = { project ->
                         viewModel.onCTAClicked(project, CtaContextName.VIDEO_SHARE)
+                    },
+                    onVideoPlaybackError = { videoFeedItem, position, error, isActive ->
+                        viewModel.onVideoPlaybackError(videoFeedItem, position, error, isActive)
                     }
                 )
             }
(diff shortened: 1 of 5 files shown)


tonyteate (approved):
Lgtm - tried with invalid hls url and added logging, just hasn't shown up in crashlytics yet I assume.

PR #2546, opened 2026-07-13
DISC-403: Statsig Feature Gates able to be enabled for Admin users

Arkariang:
# 📲 What

Extends Statsig identity propagation so feature gates can target Kickstarter admins. When a logged-in user is observed. The android_video_feed gate is extended with a rule requiring admin status.

# 🤔 Why

Lets us gate features to admin users straight from the Statsig console (staged rollout / internal validation) with no app release. Starting with the Video Feed entry points.

# 🛠 How

- `observeUserAndFetchConfigs` now carries User.isAdmin() through a private ObservedUserData, so distinctUntilChanged re-fires when admin status changes.
- `handleObservedUserData` sets `StatsigUser.privateAttributes = {"is_ksr_admin": <bool>}` for logged-in users. 
- privateAttributes (vs custom) keeps the value usable for gate evaluation but excludes it from Statsig exposure logs

# 👀 See

https://github.com/user-attachments/assets/b88c2885-fa89-4384-a00d-2ddc91a553e9

| --- | --- |
|  |  |

# 📋 QA

- As logged out user, the entry cards to VideoFeed are not visible now, the gate has been extended with a new rule, where the user needs to be an admin user.
- Once you log-in with an admin user, either navigate to Search or kill the app and open discovery, you will see the entry card available again.

# Story 📖

[DISC-403](https://kickstarter.atlassian.net/browse/DISC-403)

[DISC-403]: https://kickstarter.atlassian.net/browse/DISC-403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
+++ b/app/src/main/java/com/kickstarter/libs/featureflag/StatsigClient.kt
@@ -92,7 +92,7 @@ interface StatsigClientType {
 
     val statsigUser: StateFlow<StatsigUser>
 
-    suspend fun handleObservedUserData(stableId: String?, segmentAnonymousId: String?, userId: String?)
+    suspend fun handleObservedUserData(stableId: String?, segmentAnonymousId: String?, userId: String?, isAdmin: Boolean)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


tonyteate (approved):
Lgtm 👍

PR #2550, opened 2026-07-22
PF-408: Improve `RichTextItemPhotoComponent` image loading and handling

tonyteate:
# 📲 What

Fast-follow optimizations to fix Compose rendering bugs in Pre-launch Project story as reported in Firebase. Specifically,

- Stop `pin`-ing the `RichTextItemPhotoComponent`
- When the image for `ProjectStoryCaptionedImage`'s loads successfully, capture the aspect ratio and use it to render a placeholder for subsequent loads.

# 🤔 Why

For many Projects, the campaign story is composed heavily of images. In the current implementation, we keep all successfully loaded images in the composition to prevent surrounding content from shifting around when scrolling. However, given the memory limits Android imposes on each app, the Project Story experience sometimes suffers from performance related issues, reflected in two notable exceptions logged in Firebase ([1](https://console.firebase.google.com/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/788308c5f327bbb6afbab570cd424ba8?time=7d&types=crash&sessionEventKey=6A60AC2503A700011C664AC37F0DD272_2243794871889421666), [2](https://console.firebase.google.com/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/159e25351be7042db517f3af684684db?time=7d&types=crash&sessionEventKey=6A6114DE00160001420883751327BE5C_2243885395847163487)) (lesser reported OOM exceptions have been logged as well in internal testing).

# 🛠 How

Within `RichTextItemPhotoComponent`, we no longer call `PinnableContainer.pin()` on a successful image load, allowing images to be disposed. Instead we determine the aspect ratio of the loaded image, cache it in a state map, and pass it back down to `ProjectStoryCaptionedImage` for subsequent loads.

When provided, `ProjectStoryCaptionedImage` uses the aspect ratio to render a placeholder Box (the image's containing box itself was an sensible choice here) that simulates a crossfade on subsequent image loads, preserving the space required by the image as the user scrolls around.

# 👀 See

> [!NOTE]
> Recorded with a hardcoded `delay(1000L)` in the global `ImageLoader` interceptor chain

https://github.com/user-attachments/assets/a12084cf-3554-4652-9e6a-dd24ab00dfa9

# 📋 QA

Open any Pre-launch Project Page and scroll around. Once images have loaded, confirm that the surrounding content doesn't shift on subsequent loads.

# Story 📖

[\[PF-408\] Compose UI IllegalStateException in Pre-launch Activity - Jira](https://kickstarter.atlassian.net/browse/PF-408)

### DIFF
--- a/app/src/main/java/com/kickstarter/ui/activities/compose/PreLaunchProjectPageScreen.kt
+++ b/app/src/main/java/com/kickstarter/ui/activities/compose/PreLaunchProjectPageScreen.kt
@@ -47,6 +47,7 @@ import com.kickstarter.features.projectstory.ui.RichTextItemPhotoComponent
 import com.kickstarter.features.projectstory.ui.RichTextItemTextComponent
 import com.kickstarter.features.projectstory.ui.WebViewComponent
 import com.kickstarter.libs.utils.extensions.toHtml
+import com.kickstarter.libs.utils.extensions.toSha256
 import com.kickstarter.mock.factories.ProjectFactory
 import com.kickstarter.models.Project
 import com.kickstarter.ui.activities.compose.PreLaunchProjectPageScreenTestTag.COMING_SOON_BADGE
@@ -283,7 +284,7 @@ fun PreLaunchProjectPageScreen(
                     }
                 }
 
-                itemsIndexed(story?.items ?: emptyList(), key = { index, item -> "$item$index" }, contentType = { index, item -> item.lazyListContentType() }) { index, item ->
+                itemsIndexed(story?.items ?: emptyList(), key = { index, item -> item.lazyListKey(index) }, contentType = { index, item -> item.lazyListContentType() }) { index, item ->
                     Box(
                         modifier = Modifier
                             .padding(horizontal = screenPadding, vertical = dimensionResource(id = R.dimen.grid_1))
@@ -439,3 +440,6 @@ private fun RichTextItem.lazyListContentType(): String =
     } else {
         this::class.simpleName.orEmpty()
     }
+
+private fun RichTextItem.lazyListKey(index: Int): String =
+    "${this.toString().toSha256()}_$index"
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kickstarter/ui/activities/compose/PreLaunchProjectPageScreen.kt
```diff
@@ -413,3 +433,13 @@ fun PreLaunchProjectPageScreen(
...
+
+private fun RichTextItem.lazyListContentType(): String =
+    if (this is RichTextItem.Text.Paragraph && hasChildPhoto()) {
+        RichTextItem.Photo::class.simpleName.orEmpty()
+    } else {
+        this::class.simpleName.orEmpty()
+    }
+
+private fun RichTextItem.lazyListKey(index: Int): String =
+    "${this.toString().toSha256()}_$index"
```
tonyteate, line 445: This `key` based on hash + index is intentional, as it's a slight improvement over the default of just the index. We do not have a stable server-provided value to use, and all the `RichTextItem`s are immutable `data class`es. So, all things considered this should be a good enough indicator for now, and I explicitly want to avoid [this exception](https://console.firebase.google.com/project/android-external-release/crashlytics/app/android:com.kickstarter.kickstarter/issues/159e25351be7042db517f3af684684db?time=7d&types=crash&sessionEventKey=6A6114DE00160001420883751327BE5C_2243885395847163487) (though I believe that one is also a result of rendering performance).

Arkariang: Slightly worried about a thing here, the key is used so Compose can preserve item state/animations when the list changes. Could happend that by appending _$index makes the key position-dependent, if an item ever shifts position, its key changes and its state is dropped? 🤔

Arkariang (approved):
LGTM!

app/src/main/java/com/kickstarter/ui/activities/compose/PreLaunchProjectPageScreen.kt
tonyteate: Yes, intentionally in this special case. The default is the position-dependent index,  and I'm just adding on to that primarily to ensure the exception is no longer thrown. 

Since we do not currently have an alternative stable ID, we still want to indicate the need for key, but we don't want to risk two items having the same key since that's currently possible. We can accept items recomposing if the list changes for now, but note that all of the composed content in this list is effectively stable.

PR #2551, opened 2026-07-23
DISC-425: Discovery VideoFeed entry card edge cases

Arkariang:
# 📲 What

Introduces a `configReady` gate on `StatsigClient` that reflects when gate/config values have settled for the current Statsig user, and switches the two Video Feed Banner consumers (Discovery and Search) to gate their reads on it instead of the one-shot isReady (SDK initialized). While values are (re)loading — or come back Unrecognized — the banner keeps its current state instead of flipping to hidden.

# 🤔 Why

`isReady` only reflects the one-time SDK initialization and stays true across user changes. When a user logs in/out, Statsig runs an async `updateUser` to refetch values. Reading the ANDROID_VIDEO_FEED gate mid-reload returns the default which caused the banner would flicker to hidden during the reload window

*Note: `Loading:Unrecognized` meands the SDK has no settled values yet for the current user, so it hands back the type default (false for a boolean gate) and tags it Unrecognized.

Screenshot 2026-07-20 at 5 17 30 PM

# 🛠 How

- StatsigClient: Added configReady: StateFlow<Boolean>. It flips to false at the start of the handleObservedUserData → updateUser refetch and back to true once values have settled. This is deliberately distinct from isReady.

- SearchAndFilterViewModel: Now collects configReady (instead of isReady), skips while not ready, and skips when getFeatureGate(...).getEvalDetails().reason == EvalReason.Unrecognized, only updating visibility on a recognized value.

- DiscoveryFragmentViewModel: Renamed isVideoFeedBannerVisible → videoFeedBannerVisibility, now returning a nullable Boolean? — null means "not yet decidable" so callers keep the current banner state. Also removed the force-hide of the banner on clearPage (which was another source of the phantom flip/exposure).

- Test mock (MockStatsigClient): Added beginUserReload() / completeUserReload() to simulate the reload window (configReady = false, gate reads return Loading:Unrecognized/false), and triggerReady() now also settles configReady.

|  |  |

# 📋 QA
Compare with master branch or the released app: 
- - [Master branch behaviour] Start the app as logged out user, you see the banner in discovery, log in, when Discovery is presented again the banner has disappeared. 
- - [Current branch behaviour] Start the app as logged out user, you see the banner in discovery, log in, when Discovery is presented again the banner persitst. Note* the gate value will persist if it were false and no banner visible that is the expected outcome ofter logging in.

- - [Master branch behaviour] Start the app as logged in user, you see the banner in discovery, log out, when Discovery is presented again the banner has disappeared. 
- - [Current branch behaviour] Start the app as logged in user, you see the banner in discovery, log out, when Discovery is presented again the banner is still present.

# Story 📖

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -357,7 +357,7 @@ dependencies {
     coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.1.5'
 
     // Statsig
-    implementation 'com.statsig:android-sdk:5.1.0'
+    implementation 'com.statsig:android-sdk:5.1.3'
 }
 
 // SHA and timestamp caching courtesy of https://github.com/gdg-x/frisbee/blob/develop/app/build.gradle#L193-L218
--- a/app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
+++ b/app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
@@ -12,6 +12,7 @@ import com.kickstarter.models.Category
 import com.kickstarter.models.Location
 import com.kickstarter.models.Project
 import com.kickstarter.services.DiscoveryParams
+import com.statsig.androidsdk.EvalReason
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.flow.MutableStateFlow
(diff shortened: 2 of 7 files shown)


tonyteate:
Interestingly, this all comes down to a somewhat surprising caching issue. If we set a stable cache key via `customCacheKey` to e.g. `"$sdkKey:$stableId"`, you should no longer see such an issue. Rather than `Loading:Unrecognized` you will see `Cache:Recognized` as expected.

Relatedly, we should bump the version to [5.1.3](https://github.com/statsig-io/android-sdk/releases/tag/5.1.3) to fix a race condition with these calls, [specifically](https://github.com/statsig-io/android-sdk/compare/5.1.2...5.1.3#diff-abf312dceb8637b01f19869b4991c1897be9472db4cfc0f68bc593b199dfa062R279) related to the fact that we’re supposed to be able to call e.g. `checkGate` even while `updateUser()` is in flight; just not before `initialize()` has returned.

This is a nice solution in the mean-time to effectively bring the expected behavior 💯 

(P.S. It's worth noting that with no changes, if a user closes and opens the app they will continue to see the banner as well. This is strictly a caching & timing issue.)

tonyteate (dismissed):
Nice 👍 

Let's not forget to bump the library to 5.1.3

tonyteate (approved):

PR #2552, opened 2026-07-28
DISC-386 && Disc-388 && Disc-392: Open Calls filter Network layer

Arkariang:
📲 What
Adds the network + ViewModel + feature-gate foundation for the Open Calls (Creative Prompt) search filter:

- New `GetCreativePromptTags` GraphQL query and a `Tag` model (id/name/url/slug).
- `KSApolloClientV2.getTags()` to fetch Creative Prompt tags.
- `tagId` threaded into the `FetchProjects` query and `DiscoveryParams`, so a selected tag scopes search results.
- `FilterMenuViewModel.getTags()` exposes the tag list via FilterMenuUIState.tagsList, loaded on screen entry.
- `SearchAndFilterViewModel.isOpenCallsEnabled`, gated on the new `android_open_calls` Statsig gate.

🤔 Why
Open Calls lets backers filter search to projects participating in a Creative Prompt (Make 100, Zine Quest, etc.). This is the first of the chained PRs — it lands the data/gate layer so the UI PR can build on top without a giant single diff.

🛠 How
- `tags.graphql`: new query scoped to CREATIVE_PROMPT. Tag.id is a Long decoded from the Relay ID!; converted with id.toInt() at the DiscoveryParams.tagId (Int) boundary.
- `project.graphql`: added $tagId: Int variable, passed to projects(tagId:).
- ViewModel: getTags() fetches on SearchAndFilterActivity entry and emits into FilterMenuUIState. applyParams(...) gained a tagId argument. 
- Gate read follows the established configReady + EvalReason.Unrecognized guard pattern. This `SearchAndFilterViewModel.isOpenCallsEnabled` will be consumed to hide/show the UI piece in follow up PR's.

👀 See
Screenshot 2026-07-28 at 12 01 46 PM

📋 QA
No user-facing UI yet, but use Network Profiler to check the newly added query response. Unit test coverage added in FilterMenuViewModelTest and SearchAndFilterViewModelTest.

Story 📖
[DISC-386](https://kickstarter.atlassian.net/browse/DISC-386)
[DISC-388](https://kickstarter.atlassian.net/browse/DISC-388)
[DISC-392](https://kickstarter.atlassian.net/browse/DISC-392)

[DISC-386]: https://kickstarter.atlassian.net/browse/DISC-386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
+++ b/app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
@@ -36,6 +36,7 @@ class SearchAndFilterActivity : ComponentActivity() {
             viewModelFactory = SearchAndFilterViewModel.Factory(env)
             filterMenuViewModelFactory = FilterMenuViewModel.Factory(env)
             filterMenuViewModel.getRootCategories()
+            filterMenuViewModel.getTags()
 
             setContent {
                 KickstarterApp {
--- a/app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
+++ b/app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
@@ -59,6 +59,16 @@ class SearchAndFilterViewModel(
     private val _isVideoFeedBannerVisible = MutableStateFlow(false)
     val isVideoFeedBannerVisible: StateFlow<Boolean> = _isVideoFeedBannerVisible.asStateFlow()
 
+    /**
+     * Whether the Open Calls filter (pill, filter-menu row, and tag fetching) should be exposed,
+     * derived from the [StatsigGateKey.ANDROID_OPEN_CALLS] gate. Recomputed on the client's
+     * `configReady` — not the one-shot `isReady` — so the gate is never read mid user-reload
+     * (login/logout) where it returns `Loading:Unrecognized`/`false`, mirroring
+     * [isVideoFeedBannerVisible].
+     */
+    private val _isOpenCallsEnabled = MutableStateFlow(false)
+    val isOpenCallsEnabled: StateFlow<Boolean> = _isOpenCallsEnabled.asStateFlow()
+
     private val _searchUIState = MutableStateFlow(SearchUIState())
     val searchUIState: StateFlow<SearchUIState>
         get() = _searchUIState
@@ -96,6 +106,15 @@ class SearchAndFilterViewModel(
                 val gate = statsigClient.getFeatureGate(StatsigGateKey.ANDROID_VIDEO_FEED.key)
                 if (gate.getEvalDetails().reason == EvalReason.Unrecognized) return@collect
                 _isVideoFeedBannerVisible.value = gate.getValue()
+            }
+        }
+
+        scope.launch {
+            statsigClient.configReady.collect { configReady ->
+                if (!configReady) return@collect
+                val gate = statsigClient.getFeatureGate(StatsigGateKey.ANDROID_OPEN_CALLS.key)
+                if (gate.getEvalDetails().reason == EvalReason.Unrecognized) return@collect
+                _isOpenCallsEnabled.value = gate.getValue()
             }
         }
 
@@ -138,7 +157,8 @@ class SearchAndFilterViewModel(
         recommended: Boolean = false,
         projectsLoved: Boolean = false,
         savedProjects: Boolean = false,
-        social: Boolean = false
+        social: Boolean = false,
+        tagId: Int? = null
     ) {
         val update = params.value.toBuilder()
             .apply {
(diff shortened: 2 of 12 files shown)


app/src/main/java/com/kickstarter/features/search/ui/SearchAndFilterActivity.kt
```diff
@@ -36,6 +36,7 @@ class SearchAndFilterActivity : ComponentActivity() {
             viewModelFactory = SearchAndFilterViewModel.Factory(env)
             filterMenuViewModelFactory = FilterMenuViewModel.Factory(env)
             filterMenuViewModel.getRootCategories()
+            filterMenuViewModel.getTags()
```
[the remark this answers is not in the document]
Arkariang: Follow up PR wil hide/show the UI using the gate, can comment out this piece, is as of now more to facilitate QA.

app/src/main/java/com/kickstarter/services/KSApolloClientV2.kt
```diff
@@ -368,10 +371,18 @@ class KSApolloClientV2(val service: ApolloClient, val gson: Gson) : ApolloClient
...
+            staffPicks = if (discoveryParams.staffPicks() == null) Optional.absent() else Optional.present(discoveryParams.staffPicks()),
+            tagId = discoveryParams.tagId().toOptional()
         )
     }
 
+    /**
+     * Collapses the repeated `if (x == null) Optional.absent() else Optional.present(x)` pattern
+     * used when building Apollo query variables.
+     */
+    private fun <T : Any> T?.toOptional(): Optional<T> =
```
Arkariang, line 383: This could be applied broadly on this entire file, will do so on follow up PR if you agree @tonyteate

tonyteate: Nice! Yep, honestly surprised this isn't built-in.

app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
```diff
@@ -99,6 +106,15 @@ class SearchAndFilterViewModel(
             }
         }
 
+        scope.launch {
+            statsigClient.configReady.collect { configReady ->
```
tonyteate, line 110: Do we want to `collect` `configReady` here continuously? Could useful to do a `filter` or one-time check e.g. `statsigClient.configReady.first { it }`.

app/src/main/java/com/kickstarter/models/Tag.kt
```diff
@@ -0,0 +1,56 @@
...
+    val url: String,
+    val slug: String,
+) : Parcelable {
+    fun id() = this.id
+    fun name() = this.name
+    fun url() = this.url
+    fun slug() = this.slug
+
+    @Parcelize
+    data class Builder(
```
tonyteate, line 26: - With `Tag` as a data class with a public constructor, is the Builder still necessary?
- If we keep the Builder, I'd argue that it should not be Parcelable.

app/src/main/java/com/kickstarter/mock/factories/TagFactory.kt
```diff
@@ -0,0 +1,45 @@
...
+        return make100()
+    }
+
+    @JvmStatic
+    fun make100(): Tag {
+        return Tag.builder()
+            .id(1L)
+            .name("Make 100")
+            .slug("make-100")
+            .url("https://www.kickstarter.com/discover/advanced?tag_id=1")
```
tonyteate, line 17: Super-nit for clarity: I think the `url`s here are of the form `/discover/tags/<tags>` (though I know it's `tag_id` when used a search param)

tonyteate (approved):
Looks good! Please see [this](https://github.com/kickstarter/android-oss/pull/2552#discussion_r3675806139) comment for possible follow-up.

app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
Arkariang: I do not expect this piece to be updated unless a change on the internal statsig user state, or changes on the rollout state. I wanna avoid the flickering we saw with VideoFeed, granted I could add some logging to verify once released really how many times this is called.
Given this specific new filter is not highly visible I could see too collecting just once, maybe I'm playing overly safe here.

app/src/main/java/com/kickstarter/models/Tag.kt
Arkariang: Noted, Will change to a data class on follow up PR!

app/src/main/java/com/kickstarter/features/search/viewmodel/SearchAndFilterViewModel.kt
tonyteate: Gotcha! Would be curious to see the logging if you end up doing it. The previous case seemed like a special case, but if we have to re-use this pattern, we might want to revisit the custom cache key. Ideally, if I'm understanding the position of the documentation, accessing Statsig should be "transparent" -- we can always get a value, change the UX based on the value, but should not have to think about when or where we get that value.

PR #2556, opened 2026-08-05
DISC-433: Mute/un-mute controls

Arkariang:
# 📲 What

Adds a mute / unmute control to the VideoFeed player. Each video now shows a circular mute toggle just below the play button; tapping it silences or restores the video's audio. The choice persists as you scroll through the feed until you change it, and every toggle is tracked as an analytics event.

# 🤔 Why

The VideoFeed autoplays videos, but until now there was no way for a user to control audio — videos always played at full volume with no escape hatch, and no way to turn sound on if they wanted it. A mute toggle is table-stakes for a vertical video feed (TikTok/Reels-style) and was called for in the [Mobile Video Discovery designs](https://www.figma.com/design/SzTHTNWAEOnhawsTAC1W4p/Mobile-Video-Discovery?node-id=5787-10608). We also want to measure how often people mute/unmute to inform future audio defaults.

# 🛠 How

**State** — a single isMuted flag is hoisted in VideoFeedScreen above the pager, so it's shared across all pages and survives configuration changes. Default is unmuted. It's passed into every page's KSVideoPlayer.
Player — KSVideoPlayer gains isMuted / onMuteToggle(isMuted) params. A LaunchedEffect(exoPlayer, isMuted) applies exoPlayer.volume = if (isMuted) 0f else 1f, keyed on the player too so a freshly created player after a swipe inherits the current preference.

**UI** — a glassmorphism centered just below the play control. The icon reflects the current state. It appears and disappears in lockstep with the play/pause control (same showControls flag, same fade+scale transition), so the two read as one controls cluster. the Mute/VolumeUp vector icons were added to Icons.kt;

**Analytics** — mirrors the existing play/pause pattern. Two new CtaContextName values (video_mute, video_unmute); `trackVideoFeedCTAClicked` fires a single CTA Clicked event with context_page = video_feed, context_cta = video_mute | video_unmute, and the project/video ids.

**Analytics ID fix (all CTAs)** — while wiring this up, review surfaced that `trackVideoFeedCTAClicked` was reporting `video_feed_video_id = project.id()`, inconsistent with the impression / progress-bar events which use the video's own `item.videoId`. Fixed by threading the real `videoId` through the CTA pipeline (VideoFeedScreen → VideoFeedActivity → VideoFeedViewModel.onCTAClicked → AnalyticEvents). This corrects `video_feed_video_id` for **every** VideoFeed CTA event — play/pause, save, creator, share, and the new mute/unmute — not just this mute/unmute feature.

Tests:
- KSVideoPlayerTest (visibility tied to controls, volume application, reported toggle state), 
- VideoFeedScreenTest (label toggle, callback reports project + resulting state), 
- SegmentTest (mute/unmute event properties).

# 👀 See

https://github.com/user-attachments/assets/7de4b1a8-5108-4902-9838-9854e893e905

Analitics

| --- | --- |
|  |  |

# 📋 QA

- Open the videofeed, make sure volume is up to hear sound. Hit the middel of the screen, overlay with pause/mute buttons appear. Mute and play video again, swipe to make sure the state is persisted. Unmute again

# Story 📖

[DISC-443](https://kickstarter.atlassian.net/browse/DISC-443)

[DISC-443]: https://kickstarter.atlassian.net/browse/DISC-443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonyteate (approved):
lgtm!

PR #2558, opened 2026-08-07
DISC-436: VideoFeed deeplink

Arkariang:
# 📲 What

Adds a deeplink that opens the Video Feed (VideoFeedActivity). Tapping `https://www.kickstarter.com/video-feed` (or the `ksr://www.kickstarter.com/video-feed` scheme) now launches the video feed screen inside the app.

# 🤔 Why

We need an addressable entry point into the video feed so it can be reached from outside the app (marketing, notifications, internal links) rather than only through the in-app Discovery/Search banners. The path is app-internal and intentionally not matched as a feature on the website. Ignored any gating logic on this PR as currently the VideoFeed gate is rolled out to 100% of users.

# 🛠 How

Routed through the existing single deeplink entry point, SplashScreenActivity → SplashScreenViewModel.DeepLinkViewModel. Both URLs already resolve to the splash screen via its catch-all intent filters, and `/video-feed` doesn't collide with any existing matcher.

**UriExt.kt**: new `Uri.isVideoFeedUri(webEndpoint)` matcher (VIDEO_FEED_PATTERN = /video-feed) , following the same shape as the other matchers so it also works against staging endpoints.

**SplashScreenViewModel.kt**: new `startVideoFeedActivity(): Observable<Unit>` output that emits when the URI matches `isVideoFeedUri`, plus the corresponding `!it.isVideoFeedUri(...)` exclusion in the unsupportedDeepLink chain so the link doesn't also fall through to the browser.

**SplashScreenActivity.kt**: subscribes and launches via startNewDiscoveryActivity() → startVideoFeed(ScreenName.DEEPLINK.value) → finish(), **so Discovery sits beneath the feed for a sane back stack and the entry surface is logged as "deeplink"**.

The android_video_feed Statsig gate is currently at 100%, so the route intentionally does not re-check the gate at the splash router.

# 👀 See

https://github.com/user-attachments/assets/2c45e4c9-dd25-4363-b3cc-c22a7433a034

| --- | --- |
|  |  |

# 📋 QA

- Send an email to yourself with the link `https://www.kickstarter.com/video-feed`
- Using adb: `adb shell am start -W -a android.intent.action.VIEW -d "ksr://www.kickstarter.com/video-feed" `

# Story 📖

[DISC-436](https://kickstarter.atlassian.net/browse/DISC-436)

[DISC-436]: https://kickstarter.atlassian.net/browse/DISC-436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tonyteate (approved):
lgtm! 💯

PR #2562, opened 2026-09-01
PF-48: Upgrade Stripe & map `PaymentSheetResult` to an internal data model

tonyteate:
# 📲 What

Upgrade the Stripe dependency and map its `PaymentSheetResult` to a new `StripePaymentSheetResult` type.

# 🤔 Why

App maintenance and to provide a stable foundation for integrating Google Pay in the future.

# 🛠 How

Bump the Stripe library to the [latest version](https://github.com/stripe/stripe-android/releases/tag/v23.17.1) as of yesterday. Note that, as recommended in [Compatibility with Jetpack Compose](https://github.com/stripe/stripe-android#compatibility-with-jetpack-compose), this version is compatible with the app's current version of Compose UI based on the [BOM version mapping](https://developer.android.com/develop/ui/compose/bom/bom-mapping) for `androidx.compose:compose-bom:2026.03.01`.

The latest version of the library marks `PaymentSheetResult`'s subclass constructors as `internal`, so we create a new `StripePaymentSheetResult` type and map from those classes where needed.

# 👀 See

N/A - There are no notable UI changes w/ this upgrade given the current configuration for the PaymentSheet

# 📋 QA

Regression test the following flows:

* Add a new payment method from Profile Settings
* Crowdfunding checkout with a new payment method
* Late pledge checkout with a new payment method
* Change payment method

# Story 📖

[\[PF-48\] Upgrade Stripe (requires kotlin 2.1+) - Jira](https://kickstarter.atlassian.net/browse/PF-48)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -249,7 +249,7 @@ dependencies {
     kapt "com.google.dagger:dagger-compiler:$dagger_version"
     implementation 'com.jakewharton:process-phoenix:3.0.0'
     implementation "com.jakewharton.timber:timber:5.0.1"
-    implementation 'com.stripe:stripe-android:20.47.3'
+    implementation 'com.stripe:stripe-android:23.17.1'
 
     final okhttp_version = '5.1.0'
     implementation "com.squareup.okhttp3:logging-interceptor:$okhttp_version"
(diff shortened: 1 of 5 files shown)


Arkariang (approved):
lgtm!

